PR #67, opened 2013-10-28
Replace nova-network for Quantum using Linux Bridge plug-in

emagana:
In this commit, PLUMgrid team has replaced the OpenStack Grizzly configuration using nova-network for Quantum with Linux Bridge as back-end plugin.
These changes have been successfully tested in PLUMgrid Lab.
Please, review this pull request and provide feedback.


kamidzi:
We're off grizzly.

PR #80, opened 2014-02-11
Build bins changes to support:

cbaenziger:
* A local Apt repo
* A local PyPi repo
* A local RubyGem repo
* Updating recipes to not use cookbook_file actions for large files
* Fixing the Apache mirror recipe to use the BCPC installed Apache server

Additional changes:
* Added a number of set -o errtrace's into files to see failures
* Addition for chefit.sh to build apt.conf if http_proxy is in use
* Fix to resolve management IP for bootstrap node if it is not the first interface (e.g. a VM)
* Addition to proxy_setup.sh for providing no_proxy env for local IP's and the chef server IP (as provided by Chef)


kamidzi:
This is stale.

Issue #116, opened 2014-04-25
Upgrade hypervisors to trusty (14.04)


jerenkrantz:
We should think about when we want to upgrade from precise/12.04 to trusty/14.04.

One notable change that we know should be fixed in upstream packages captured in 14.04 is #54 where keepalived can drop the VIP under load.  I expect there's more.

Thoughts?

erhudy:
Closing as this work was done as part of 4.0/5.0.

PR #233, opened 2014-07-26
Upgrade hypervisors to Trusty (14.04)

jerenkrantz:
This is a branch to support the hypervisors running Trusty (14.04).

See also #116 and #141.


erhudy:
I'm cancelling this pull request to avoid conflict/confusion since there will be another PR for the unstable/trusty branch soon.

PR #382, opened 2014-11-13
sshd listens on additional ip

caiush:
if something bad happens on eth0, we can still get in on something else.


kamidzi:
Are we still trying to do this?

mihalis68:
I am in favor of this. Do you think it is safe Kundai?

kamidzi:
I think it's safe enough, though the intent is a little unclear. As-is, this just binds to another address. It doesn't necessarily address the issue of a problem on eth0. Also, I imagine that if the machine is not accessible on the management ip, there are other, larger problems at hand that won't necessarily be solved by logging in via the storage network. Also, the idea of mixing storage traffic with other traffic seems a bit unclean.

kamidzi:
Offline, it was discussed that this is a non-ideal, temporary convenience.

PR #393, opened 2014-11-24
Preseed ssh keys during OS installation

### LINKED ISSUE #380: Bootstrap process should allow password-less SSH as root to bcpc-vm* machines
Currently, the bootstrap process requires manual intervention due to password prompting. This should not be the case.

kamidzi:
Adds mechanism to generate and authorize temporary root key for `knife bootstrap`. It should be noted that the generated key is only readable by the initial user calling `vbox_create` on the bootstrap node and it will be unauthorized during the bootstrap procedure when another key is generated/authorized. closes #380


mihalis68:
This looks good to me to. My method is simply using sshpass all over the place. Deeply inferior to this. Let's do it!

PR #407, opened 2014-12-02
Add more functional tests to nova check

kkoh61:
This commit introduces following changes:
1. Support "--host" parameter for instance creation on specific compute host. Forcing instance creation on a specific compute host may require a more privileged user, but this will be handled separately.
2. Test floating IP association as part of instance creation test
3. Some indentation updates (use 4 whitespaces)


kkoh61:
Updated. It now looks up the 'host' key in nova.yml. Also included a role change for tester user as current nova policy mandates that compute:create:forced_host requires admin.

caiush:
Looks good.Merge it.

PR #436, opened 2015-01-07
Routemon v2

mihalis68:
This pull request significantly upgrades the "routemon" service. Previously it just attempted to log when routes in the 'mgmt' and 'storage' tables went away and what the very last event reported by "ip monitor" was.

Now when routes go away it reports the last 10 or so events. This is so that if the critical event was a couple of events ago, we stll record it.

Additionally (and optionally) routemon can attempt to actually fix the routes. This may be of use when routes in the BCPC named routing tables disappear as they sometimes do (not sure why).

For this you need to set the numfixes attribute to > 0 so for example in an environment file this could look like 

```
   "name": "Test-Laptop",
   "override_attributes": {
     "bcpc": {
+      "routemon": {
+          "numfixes": "1000"
+       },
```

When routes are "fixed" the routemon service throws away the resultant output (which is noisy) by re-running the "ip monitor" command and emptying the trace buffer of previous events. The actual fix involves simply re-running /etc/network/if-up.d/bcpc-routing

### DIFF
--- a/cookbooks/bcpc/files/default/routemon.pl
+++ b/cookbooks/bcpc/files/default/routemon.pl
@@ -37,13 +37,18 @@ sub hasdefaultroute {
 }
 
 # number of times we are allowed to fix the routes
-my $fixes = 1000;
+# default 0 => do not try at all
+my $fixes = 0;
 
 my $override = shift @ARGV;
-if ($override) {
+if ($override>0) {
     $fixes = $override;
-    myprint "Allowable fix attempts set to: $fixes\n";
 }
+
+myprint "-----------------------------------------------------\n";
+myprint "Service \"routemon\" starting ...\n";
+myprint "-----------------------------------------------------\n";
+myprint "Allowable fix attempts set to: $fixes\n";
 
 # check and retain current status of routes for comparison later
 my $mgmt_up    = hasdefaultroute ("mgmt", 1);
--- a/cookbooks/bcpc/files/default/routemon.pl
+++ b/cookbooks/bcpc/files/default/routemon.pl
@@ -25,7 +25,7 @@ sub hasdefaultroute {
     my $network = shift(@_);
     my $verbose = shift(@_);
     myprint "checking $network network\n" if "$verbose";
-    my $command = "ip route show table $network | grep -i default";
+    my $command = "ip route show table $network | grep -i '^default'";
     my $result = system("$command >/dev/null 2>&1");
     if ($result) {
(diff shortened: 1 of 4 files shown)


cookbooks/bcpc/files/default/routemon.pl
```diff
@@ -1,61 +1,158 @@
...
+# Param 2 can optionally specify verbose output
 # Returns 1 if network has default route present, else return 0
-sub checkroute {
+sub hasdefaultroute {
 
     my $network = shift(@_);
     my $verbose = shift(@_);
-    print "checking $network network\n" if "$verbose";
+    myprint "checking $network network\n" if "$verbose";
     my $command = "ip route show table $network | grep -i default";
```
kamidzi: maybe anchor the regex?

mihalis68: yes that's a good point. I will do so.

cookbooks/bcpc/files/default/routemon.pl
```diff
@@ -1,61 +1,158 @@
...
-    my $currentmgmt = checkroute("mgmt",0);
-    if ($currentmgmt != $mgmt) {
-	if ($currentmgmt) {
-	    print "Info: Default route established on mgmt network after \n $LINE\n";
-	} else {
-	    print "WARN: Default route disappeared on mgmt network after \n $LINE\n";
-	}
+sub fix_routes {
+    if ($fixes>0) {
+        myprint "Now attempting to fix default routes...\n";
```
kamidzi: How does running this multiple times fix the issue? I'm not quite sure why this the case.

kamidzi:
Something to consider might be rtmon, logging to a file stream and reading that.

cookbooks/bcpc/files/default/routemon.pl
mihalis68: routemon will run for a long time so it might fix routes and then they might break again. I have a maximum number of times I'll try to fix the routes to avoid any possible "runaway" states where it somehow tries to fix them, but they break during that process and it tries again - eating its own tail. With the limiter, a runaway case would exhaust the allowed attempts. Normal business should allow a quota of, say 1000 to allow the routes to keep getting fixed for weeks or months.

cookbooks/bcpc/files/default/routemon.pl
mihalis68: done

mihalis68:
For the record after the commit above I received a verbal "looks good to me on this" from kamidzi :)

PR #438, opened 2015-01-08
basic tpm recipe

mihalis68:
This can enable use of a hardware based random number generator as provided by some TPM modules. Default disabled - VMs don't have this support.

### DIFF
--- a/cookbooks/bcpc/recipes/tpm.rb
+++ b/cookbooks/bcpc/recipes/tpm.rb
@@ -1,0 +1,45 @@
+#
+# Cookbook Name:: bcpc
+# Recipe:: tpm
+#
+# Copyright 2014, Bloomberg Finance L.P.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+if node['bcpc']['enabled']['tpm'] then
+
+  include_recipe "bcpc::default"
+
+  package "rng-tools"
+  package "tpm-tools"
+  package "trousers"
+
+  service "rng-tools" do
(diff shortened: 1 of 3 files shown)


cookbooks/bcpc/recipes/tpm.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  service "rng-tools" do
+    action :stop
+  end
+
+  bash "enable-tpm" do
+    user "root"
+    code <<-EOH
+       if [[ ! -f /etc/default/rng-tools.orig ]]; then cp /etc/default/rng-tools /etc/default/rng-tools.orig ; fi
+       sed -i -e '/^#RNG.*tpm.*/s/^#//' /etc/default/rng-tools
```
kamidzi: It may be better to explicitly write out the configuration rather than attempt to modify a file whose contents maybe unknown.

mihalis68: Yes I can do that.  Had similar feedback on a more elaborate version i had a while back that never got merged

Sent from my iPhone

----- Original Message -----
From: Kundai Andrew Midzi <reply@reply.github.com>
To: chef-bcpc@noreply.github.com
CC: CHRIS MORGAN
At: Jan 8, 2015 6:04:55 PM


In 
cookbooks/bcpc/recipes/tpm.rb:> +
> +  include_recipe "bcpc::default"
> +
> +  package "rng-tools"
> +  package "tpm-tools"
> +  package "trousers"
> +
> +  service "rng-tools" do
> +    action :stop
> +  end
> +
> +  bash "enable-tpm" do
> +    user "root"
> +    code <<-EOH
> +       if [[ ! -f /etc/default/rng-tools.orig ]]; then cp /etc/default/rng-tools /etc/default/rng-tools.orig ; fi
> +       sed -i -e '/^#RNG.*tpm.*/s/^#//' /etc/default/rng-tools 

It may be better to explicitly write out the configuration rather than attempt to modify a file whose contents maybe unknown.

caiush:
I believe that kundai is correct that it is a good idea to template, but not because I think is the right way to do it, but because it is the way we have chosen to do it in bcpc, both ways have merits. 

So I'm okay with this being merged on the condition that a issue is cut and assigned to @mihalis68 to fix this and the other places we are not tempting in bcpc. That issue should be addressed with some urgency.

mihalis68:
Issue for cleaning up the in-place sed expressions : https://github.com/bloomberg/chef-bcpc/issues/439

PR #440, opened 2015-01-09
Correct poorly-formed tarfile for cookbooks. Add comments

mihalis68:


mihalis68:
This is a utility I use for extracting binaries from VM-based builds for use on hardware elsewhere. This currently has no dependents (I use this by hand, nothing else uses it). Hence merging.

PR #442, opened 2015-01-12
Reduce DHCP lease time to allow quicker recycling of IP addresses

kkoh61:
This is possibly a temporary mitigation until we can assess flipping force_dhcp_release to True.


kamidzi:
This looks a forward-port of a configuration that has existed since release series 2.x. Looks like this was integrated internally around Aug 2014...Good catch!

PR #445, opened 2015-01-16
Increase minimum reserved memory on hosts, vm.min_free_kbytes

dkaplan34:
Increase reserved memory on Ubuntu Hosts to reduce memory paging issues


caiush:
Looks good.Do we have an issue for this open in github?

dkaplan34:
Issue
https://github.com/bloomberg/chef-bcpc/issues/446

PR #453, opened 2015-01-20
Fix for LDAP initialization on new cluster

pchandra:
Calls to get_config that are contained in anything *except* a `ruby_block` will get early resolution when chef initially compiles all the recipes. For a brand new cluster build (specifically, with an empty data bag) this results in the `setup-389ds-server` clause leaving an empty replication passwd for the very first node coming up. There's probably a better way to do this, but I converted it from `bash` to `ruby_block` in this PR in case a quick fix is needed. This hasn't been tested yet since my cluster test bench is mid update, so caveat emptor.

### DIFF
--- a/cookbooks/bcpc/recipes/389ds.rb
+++ b/cookbooks/bcpc/recipes/389ds.rb
@@ -43,32 +43,35 @@ template "/tmp/389ds-install.inf" do
     mode 00600
 end
 
-bash "setup-389ds-server" do
-    user "root"
-    code <<-EOH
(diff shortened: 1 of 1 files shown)


mihalis68:
I replicated the issue by building a three-head-node cluster on VirtualBox. I could see the slapd passed unconfigured on my first head node and I easily provoked errors in creating tenancies and users via Horizon. I merged this PR into my working branch manually and rebuilt from scratch and then the resulting cluster was quite solid by contrast. This looks good to me.

PR #456, opened 2015-01-21
Check get_config for nil return

pchandra:
To prevent timing bugs in the future, I added a quick check to raise an exception when get_config is about to return nil. As it turns out, this is a very good lint test and it turned up a number of bugs that needed to be fixed in order to build a new cluster. So, this PR includes all those fixes too. It's probably a lot easier to review this commit-by-commit instead of all at once.


kamidzi:
I found some additional changes that were required. See #463

kamidzi:
Superseded by #463

PR #461, opened 2015-01-28
Move Graphite from Headnode to new Monitoring role

kkoh61:
This creates a new monitoring role to offload some of head node's responsibilities permanently. Initially, only Graphite will be moved.

1. Creation of BCPC-Monitoring role, which is intended to be home for metrics, alerting and logging.
2. Upgrade of Graphite components
3. Direct Graphite-memcached access.
4. Split MySQL recipe into two - one for monitoring (currently graphite, but could take on Zabbix, etc) and the other (mysql-head) for Openstack and everything else. Credentials for haproxy's MySQL check will be reused in new database. 
5. Corresponding haproxy updates. The "backup" parameter is omitted in mysql-monitoring backend so database traffic is expected to be evenly distributed.
6. Extension of utils library to allow node search based on role/recipe, which may be of increasing use if chef-bcpc nodes become more specialized.


cookbooks/bcpc/recipes/mysql-head.rb
```diff
@@ -0,0 +1,88 @@
...
+                mysql -u root -p#{get_config('mysql-root-password')} -e "GRANT USAGE ON *.* to #{get_config('mysql-galera-user')}@'%' IDENTIFIED BY '#{get_config('mysql-galera-password')}';"
+                mysql -u root -p#{get_config('mysql-root-password')} -e "GRANT ALL PRIVILEGES on *.* TO #{get_config('mysql-galera-user')}@'%' IDENTIFIED BY '#{get_config('mysql-galera-password')}';"
+                mysql -u root -p#{get_config('mysql-root-password')} -e "GRANT PROCESS ON *.* to '#{get_config('mysql-check-user')}'@'localhost' IDENTIFIED BY '#{get_config('mysql-check-password')}';"
+                mysql -u root -p#{get_config('mysql-root-password')} -e "FLUSH PRIVILEGES;"
+            ]
+        end
+    end
+end
+
+db_user = get_config('mysql-root-user')
```
kamidzi: I don't believe this will work as expected since this value is initialized above during _convergence_ in the ruby_block whereas this is being referenced prior to that during compilation. See https://docs.chef.io/resource_ruby_block.html . Specifically, the result is:

    Recipe Compile Error in /var/chef/cache/cookbooks/bcpc/recipes/mysql-head.rb

    RuntimeError
    ------------
    No config found for get_config(mysql-root-user)!!!

    Cookbook Trace:
    ---------------
      /var/chef/cache/cookbooks/bcpc/libraries/utils.rb:69:in `get_config'

kkoh61:
@caiush and I spoke offline and there is a concern with depending on headnode's VIP for Graphite/monitoring, as VIP jumps may result in intermittent loss of metrics and access. It is also a potential chokepoint. Some more work is expected to break out haproxy and keepalived recipes for deployment on monitoring nodes.

mihalis68:
To capture some more internal discussion here, a concern was expressed that this is a large PR. It was explained that a lot of it is due to necessary changes to the utils. My suggestion is to land those changes first as a neutral "no-op" first which is easier to verify - support everything as before with whatever changes are necessary when calls must be different or perhaps more specific but no changes in cluster architecture. For example, you could adjust everything that used get_all_nodes to whatever is correct in the new system. A cluster built from the resulting code need only behave the same way as previous clusters, so we could verify the utils changes before seeing the changes for the new role.

kkoh61:
That's a valid suggestion. I will create separate PRs.

kamidzi:
See #483

PR #462, opened 2015-01-28
Extend setting of MTUs to work properly in the presence of VLANs

mihalis68:
On real hardware using VLAN interfaces for floating and/or storage networks, you cannot simply set the MTU for those networks without considering the underlying physical interface. With this pull request, if a parent interface is specified in the environment file, we will apply the MTU to that interface as well as to the VLAN interface (statically) and also configure this in network interface fragment files.


caiush:
Love it.

caiush:
Can we get some docs written in the wiki section on deploying a real hardware cluster about how to configure if using vlans?

mihalis68:
ok will do

mihalis68:
Done: https://github.com/bloomberg/chef-bcpc/wiki/BCPC-network-configuration-when-using-VLANs

PR #470, opened 2015-01-30
Implement search_nodes method that searches by role/recipe

kkoh61:


cookbooks/bcpc/libraries/utils.rb
```diff
@@ -70,8 +70,21 @@ def get_config(key)
     return result
 end
 
+def search_nodes(key, value)
+    if key == "recipe"
+        results = search(:node, "recipes:bcpc\\:\\:#{value} AND chef_environment:#{node.chef_environment}")
+    elsif key == "role"
+        results = search(:node, "role:#{value} AND chef_environment:#{node.chef_environment}")
```
caiush: Bit worried about this. Often when I have hardware specific roles, e.g. a worker node with all HDDs I create a new role e.g. `BCPC-Worknode-HDD`, I do a node override attributes for this node and create a run list that includes the generic Worknode role. e.g.:
```
{
    "name": "BCPC-Worknode-HDD",
    "default_attributes": {
    },
    "json_class": "Chef::Role",
    "run_list": [
      "role[BCPC-Worknode]"
    ],
    "description": "A functional compute node in a BCPC cluster",
    "chef_type": "role",
    "override_attributes": {
      "bcpc": {
        "ceph" : {  
            "hdd_disks" : [ "sdb", "sdc", "sdd"],
            "ssd_disks" : []
        }       
      }
    }
}
```
In this case, do we want the search to return the HDD node when asking about "wroknodes"?

caiush:
I think this is okay to merge

kkoh61:
Submitted an amendment, but I could go either way with this. Search by recipe should suffice in most cases. If it suits, I can watch how use cases develop and change accordingly.

PR #475, opened 2015-02-05
Ensure replicas value is at least 1

kkoh61:


cookbooks/bcpc/recipes/ceph-head.rb
```diff
@@ -167,6 +167,10 @@
 end
 
 replicas = [search_nodes("recipe", "ceph-work").length, node['bcpc']['ceph']['default']['replicas']].min
+if replicas < 1; then
```
kamidzi: How about just adding 1 to the array on the line above?

kamidzi:
Otherwise, looks good. It builds.

PR #476, opened 2015-02-06
Short term fix for DNS using EC2 style hostnames

caiush:
# Problem statement: 

Currently in BCPC a VM gets a hostname communicated to it through 2 possible vectors:

1. via cloud-init and the meta-data service 
2. via dnsmasq dhcp leases 

The hostname supplied to it is,  currently `<vmname>.bcpc.exmaple.com` (assuming your are using the default environment). However, this doesn't resolve to .... anything. The only valid VM hostnames in pdns are of the form `<vmname>.<tanantname>.bcpc.exmaple.com` which resolve to floats. Fixed ips do not currently have any entry in pdns. This causes all kinds of problems for various tenant software that assume a working DNS. Note, that the search domain issued by dhcp doesn't work either. 

Secondly, the BCPC pdns structure is very complex involving multiple "temp" tables, views and a cron job that collapses them down to a static table every min. Apart from being brain numbingly complex I've also seen intermittent DNS failures. I think this is because the cron job is completely nuking the records tables every min and rebuilding them. On a busy cluster this means you can hit the records table when its empty (or partially empty) and get DNS failures.

Thirdly, Openstack as it is configured in BCPC to use special, precious, host names as typed in by the user. I think the EC2 model is more sane, where one is given a unique generated hostname and (if required) after creation the user creates a CNAME to point to this A/PTR record pair. Nothing could be more "pet-like" than naming your VMs. 

# Proposed solution 

This PR addresses these 3 issues by giving VMs a host name of the form "ip-X-X-X-X.bcpc.exmaple.com" for the fixed ips and "public-X-X-X-X.bcpc.example.com"  for the float. As we know all possible future IP/Hostname mappings a head of time these are preloaded into power. I do still provide the ability to resolve the old `<vmname>.<tenant>,bcpc.example.com` name, but *only as a convince for the users* and for the sake of backwards compatibility. I do this by creating a CNAME in power pointing to the float A record. Note, this last mapping is done via a cron job (see notes below on why I did such an evil thing) every two mins, Note, it *does not* blow away the records DB every time, so the second issue above should be resolved.   

I investigated setting the hostname in nova db to "ip-X-X-X-X" however, nova has the idea of pet names for its VMs at is core (indeed its not even consistent internally on this point either, sometimes using instance.name, sometimes instance.display_name), and after a couple of hours faffing around with it I went with a different approach. Rather than changing the hostname in instances table, I inject my "ip-" hostname into the VM at the last possible moment via the metadata service and the dnsmasq dhcp leases file.  

For loading the A/PTR records into power, in each chef run I compare the nova.fixed_ip table with the pdns.records table and only repopulate the records if these are different. This means that if you expand the number of fixed subnets at a later time the pdns tables will be automatically be updated on the next chef run. Note, that if there is no change in network configs, no action is taken on the pdns tables. 

Note I also enabled the pdns web interface.. its quite useful. 

# Notes on alternative archs explored and suggestions for future work.

I explored two other variations on populating pdns and can supply (only slightly broken) code on request.  

PowerDNS has the awesome feature of changing out the back ends, so rather than load a bunch records into the table I wrote a small python script to algorithmically do the translation of "ip-X-X-X-X" to "X.X.X.X". This works really nicely, especially if we can couple it with a mysql backend for the CNAMEs and other A records (e.g. the VIP records). If this could cleaned up it would scale much better and, perhaps, remove the need for mysql pdns all together. 

I also explored feeding off the notifications rabbit q's, when I see a message saying a float is attached I created the CNAME records and when detached I destroy them. This is basically what Designate does. However, I didnt go with this for two reasons. Firstly, if there was ever an issue with rabbitMQ (and gosh that would *never* happen) then we could fall of sync between nova and power. I could periodically check that the two match ... but then thats basically the cron job I have. Second, openstack allows the users to rename VMs (which is probably insane), at this point I want the CNAME to change to reflect this. 

One other thing I played with but havnt put into this PR is re-designing the front end of power. Currently, pdns binds to the vip meaning we have to bounce pdns everytime the vip jumps. This has been unstable. I've seen pdns get really wedged up during these VIP jumps. In addition the current arch isnt actually active-active HA. 

I played with putting all headnodes into a global dnsmasq config for the hypervisors, then have the VM dnsmasqs talk to localhost. I move powerdns off the vip and put it on the mgmt ip. In the global config I put all N headnodes mgmt ips (note, unlike resolv.conf, dnsmasq allows >3 dns servers).   This means we now have true HA for dns and one less thing going through the vip. Downside is, it adds a bit more latency for the dns requests but on large clusters this might be made up for by the better load balancing. It would also require changes to NS records of any upstream DNS servers. 

One more minor thing: I think I should probably use my method for populating the fixed ip pdns records for floats. Floats are currently done by ripping through all possible floats and checking the DB of each. This gets slow quickly. Fixed ip are populated by selecting nova.fixed_ips into pdns.records which is much faster.  
  

# Notes on upgrading to this PR

:warning: This is not a change you can just chef in. To deploy this change one needs to:
1. Backup and drop the current pdns database
2. `sudo crontab -e` and remove the old `populate_records` cron entry.
3. Chef in the new config.


erhudy:
Picking this PR up at Chris M's suggestion.

erhudy:
Since this pull request was 2 months old, it required slightly more extensive overhauling after merging master back in. I pushed the re-merged branch to my own copy of the repo (https://github.com/erhudy/chef-bcpc/tree/dns_ip_hostnames) with the following additional changes:
* Chef will remove the old PowerDNS update cron job
* Chef will back up and drop the pdns database if the `bcpc:enabled:automate_migration_to_fixed_ip_dns` attribute is set to true, which I added to make it possible to do the migration without manual intervention
* fixed a few issues with resource ordering and file resources being named something different from what the recipe expected

erhudy:
After discussing with Caius, he asked me to create a new pull request with my updated branch. Closing this one to avoid ambiguity.

PR #477, opened 2015-02-06
Method to generate VRID without conflicting with existing ones

kkoh61:
This is to facilitate running additional keepalived instances.

### DIFF
--- a/cookbooks/bcpc/libraries/utils.rb
+++ b/cookbooks/bcpc/libraries/utils.rb
@@ -223,3 +223,13 @@ def ping_node_list(list_name, ping_list, fast_exit=true)
         raise ("Network test failed: #{list_name} unreachable")
     end
 end
+
+def generate_vrrp_vrid()
+    init_config if $dbi.nil?
+    dbi = Chef::DataBagItem.load('configs', node.chef_environment)
+    a =  dbi.select {|key| /^keepalived-.*router-id$/.match(key)}.values
+    exclusions = a.collect {|a| [a-1, a, a+1]}.flatten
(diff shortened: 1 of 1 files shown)


caiush:
I like this. I spume once generated you dump it back in the Databag?

cookbooks/bcpc/libraries/utils.rb
```diff
@@ -223,3 +223,22 @@ def ping_node_list(list_name, ping_list, fast_exit=true)
         raise ("Network test failed: #{list_name} unreachable")
     end
 end
+
+def generate_vrrp_vrid()
+    init_config if $dbi.nil?
+    excluded_vrid = Array.new
+    dbi = Chef::DataBagItem.load('configs', node.chef_environment)
+    dbi.each do |key, value|
```
kamidzi: I would try to avoid this sort of looping, maybe something like:

````ruby  
  a =  dbi.select {|k,v| /^keepalived-.*router-id$/.match(k)}.values
  ex = a.collect {|a| [a-1,a,a+1] }.flatten
  res = (1..254).to_a - ex
  raise "Unable to generate unique VRID" if res.empty?
  res.first
````
though it's not like I profiled it or anything of that sorts...

kkoh61:
Implementing @kamidzi's suggestion.

@caiush, that will be left to make_config():
make_config('keepalived-router-id', generate_vrrp_vrid)

kamidzi:
Great, could you squash the commits into one? I don't think there's a need for two separate ones.

kkoh61:
Updated.

PR #479, opened 2015-02-09
script to check if MySQL node is in quorum

kkoh61:
Potentially useful as a keepalived check script.


caiush:
talked with @kkoh61 offline. looks okay for now.

PR #487, opened 2015-02-18
Enable sub domain bucket format for RGW

### LINKED ISSUE #152: Ceph not configured to use bucket subdomains
Need to do the Ceph configuration necessary for bucket-based s3 subdomains. We decided to wait until Firefly upgrade was complete. DNS changes required are in #137.

caiush:
Most 3rd part s3 tools expect the url format: `<bucketname>.s3.bcpc.example.com`. I kept having to hack in the old calling format `s3.bcpc.example.com/<bucketname`. This PR fixes that. 

I tested that I was still able to use the old "path bucket names" and the subdomain bucket names side by side. I also checked it works over http and https. 

fixes #152 

:warning: This will not be a straight "chef in" but will require the pdns database to be dropped and reloaded by chef OR just the `*.s3.bcpc.example.com` CNAME records added by hand to the pdns.recrods_static table.


mihalis68:
For the record, an existing S3/RGW test that I used for the admin tenant produced the URL http://192.168.100.5/swift/v1/fill_test_admin_tenant

I presume the /swift/v1 part disappears when this change moves the bucket from being a sub-path of a domain to being an actual sub-domain?

mihalis68:
The S3 style URL also works, ignore previous comment.

mihalis68:
I tested this and it works. Had to fix up the merge by hand. Will close

PR #491, opened 2015-02-21
Enable keystone token caching with memcached

pchandra:
This enables keystone token caching using the `memcached` running on the VIP holder machine. Enabling this caching helps with scaling the API services that make a lot of keystone calls (ultimately should hit the database a lot less and more token reuse). I can't properly performance test it in VMs, but I can verify that this does result in memcached being hit quite a bit (with a large percentage of cache hits). I've seen this change recommended in a number of places (such as [this preso on scaling openstack neutron](http://www.slideshare.net/vbannai/neutron-scaling-ebay-openstack)).

### DIFF
--- a/cookbooks/bcpc/recipes/keystone.rb
+++ b/cookbooks/bcpc/recipes/keystone.rb
@@ -38,8 +38,10 @@ ruby_block "initialize-keystone-config" do
     end
 end
 
-package "keystone" do
-    action :upgrade
+%w{keystone memcached}.each do |pkg|
+    package pkg do
+        action :upgrade
+    end
 end
 
 template "/etc/keystone/keystone.conf" do
--- a/cookbooks/bcpc/templates/default/keystone.conf.erb
+++ b/cookbooks/bcpc/templates/default/keystone.conf.erb
@@ -534,31 +534,31 @@ log_dir = /var/log/keystone
 # cache region. This should not need to be changed unless
 # there is another dogpile.cache region with the same
 # configuration name. (string value)
-#config_prefix=cache.keystone
+config_prefix=cache.keystone
 
 # Default TTL, in seconds, for any cached item in the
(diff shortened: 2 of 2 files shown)


caiush:
Ugh I was just testing my own or for this last night.  As my solution was exactly the same (expect I upped the ttl at the same time) lets go with this. 

As this only effects auth it should be a straight chef in. All tenants currently using the Apis will just have to re auth. I guess some people might get booted from the ui too

pchandra:
Yeah, sorry for doing this at the same time, I just needed it for testing some neutron/contrail bits so submitted the PR. One thing I did find was that on a new cluster build, it was failing since memcached wasn't installed yet (54d38ed), but chefing it into a running cluster seems to work like a charm (modulo the re-auth's you mentioned).

caiush:
@kamidzi I'd like to get this into a patch release this week, are you good with that?

mihalis68:
Since this went in I seem to see more occasions where the cluster foundations are healthy but I can't login to Horizon. Hitting the cluster with a big hammer (rechef, hup_openstack every node) will fix it. Could this be simplified as per openstack recommendations here

http://docs.openstack.org/high-availability-guide/content/_memcached.html

?

mihalis68:
I've diagnosed the above issue. There is actually NO PROBLEM with the token caching, instead what happens is after a cold shutdown and restart of a BCPC cluster, to be able to log in (at least with Safari on my Mac) you must completely exit the browser and restart. I imagine there are some sort of cached session credential in the browser that are no longer valid after a cluster restarts.

erhudy:
I've seen issues with Safari doing weird things with credentials as well; not this specific issue, but other ones that could only be resolved by exiting and restarting the browser.

pchandra:
I think that's been an issue in Horizon for some time now. I believe the issue is with stale cookies in the browser cache (if I clear the cache and reload, it works).

PR #493, opened 2015-02-22
Live migration and patch cleanup

pchandra:
I wanted to test live migration with Contrail, so I needed to reapply the live migration patch. Turned out it didn't apply cleanly (maybe that's why it was disabled in master?), so I fixed it. Also cleared out the SQL-related patches since these are packaged upstream in the openstack deb packages. In my testing, it does all work fine with this PR.... but if the live migration patch was disabled for some other reason, feel free to ignore this.


kamidzi:
@caiush, @mihalis68, @dkaplan34: Any insight into why this was disabled to begin with?

pchandra:
I had spoken to @caiush about it and I think both of the original patches were failing to apply cleanly (as several upstream things were merged into Ubuntu packages, so these started to fail because of contextual changes). I can verify that this definitely applies cleanly now, and since there's unlikely to be more revs on upstream icehouse packages, it should stay working (and, in theory, this works out-of-the-box on Juno, so hopefully we won't have to do this silliness anymore after we upgrade). I tested this with ephemeral VMs, volume-based VMs, and each of those with extra volumes attached... all worked. So, it should be safe to merge (I've had it merged on my contrail testing branches)

kamidzi:
Okay, could you rebase this against the current HEAD at master?

pchandra:
Just merged master on top so it didn't blow away the conversation above, but should be able to merge again now.

PR #494, opened 2015-02-22
Add hooks to restart radosgw daemon

pchandra:
I needed to add these in order to get the `radosgw` process to come up correctly on new cluster builds. Without it, I was seeing that the package would install it and upstart would try to start it, but then it would die since it wasn't set up yet (directories and keys and such), and then our recipes would never call for a restart. Manual restart cured it, this should just automate that.

### DIFF
--- a/cookbooks/bcpc/recipes/ceph-rgw.rb
+++ b/cookbooks/bcpc/recipes/ceph-rgw.rb
@@ -54,6 +54,7 @@ bash "write-client-radosgw-key" do
         chmod 644 /var/lib/ceph/radosgw/ceph-radosgw.gateway/keyring
     EOH
     not_if "test -f /var/lib/ceph/radosgw/ceph-radosgw.gateway/keyring"
+    notifies :restart, "service[radosgw-all]", :delayed
 end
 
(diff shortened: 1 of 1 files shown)


mihalis68:
This is equivalent to manual steps I have to perform on fresh cluster builds and so from that point of view I can't see any possible risk of merging this.

caiush:
I actually had this in an earlier release, for example: https://github.com/bloomberg/chef-bcpc/commit/81cc4ae9e0606074e8ca84d8a72cf620af32cbc7 but then had to pull it out because it was causing havoc on running clusters when we were reconfiguring ceph.

mihalis68:
Perhaps it would be possible to "start if never yet started"

pchandra:
Yeah, I hadn't thought of the ceph.conf rewrites, but in this PR, I'm not hooking those.... Just the radosgw keyring writing (should only ever happen once on new install). Probably not necessary to hook the rewrite of the cgi file, I just added it for good measure in case it ever changes in the future.

mihalis68:
This also hooks the fcgi script, right? I can't tell if that might fall foul of @caiush's objection or not.

caiush:
@pchandra chatted offline. its fine to merge

PR #495, opened 2015-02-22
Maybe a fix for get_all_nodes

pchandra:
With the code that specifies `search(:node, "recipes:bcpc\\:\\:default ...")`, I would consistently get back an empty set from the search (which broke a bunch of things including giving me rabbit split-brain on new cluster builds). I verified the return using `shef -c /etc/chef/client.rb` on the local machine and also checked that `search(:node, "recipes:bcpc")` would return the right thing. Admittedly, I was testing on an older version of chef (10), but the fix seemed to be to simply specify `bcpc` instead of `bcpc::default`. If this is just an artifact of old chef versions, just close this PR and I'll upgrade my clients in my test setup.

### DIFF
--- a/cookbooks/bcpc/libraries/utils.rb
+++ b/cookbooks/bcpc/libraries/utils.rb
@@ -84,7 +84,7 @@ def search_nodes(key, value)
 end
 
 def get_all_nodes
-    results = search(:node, "recipes:bcpc\\:\\:default AND chef_environment:#{node.chef_environment}")
+    results = search(:node, "recipes:bcpc AND chef_environment:#{node.chef_environment}")
     if results.any? { |x| x['hostname'] == node['hostname'] }
         results.map! { |x| x['hostname'] == node['hostname'] ? node : x }
     else


pchandra:
Ok, I can confirm that `master` is broken right now (using chef client version 10.32.2) without this fix. Basically, `get_all_nodes` always returns a list of 1 member (the node object for the running node) which (amongst other things) causes the `/etc/hosts` file to only have one entry. This causes all references to the machines by name, such as in rabbitmq clustering setup, to fail. So, a three headnode cluster build results in three rabbitmq servers that are unclustered, thus lots of stuff fails. I also tested briefly using chef 11 (11.18 IIRC) and it seemed to do the same.

kkoh61:
I am able to reproduce this on Chef 10 client, though not Chef 11, which master is based on. The difference in behavior seems best explained in https://tickets.opscode.com/browse/CHEF-3645.

Seems harmless to merge this PR in any case.

kamidzi:
As per @kkoh61, this appears to be an artifact of older chef versions.

pchandra:
Actually, can we just merge this anyways since it's semantically equivalent and the bare `bcpc` works on more chef platforms? I'm not so concerned about the older Chef-10 based servers, but this change is needed to fix running via littlechef for instance.

PR #504, opened 2015-03-05
Increase keystone token cache TTL to 7d.

### LINKED ISSUE #501: Increase Keystone token caching TTL
Spoke offline with @caiush. The suggestion is to raise it to 7d.

kamidzi:
Closes #501

### DIFF
--- a/cookbooks/bcpc/templates/default/keystone.conf.erb
+++ b/cookbooks/bcpc/templates/default/keystone.conf.erb
@@ -1292,7 +1292,7 @@ ca_certs = /etc/keystone/cert.pem
 
 # Amount of time a token should remain valid (in seconds).
 # (integer value)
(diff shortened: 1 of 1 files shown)


caiush:
Think you will prob want to up the timeout on horizon, but not sure. Did you test that?

kamidzi:
reply@reply.github.com At: Mar  5 2015 14:42:59" data-digest="From: reply@reply.github.com At: Mar  5 2015 14:42:59" style="">

kamidzi:
Test which part? What's the suspected effect of this timeout on horizon?

kamidzi:
Discussed with @caiush offline and this change should be merged in conjunction with an increase in the Horizon session timeout.

We should just roll the combined changes into the next release.

PR #508, opened 2015-03-08
Override CPU and RAM allocation numbers to be useful on a laptop.

jerenkrantz:
This makes a laptop useful when creating VMs by upping from a CPU/RAM allocation from 2.0/1.0 to 20.0/20.0.


kamidzi:
@caiush ?

caiush:
love it

PR #511, opened 2015-03-09
Use vagrant ssh-config to determine path to private key.

kamidzi:
Closes 506.


kamidzi:
This was tested to be backwards compatible with vagrant 1.4.3.

PR #521, opened 2015-03-18
Modified for handle large number of requests.

ghost:
Modified the ServerLimit to 500 and increased the number of start up server/processes so as to keep new instances from being created with large load. These values can be lowered for simple installs or laptop installs.

### DIFF
--- a/cookbooks/bcpc/templates/default/apache-radosgw.conf.erb
+++ b/cookbooks/bcpc/templates/default/apache-radosgw.conf.erb
@@ -4,6 +4,19 @@
 #
 ################################################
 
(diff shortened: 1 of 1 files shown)


kamidzi:
If this won't necessarily work on a laptop, you might consider some parameterization such as that found in https://github.com/svanzoest-cookbooks/apache2/blob/v3.0.1/attributes/mod_ssl.rb

erhudy:
@cloudm2, can you redo this PR from an existing repo/branch?

cloudm2:
I can redo it. It needs conditions. Thx

erhudy:
Closing PR since it will be superseded.

PR #529, opened 2015-04-06
Add cpufrequtils to basic debugging packages.

jerenkrantz:


erhudy:
One concern: installing the `cpufrequtils` package automatically launches the `loadcpufreq` and `cpufrequtils` init scripts, and I'm not sure what defaults they start up with, but they could potentially enable speed governing on the processors on the hypervisors, which might make some people kind of mad.

Since I'm assuming you just want access to `cpufreq-info`, maybe these services could be disabled in the recipe, or is one or both of them required to make the utilities work correctly?

jerenkrantz:
Thanks - I protected the enablement with an attribute.  Subsequent PRs may add in tweaking of the governor based upon information collected.

FWIW, while `cpufrequtils` is still available on trusty, it looks like `cpufrequtils` has been superseded by `cpupowerutils` - see https://lwn.net/Articles/433002/ and https://launchpad.net/ubuntu/trusty/+source/cpufrequtils/+changelog

erhudy:
Thanks for the info, I'll update unstable/trusty after this merge.

PR :+1:

erhudy:
Closing this pull request since its content was merged into #550.

PR #531, opened 2015-04-07
Ignore IntellIJ project files/directories

erhudy:
adding .idea/ and *.iml to .gitignore so that IntelliJ/RubyMine project files won't be checked in


erhudy:
Nobody really cares, so merged.

PR #532, opened 2015-04-08
Support for custom bcpc metadata

caiush:
A metadata service to provide tenancy information to a VM.

example usage (from within the VM):
```
$  curl http://169.254.169.254/openstack/latest/vendor_data.json
{"users": [{"username": "admin", "uuid": "e96554f436384387868a61a5bfc55abf"}, {"username": "tester", "uuid": "f6ca026c1570449598060901aa813fea"}, {"username": "caius", "uuid": "64e0b81d4b21437b92b38daa5276b188"}], "name": "AdminTenant","uuid": "1e093084ea294ff6b261184ec7818459"}
```
Note, the metadata service caches data for 15s (default) so the actual load this places on the system is pretty minimal.

Note, you can then use this information for setting up per-tenant info on VMs.


jerenkrantz:
LGTM.

erhudy:
:+1:

jerenkrantz:
Upon further reflection, two quick comments:

1) We should probably parameterize this as not all clusters may wish to expose the tenancy information to the VMs.  My hunch is that it should default to off with a setting in the environment to enable, but I can be convinced either way.

2) I actually think that it might be a bit more compact and parse friendly if we switch the output to:

```
{"users": [{"admin" : "e96554f436384387868a61a5bfc55abf"},
                {"tester" : "f6ca026c1570449598060901aa813fea"},
                {"caius" : "64e0b81d4b21437b92b38daa5276b188"}],
"name": "AdminTenant", "uuid": "1e093084ea294ff6b261184ec7818459"}
```

IOW, line 45 goes to something like:

```
            self._data['users'].append( { user.username, user.id } )
```

There may be a more compact way to id the tenancy name, but nothing strikes me as obvious unless we add a nested "tenancy" array or such.

caiush:
I think I should explain my thinking a bit more before I comment on your points. First off, (as you say) i expect different clusters to require different data about tenancies be passed to the VMs, this PR is an example of how to do this. Secondly, I would expect the current returned data to be expanded to include (for example) other information from LDAP servers to enable user auth on the VMs. That in mind:
1) Agreed
2) Disagree, because I would like the ability to add new keys to the user data, for example "ldap_group" or "ssh_key"  or...

With this in mind I add a new suggestion for myself. I think I'll see if I can add a new path to the `sys.path` that is somewhere out of the nova distribution path, e.g. `/usr/local/lib/nova-plugins` that we can dump extensions like this and not muck around in distribution python paths.

pchandra:
Yeah, this looks good in general. I like the suggestion from @caiush to separate this into a separate python package, maybe something like python-bcpc where we can put all such extensions and then adjust the nova.conf to point at that location.

caiush:
redone against a newer master, #558

kamidzi:
Superseded by #558.

PR #533, opened 2015-04-08
EC2-style hostnames (supersedes PR #476)

erhudy:
This pull request is a replacement for PR #476. Please see that pull request for the full details of what this is accomplishing.

tl;dr: this adds EC2-style hostnames (ip-1-127-0-x and public-100-100-100-x) and converts instance names to CNAMEs.


caiush:
I like. Can we merge.

cookbooks/bcpc/recipes/powerdns.rb
```diff
@@ -55,6 +56,26 @@
         mode 00600
         notifies :restart, "service[pdns]", :delayed
     end
+    
+	# if this attribute is true, Chef will back up the PowerDNS database and drop it if
+	# the specified lock file does not exist
+	# (a drop is necessary to facilitate migrating to the bulk fixed IP scheme)
+	if node['bcpc']['enabled']['automate_migration_to_fixed_ip_dns']
+		lock_file = "/usr/local/etc/.already_migrated_to_fixed_ip_dns"
```
kkoh61: If I am reading this correctly, this will be executed more than once if we have powerdns in runlist of > 1 node. Is this expected?

erhudy: Is your concern that this attribute could be set to true on a second node, which would cause it to re-perform the migration even if not needed? That's a fair concern, and the lock file should probably be moved to be an attribute.

erhudy: After discussing with kkoh61 and thinking about it, I am removing this section entirely since it's potentially a lot of automation work for something that would take a human about 30 seconds one time.

kkoh61:
The build raised an exception on my cluster.

NoMethodError: undefined method `run_action' for "service[pdns]":String
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/runner.rb:49:in `run_action'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/runner.rb:111:in `run_delayed_notification'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/runner.rb:99:in `block in run_delayed_notifications'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/runner.rb:98:in `each'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/runner.rb:98:in `run_delayed_notifications'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/runner.rb:88:in `converge'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/client.rb:345:in `converge'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/client.rb:431:in `do_run'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/client.rb:213:in `block in run'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/client.rb:207:in `fork'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/client.rb:207:in `run'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/application.rb:236:in `run_chef_client'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/application/client.rb:338:in `block in run_application'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/application/client.rb:327:in `loop'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/application/client.rb:327:in `run_application'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/lib/chef/application.rb:55:in `run'
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.18.6/bin/chef-client:26:in `<top (required)>'
/usr/bin/chef-client:23:in `load'
/usr/bin/chef-client:23:in `<main>'

The ruby block "add-float-ips" is not executing too. I removed "action:nothing" in the code block to create the records, but noticed subsequently that legitimate records (i.e. CNAMEs) are now missing.

erhudy:
I did encounter the issues above; I think the exception that Chef threw was a symptom of something going wrong in `add-float-ips` rather than a problem in itself.

erhudy:
I tested further against Precise and got the exception to happen once, then it would not re-occur after that. I removed `action :nothing` from `add-float-ips` as well; maybe that was mistakenly left in, or maybe it was originally triggered by another resource that was removed or changed. @caiush may be able to elaborate on that.

However, there appears to be a larger issue where the different queries being run to seed the *records* table are not idempotent and fluctuate in the entries that are created, whereas they should remain static (because they should be pre-creating all the IPs at once).

erhudy:
I'm going to rewrite this entire pull request to fix the lack of idempotence and will re-open when it's behaving properly.

erhudy:
I rewrote the majority of how DNS works, per my commit note, the bullet points of which I reprint here:
- most ruby_block resources have been removed in favor of templates
- powerdns recipe split into two, one that can stand independently of the OpenStack environment and one that requires the nova database to be populated
- INSERT ON DUPLICATE KEY UPDATE is used extensively to simulate idempotence within the records table, so nobody should hit issues now where DNS lookups mysteriously fail if a cron job or Chef is doing something with the table at that time
- removed all sorts of old stuff corresponding to previous incarnations of DNS handling

I added/changed a few additional things as well:
- there is a `bcpc_record_type` column on the `records` table now, which is **STATIC** for any entries inserted from the templates and **DYNAMIC** for any entries inserted by the `dns_fill.py` cron job
- also slightly modified the records table so that the `ttl` column will default to **300**, since that's what we use everywhere, and removed it from all the `INSERT` statements
- fixed the weirdness with Chef throwing a **run_action** exception by changing the format of the `self.notifies` invocation inside **ruby_block** resources (the error was seemingly being thrown whenever PowerDNS was being notified to restart in a delayed fashion)

Scenarios tested:
- repeatedly recheffing a head node both with and without the `pdns` database existing prior
- adding and deleting VM instances and associating/disassociating floating IPs

It is assumed that you will not ever be decreasing the size of the IP ranges, so this doesn't check for that, but you can always just vaporize the contents of the `records` table and rechef.

There is also a very small issue where the SOA records will "walk" to the end of the table each time you run Chef, because the timestamp changes and they are not recognized as duplicates and consequently are re-inserted. I added some logic to scrape out old SOA records so that we don't wind up with a huge archaeological dig of old SOA records, and this doesn't affect things functionally.

erhudy:
I left a little grodiness in the template with some variable assignments in there instead of the recipe, will fix tomorrow. Won't affect the functionality.

erhudy:
Updated pull request with variables put in the proper place.

caiush:
I like it, merge it

PR #534, opened 2015-04-09
firewall rules for monitoring dashboards

kkoh61:
Additional rules for Kibana and Graphite dashboards. Elasticsearch head plugin (port 9200) is excluded as unrestricted access to it would be considered excessive.


erhudy:
:+1:

PR #536, opened 2015-04-09
Disable live migration patch by default 'cos it does not apply cleanly

kkoh61:


erhudy:
:+1:

erhudy:
commonage complained that GIFs are broken in his e-mail client, so: thumbs up

Issue #525, opened 2015-03-20
Incorrect number of connections in mysql config file


caiush:
I think this was a regression. The `max_connections` in the `mysql-head.rb` recipe is set too low and should be 2x of the current value. 

Compare the new mysql-head recipe:
https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/recipes/mysql-head.rb#L64

with the old mysql.rb recipe
https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/recipes/mysql.rb#L88

kamidzi:
Out of curiosity, from where are these scaling factors derived: 110 connections/head node && 10 per all others?

kkoh61:
@pchandra could provide more insight on the calculation, but it seems like a straight doubling to provide additional headroom since we've seen to-be-timed-out connections (another issue to look into) on MySQL exhausting max_connections around the time of VIP jumps.

I am parameterizing max_connections as well in the PR for some operational flexibility.

pchandra:
Yeah, the scaling factors were derived empirically (back in the grizzly days) since I saw ~25 conns from a head and 2 from each worknode, so I made them 50 & 5 to allow for headroom. The later doubling again (to 100 & 10) was due to the storm of reconnects that happens when the VIP jumps (which was exacerbated by various nova processes doing silly things like defaulting to launching a number of instances equal to the number of procs in the machines). Should be totally safe to leave it at the higher 100/10 factors since this is a control-plane only DB.

PR #538, opened 2015-04-10
Parameterize headnodes MySQL max_connections and increase its default

kkoh61:


kkoh61:
Handles #525

erhudy:
Code itself looks fine to me, as long as everyone else is okay with the max connections formula changing from `get_head_nodes.length*50+search_nodes("recipe", "nova-work").length*5` to `get_head_nodes.length*100+get_all_nodes.length*10` (though in practice this change will only impact very small clusters like local testing, so probably NBD).

kkoh61:
The original committer has provided further commentary in #525 so I take it as good to merge.

PR #545, opened 2015-04-14
Remove old mysql recipe

kkoh61:


erhudy:
Did not find any instances of `include_recipe "bcpc::mysql"` in any other recipes or `"recipe[bcpc::mysql]"` in any run list, so this looks safe to merge.

PR #546, opened 2015-04-14
Remove phpmyadmin

kkoh61:
@caiush suggested some time ago to remove phpmyadmin since it's not known to be used.


caiush:
less is more! love it

PR #550, opened 2015-04-18
Cpufreq scaling governor control

erhudy:
This PR is difficult to test because the vCPUs in VirtualBox do not have cpufreq enabled, and thusly the recipe just skips over them. I tested each part in isolation on my desktop outside of Chef and verified that the behavior is as intended (it reads the current and available governors correctly and writes the requested governor mode out).

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -273,6 +273,22 @@ default['bcpc']['routemon']['numfixes'] = 0
 # If set to 0, max_connections for MySQL on heads will default to an
 # auto-calculated value.
 default['bcpc']['mysql-head']['max_connections'] = 0
+
+###########################################
+#
+# CPU governor settings
+#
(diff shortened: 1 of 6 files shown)


erhudy:
This is a followup to #529 (which should also be merged).

jerenkrantz:
If we are going to depend upon this (which is likely a good thing), we
should probably move the installation of the packages to this recipe rather
than the debugging package.

erhudy:
Do you want me to incorporate #529 into this commit and close that PR?

jerenkrantz:
Fine by me!

erhudy:
I validated the behavior of this provider on Trusty (the layout of `/sys/devices/system/cpu` is identical on Precise, so it should function identically there), testing different CPU governor types and also invalid governor types to ensure that Chef would properly bail out.

kkoh61:
Tested OK on Precise. Merging.

PR #551, opened 2015-04-21
Upgrade Kibana to latest stable version

kkoh61:
https://www.elastic.co/downloads/kibana


erhudy:
One minor issue: when accessing through HAProxy, the URL `/kibana` does not work, but `/kibana/` with the trailing slash does work.

erhudy:
Other than that, everything seems fine on Trusty.

kkoh61:
Updated to append slash if it's missing.

erhudy:
Seems fine to me otherwise. If we're not using `/etc/apache2/sites-available/kibana-web` any more, maybe delete that as cleanup. Everything worked fine on Trusty, so I think it's good to merge if everything was happy on Precise.

kkoh61:
Good reminder. Updated PR once more to remove old files in the code base. However, I am not explicitly removing the configuration on deployed systems as I don't believe Kibana is widely enabled.

PR #552, opened 2015-04-21
Parameterize Graphite metrics default retention

kkoh61:
It will likely be desirable to override value in non-VM (larger) deployments.


erhudy:
looks pretty straightforward to me, +1

PR #555, opened 2015-04-27
Making bootstrapping with proxies a bit easier

erhudy:
These are a few uglity hacks to ease working with proxies and also injecting custom CA certs into VMs when bootstrapping. (This PR is not exactly a work of art, but it may help when bootstrapping under certain circumstances.)


kamidzi:
Already on 5.x series.

PR #559, opened 2015-05-06
Toggle for nova debug

kkoh61:


erhudy:
Testing against pre-existing cluster.

erhudy:
Tested against existing nodes with attribute toggled both ways, LGTM.

PR #560, opened 2015-05-07
Rally setup recipes

cloudm2:
Rally has been added and baked into bcpc bootstrap nodes.


cloudm2:
Updating rally so pulling pull request for now.

Issue #561, opened 2015-05-11
ubuntu preseed support of specified ubuntu mirrors not sufficiently general


mihalis68:
The Ubuntu preseed expects the URL for the ubuntu apt mirror to be the ['bcpc']['bootstrap']['mirror'] with /ubuntu appended. This scheme is insufficient to use non-core apt mirrors such as the columbia mirror at mirror.cc.columbia.edu/pub/linux/ubuntu/archive which requires /pub/linux/ubuntu/archive to be appended to mirror.cc.columbia.edu. The net result is difficulty using non-standard (but possibly faster/less rate-limited mirrors) when PXE-booting cluster member nodes. This could be an issue with company-internal mirrors (e.g. centralized mirrors not present on the bootstrap node).

I have the fix for this, just tracking via tickets to provide clarity.

PR #562, opened 2015-05-11
bootstrap apt mirrors

mihalis68:
- Provide more general support of non-default apt mirrors for booting
- Fixes https://github.com/bloomberg/chef-bcpc/issues/561


mihalis68:
@jerenkrantz this extends the bcpc.bootstrap,mirror support you did a couple of years ago, extends it to use nice fast (but different layout) non-core mirrors.

mihalis68:
replaced by https://github.com/bloomberg/chef-bcpc/pull/563

PR #565, opened 2015-05-14
Replace execute resources with mount resource

jeffbyrnes:
The `mount` resource is idempotent, and will only act as-needed.

* Less code
* Doesn’t execute every time
* Just as functional

### DIFF
--- a/cookbooks/bcpc/recipes/ceph-work.rb
+++ b/cookbooks/bcpc/recipes/ceph-work.rb
@@ -80,16 +80,8 @@ ruby_block "reap-ceph-disks-from-dead-servers" do
     end
 end
 
-execute "cephfs-in-fstab" do
-    command <<-EOH
-        echo "-- /mnt fuse.ceph-fuse rw,nosuid,nodev,noexec,noatime,noauto 0 2" >> /etc/fstab
-    EOH
-    not_if "cat /etc/fstab | grep ceph-fuse"
-end
-
-execute "cephfs-mount-fs" do
-    command <<-EOH
-        mount -a
-    EOH
-    not_if "mount | grep ceph-fuse"
+mount 'cephfs' do
+  fstype 'fuse.ceph-fuse'
+  options %w(rw nosuid nodev noexec noatime noauto)
+  action [:mount, :enable]
 end
--- a/cookbooks/bcpc/recipes/ceph-work.rb
+++ b/cookbooks/bcpc/recipes/ceph-work.rb
@@ -81,7 +81,7 @@ ruby_block "reap-ceph-disks-from-dead-servers" do
 end
 
 mount 'cephfs' do
-  fstype 'fuse.ceph-fuse'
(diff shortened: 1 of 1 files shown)


erhudy:
This seems fine to me, anyone else have any comments?

mihalis68:
Yes, good stuff. Any chance you can test it Ned?

erhudy:
I merged it onto my current unstable/5.x branch and got the following failure:
```
  * mount[cephfs] action mount
    
    Error executing action `mount` on resource 'mount[cephfs]'
    
    TypeError
    ---------
    no implicit conversion of nil into String
    
    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/bcpc/recipes/ceph-work.rb
    
     83: mount 'cephfs' do
     84:     fstype 'fuse.ceph-fuse'
     85:     options %w(rw nosuid nodev noexec noatime noauto)
     86:     action [:mount, :enable]
     87: end
    
    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/bcpc/recipes/ceph-work.rb:83:in `from_file'
    
    mount("cephfs") do
      action [:mount, :enable]
      supports {:remount=>false}
      retries 0
      retry_delay 2
      mount_point "cephfs"
      device_type :device
      fsck_device "-"
      fstype "fuse.ceph-fuse"
      options ["rw", "nosuid", "nodev", "noexec", "noatime", "noauto"]
      dump 0
      pass 2
      declared_type :mount
      cookbook_name "bcpc"
      recipe_name "ceph-work"
    end
```

erhudy:
Turns out we don't even install the ceph-fuse package and we don't use CephFS, so these resources are probably good candidates for deletion (want confirmation from @caiush first).

mihalis68:
I am interested in eventually using CephFS. I wouldn't want to make it harder to get to in our stack. It may be a good fit for intra-tenancy file sharing

erhudy:
Since we don't actually use it at present and the resources involved in this pull request look vestigial to me, we might be better served by developing it as an entirely new feature for inclusion in a 5.x release. I'm not sure we'd get any value from retaining these resources as is.

kamidzi:
Seems to have reached a dead-end.

mihalis68:
I wanted to follow-up on this one. @kamidzi closed it as part of a much needed cleanup of dead pull requests. In this case I actually liked the direction, but since we actually do not currently use CephFS, @erhudy's cleanup is more apropos (remove all references to it). If it were to be put back then I think using the mount resource instead of an execute resource would get my vote.

erhudy:
Yes, I agree. The PR was helpful in that it exposed some vestigial resources that were better off cleaned up, but I agree with @mihalis68 that proper mount resources are definitely the way to go.

jeffbyrnes:
Glad I could help, even if it was just to realize that some pruning was necessary.

PR #566, opened 2015-05-15
Added Cgroup capabilities to OSD processes

bcpc:
Added Cgroup packages and configuration to attempt to guarantee some
CPU processing bandwidth to the OSD processes.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -314,6 +314,44 @@ default['bcpc']['cpupower']['governor'] = "ondemand"
 
 ###########################################
 #
+# Cgroups settings
+#
+###########################################
+#
+# These settings configure the CPU and RAM priority and weight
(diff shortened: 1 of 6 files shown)


erhudy:
Deploying to hardware cluster for testing.

erhudy:
This doesn't appear to work as is:
* neither the cgconfig nor cgred service launch (both terminate with errors, cgconfig in `/var/log/syslog` and cgred in /var/log/upstart/cgred.log`
* the upstart scripts do not appear to be functional and it looks almost like $ENVVAR references were stripped from both of them
* cgconfigparser reports `cgconfigparser; error loading /etc/cgconfig.conf: This kernel does not support this feature` when attempting to manually parse `/etc/cgconfig.conf` (this seems clearly wrong)

erhudy:
@jnegron11, please recreate this PR from your fork so we can proceed with debugging the above issues and get this feature merged.

PR #570, opened 2015-05-17
Add support for bootstrap-specific apt mirror when using Vagrant

kamidzi:
Leverage an existing apt mirror for a marginally faster build.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -12,7 +12,6 @@ if $local_mirror.nil?
   $repos_script = <<EOH
 EOH
 else
-  puts "Configuring inline provisioner with local apt mirror for bootstrap"
   $repos_script = <<EOH
 #!/bin/bash
 hash -r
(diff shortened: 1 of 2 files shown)


erhudy:
Tried the branch out and found that for whatever reason it didn't take right away, as when `bootstrap_chef.sh` ran the output from apt was still for the main Ubuntu archive servers. Maybe it just needs another apt-get update to be run first? I didn't investigate too closely.

It did take effect after that, as the later apt-get update failed to work and bailed out because I had defined an HTTP proxy and so apt was trying to route its requests through that. I think that can be solved pretty easily by adding another line to the script to add an opt-out for the local mirror so that apt doesn't try to go via a proxy server, something like:

```
echo Acquire::http::Proxy::#{local_mirror.split(':')[0]} "DIRECT"; | tee -a /etc/apt/apt.conf
```

(I didn't test that but it should work for parsing the IP address out of an IP:port pair and adding the appropriate line to `/etc/apt/apt.conf`.)

erhudy:
Just pushed a commit that fixes the above complaint and resolves the HTTP proxy issue.

PR #575, opened 2015-05-18
Bugfix for #572 - increases default conductor workers to 5

### LINKED ISSUE #572: Consider upping default nova-conductor workers from 2 to 5
In testing locally with 1 head node and 2 worker nodes, `nova-conductor` could not keep up with the number of incoming messages with the default value of workers (currently 2).  Upping it to 5 helped dramatically.

```
default['bcpc']['nova']['workers'] = 5
```

(If we agree, I can submit a PR or someone can just push a fix.)

erhudy:
Fixes #572.


jerenkrantz:
LGTM.

PR #576, opened 2015-05-18
In Kilo, glance settings must now be in a separate config section.

jerenkrantz:
(Otherwise, VMs can't launch on worker nodes as cinder tries to talk to
glance on the local IP - which doesn't exist.)


erhudy:
This looks like it specifically fixes the problem where image-based instances wouldn't always launch correctly (creating a volume from an image would still work). Tested and LGTM.

PR #580, opened 2015-05-18
Fixes #573 - patches Nova networking to not use the same DHCP server

### LINKED ISSUE #573: Multiple tenants broken in 5.x due to dhcp_server nova_network bug
In local tests w/Trusty+Kilo, `nova-network` creates the `dhcp_server` setting with an invalid value - it's in the wrong subnet.  This causes machines in a non-default tenancy not to receive its DHCP address as the dnsmasq's `listen-address` is what the associated bridge is listening on (e.g., `br1001` should listen on 1.127.1.1, but is listening instead on 1.127.0.1).  Chaos ensues.

I have not fully tracked down where the specific issue is - it's likely in the `nova-manage` call that we have in `nova-setup.rb`:

```
nova-manage network create --label fixed --fixed_range_v4=#{node['bcpc']['fixed']['cidr']} --num_networks=#{node['bcpc']['fixed']['num_networks']} --multi_host=T --network_size=#{node['bcpc']['fixed']['network_size']} --vlan_start=#{node['bcpc']['fixed']['vlan_start']} --bridge_interface=#{node['bcpc']['fixed']['vlan_interface']}
```

I will bet that the `fixed_range_v4` code path sets `gateway` properly, but doesn't update `dhcp_server` accordingly.

IOW, the `dhcp_server` value is like:

```
mysql> select id, gateway, dhcp_server from networks where id=2;
+----+-----------+-------------+
| id | gateway   | dhcp_server |
+----+-----------+-------------+
|  2 | 1.127.1.1 | 1.127.0.1   |
+----+-----------+-------------+
1 row in set (0.00 sec)
```

The hacky workaround is:

```
mysql> update networks set dhcp_server='1.127.1.1' where id=2;
Query OK, 1 row affected (0.00 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> select id, gateway, dhcp_server from networks where id=2;
+----+-----------+-------------+
| id | gateway   | dhcp_server |
+----+-----------+-------------+
|  2 | 1.127.1.1 | 1.127.1.1   |
+----+-----------+-------------+
1 row in set (0.01 sec)
```

Clearing out the bridges, killing `dnsmasq`, and restarting `nova-network` after that change suffices.

erhudy:
This fixes bug #573, where when creating multiple subnets via nova-manage, the DHCP server for all of them would be the gateway IP for the first subnet in the range.

### DIFF
--- a/cookbooks/bcpc/files/default/nova-network-dhcp-server.patch
+++ b/cookbooks/bcpc/files/default/nova-network-dhcp-server.patch
@@ -1,0 +1,24 @@
+diff --git a/nova/network/manager.py b/nova/network/manager.py
+index 3e8e8b1..832fd1b 100644
+--- a/nova/network/manager.py
++++ b/nova/network/manager.py
+@@ -1351,13 +1351,15 @@ class NetworkManager(manager.Manager):
+                 else:
+                     net.gateway = current
+                     current += 1
+-                if not dhcp_server:
+-                    dhcp_server = net.gateway
++                if dhcp_server:
++                    subnet_dhcp_server = dhcp_server
++                else:
++                    subnet_dhcp_server = net.gateway
+                 net.dhcp_start = current
+                 current += 1
+-                if str(net.dhcp_start) == dhcp_server:
++                if str(net.dhcp_start) == subnet_dhcp_server:
+                     net.dhcp_start = current
+-                net.dhcp_server = dhcp_server
++                net.dhcp_server = subnet_dhcp_server
+                 extra_reserved.append(str(net.dhcp_server))
+                 extra_reserved.append(str(net.gateway))
+
--- a/cookbooks/bcpc/recipes/nova-setup.rb
+++ b/cookbooks/bcpc/recipes/nova-setup.rb
@@ -20,6 +20,37 @@
 include_recipe "bcpc::keystone"
 include_recipe "bcpc::nova-head"
 
+#  _   _  ____ _  __   __  ____   _  _____ ____ _   _
+# | | | |/ ___| | \ \ / / |  _ \ / \|_   _/ ___| | | |
+# | | | | |  _| |  \ V /  | |_) / _ \ | || |   | |_| |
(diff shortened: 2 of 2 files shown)


erhudy:
I did not write any resources to try to migrate from the old broken setup to the new fixed one, because they could be potentially dangerous.

**Fixing the MySQL way (recommended)**
Within MySQL, run `UPDATE nova.networks SET dhcp_server = gateway;`. I don't think it's required to relaunch any OpenStack processes, but probably for safety it wouldn't hurt to HUP everything.

**Fixing the Nova way**
To clean out networks that are not associated with any project, use `nova-manage network list 2>/dev/null | tail -n +2 | awk '{print $2}' | xargs -I% nova-manage network delete %`. This will error out on networks that are already associated with a project. You can then disassociate the networks from the projects with `nova network-disassociate UUID`, terminate instances using those networks, delete the remaining networks, and re-chef to regenerate the network list.

erhudy:
Here is a one-liner script to do the whole thing via MySQL. If people are comfortable with it, I can incorporate it into a resource to fix existing clusters in-place:
```
export MYSQL_PASSWORD=xxxxxxxxxxxxxxxxxxxxx; if [[ $(mysql --skip-column-names -s -uroot -p$MYSQL_PASSWORD -e 'SELECT count(*) FROM nova.networks WHERE dhcp_server != gateway' nova) != 0 ]]; then mysql -uroot -p$MYSQL_PASSWORD -e 'UPDATE nova.networks SET dhcp_server = gateway'; fi
```

jerenkrantz:
I doubt anyone is using 5.x right now in production.  As long as new clusters going forward are fine, we can point people to this issue for the workaround.

erhudy:
I've tested the fix on hardware clusters with several tenants and it seems to be fine, so if somebody else can confirm it seems to work for them, I think this can be merged.

jerenkrantz:
LGTM.

PR #584, opened 2015-05-18
changing default domain for Test-Laptop to local.lan

erhudy:
This is a trivial change and one I think we all agreed on making, but probably forgot to change in master because we had all edited our own environments locally.


kamidzi:
I believe we should stick to the example domain, as it is compliant with https://tools.ietf.org/html/rfc6761. We should fix where this is problematic in the infrastructure instead.

erhudy:
Closing this PR because I fixed the underlying broken behavior in the revised Vagrant build process.

PR #586, opened 2015-05-19
we don't actually use CephFS, so removing these resources

erhudy:
(this was discovered in PR #565)


erhudy:
Added another resource to remove the junk from `/etc/fstab` as well, which tested out fine for idempotence and also for won't-destroy-fstab-ence.

kelvk:
LGTM

mihalis68:
LGTM too

PR #587, opened 2015-05-19
Clean up Chef run output

erhudy:
This PR eliminates a lot of extraneous noise during Chef runs:
* anywhere `puts` was used has been replaced by a call to `Chef::Log` and an appropriate log level (usually **info**)
* invocations of `mysql` now receive their password via the `MYSQL_PWD` environment variable rather than the `-p` switch, which suppresses the warning about using a password on the command line
* guards that were created as `if` or `if not` statements inside a `ruby_block` resource type have been moved out of the main block into an `only_if` or `not_if` guard
* command stdout and/or stderr was redirected to `/dev/null` in certain cases where the command is known to always output something that isn't needed (mostly used in cases where `nova-manage` is invoked to suppress the complaint about `oslo_config.log`)
* added a guard on `load-virsh-keys`
* rewrote `wait-for-mon-quorum` to have a timeout (`ruby_block` resources do not have a `:timeout` attribute of their own)

### DIFF
--- a/cookbooks/bcpc/libraries/utils.rb
+++ b/cookbooks/bcpc/libraries/utils.rb
@@ -29,7 +29,7 @@ end
 
 def init_config
     if not Chef::DataBag.list.key?('configs')
-        puts "************ Creating data_bag \"configs\""
+        Chef::Log.info("************ Creating data_bag \"configs\"")
         bag = Chef::DataBag.new
         bag.name("configs")
         bag.save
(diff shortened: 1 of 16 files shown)


erhudy:
I have tested this against already built bare-metal and virtualized clusters, and also built a new virtualized cluster (head/work/monitoring) node off this branch.

erhudy:
I missed a few database-related things on the monitoring role recipes yesterday, so those are added now. Tested both as building a fresh node and recheffing an existing monitoring node.

erhudy:
One little annoyance to note: on the very first run, since the guard on the initial MySQL setup for monitoring/head roles has a password in it and there is initially no password set, the guard will print to stderr
`ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)`. This is fine because the guard still fails (as it should) and the database setup runs, and the message disappears on subsequent runs when the password is set. If we really want to get rid of this error, we can add the stderr->stdout redirects.

PR #588, opened 2015-05-19
Destroys nova-service-restart-wrapper and salts the earth

### LINKED ISSUE #578: Make nova-service-restart-wrapper a tunable option
In debugging clusters, having `nova-service-restart-wrapper` run every five minutes is *really* annoying.  We should have a tunable to enable or disable it.

I'd recommend leaving it off - but, at the very least, we need an easy way to skip it.

(My general philosophy is to try to fix the underlying issues rather than applying crude sledgehammers.)

erhudy:
This resolves #578, with fire.


erhudy:
Tested out fine on existing clusters.

PR #589, opened 2015-05-20
Upgrading Cirros to 0.3.4 to stay current

erhudy:
This PR upgrades Cirros to the most current version, 0.3.4. I deployed it on test hardware and everything seems fine. Checks pass, can launch instances from image or volume with it, assign IP, inject keys, everything you'd expect.

To deploy to an existing cluster, rerun `cookbooks/bcpc/files/default/build_bins.sh` and then upload the cookbook.


erhudy:
Tested without issue.

PR #591, opened 2015-05-20
Fixes monitoring role requiring two runs to set up

erhudy:
This moves the lookup of monitoring nodes from compile time to runtime. The first time you run the monitoring recipe, the node is not registered with the Chef server as actually being of the **BCPC-Monitoring** role until after everything is compiled and convergence begins (note that this is different from setting the run list of the node to include `role[BCPC-Monitoring]`), and so the call to `search_nodes` would return an empty list, which would cause HAProxy to not have any backends configured for MySQL, which caused Graphite database setup to explode. Moving the node lookup to a utility function means that it's looked up at runtime after convergence has begun, and so the node will be able to find itself on the first run and configure HAProxy appropriately.

### DIFF
--- a/cookbooks/bcpc/libraries/utils.rb
+++ b/cookbooks/bcpc/libraries/utils.rb
@@ -119,6 +119,15 @@ def get_head_nodes
     return results.sort! { |a, b| a['hostname'] <=> b['hostname'] }
 end
 
+def get_monitoring_nodes
+    results = search(:node, "role:BCPC-Monitoring AND chef_environment:#{node.chef_environment}")
+    results.map! { |x| x['hostname'] == node['hostname'] ? node : x }
+    if not results.include?(node) and node.run_list.roles.include?('BCPC-Monitoring')
+        results.push(node)
+    end
+    return results.sort! { |a, b| a['hostname'] <=> b['hostname'] }
+end
+
 def get_bootstrap_node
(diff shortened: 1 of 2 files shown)


erhudy:
Tested on existing clusters without issue.

kelvk:
How about patching search_nodes, like so:
https://github.com/kelvk/chef-bcpc/commit/eefc411f6b4c29011cf2383d08f4e0549e7f6993

Tested OK on existing cluster.

kelvk:
@erhudy , I've updated my branch some more to extend this to recipe search as well - https://github.com/kelvk/chef-bcpc/commit/836fd978de308c798ff59fe9e4336d49e976fbd5. Setup went well with a single chef-run. If agreeable, you can either pull/modify where required or I will simply make another PR.

erhudy:
I cherry picked your commits but it failed to compile. Can you just push your changes onto this branch to update this PR?

kelvk:
Done. I wanted to re-test this again but my bootstrapping is currently failing due to other reasons. If you could test this again I'd be grateful.

erhudy:
Tested on an existing cluster by uploading the changes and re-cheffing. Nothing got mad. I'd like to test this from a fresh cluster build as well.

kelvk:
Tested OK on fresh cluster too.

mihalis68:
Based on eyeballing the code and the evidence of two team members independently building clusters with this (one rechefed, one rebuilt), this LGTM.

PR #594, opened 2015-05-20
Moving a few envvars from bootstrap into an ignored file

erhudy:
This PR exists mainly to make it a touch easier to deal with overriding certain variables used during the bootstrap process (primarily `BOOTSTRAP_APT_MIRROR`).


vbox_create.sh
```diff
@@ -265,8 +263,8 @@ function create_cluster_VMs {
       if ! $VBM list vms | grep "^\"${vm}\"" ; then
           $VBM createvm --name $vm --ostype Ubuntu_64 --basefolder $P --register
           $VBM modifyvm $vm --memory $CLUSTER_VM_MEM
-          $VBM modifyvm $vm --cpus $CLUSTER_VM_CPUs
-          $VBM modifyvm $vm --vram $CLUSTER_VM_VRAM
+          $VBM modifyvm $vm --cpus $CLUSTER_VM_CPUS
+          $VBM modifyvm $vm --vram 16
```
kamidzi: Why take away this option?

erhudy: I originally made the change to set VRAM explicitly to 16MB to stop VirtualBox from saying the configuration of the VM was invalid when I went into the settings panel for the VM. Since this change was made only to stop VirtualBox from getting angry and the amount of VRAM doesn't have an impact on whether the VM itself can do the job it needs to do, I decided it was fine to set it statically at 16MB across the board.

erhudy:
Retested with Kundai's changes without issue, merging.

PR #595, opened 2015-05-22
Redefine and parameterize nova scheduler filters

kelvk:


erhudy:
For reference, these are the default configuration filters as specified by the Kilo documentation (http://docs.openstack.org/kilo/config-reference/content/list-of-compute-config-options.html):
`scheduler_default_filters = RetryFilter, AvailabilityZoneFilter, RamFilter, ComputeFilter, ComputeCapabilitiesFilter, ImagePropertiesFilter, ServerGroupAntiAffinityFilter, ServerGroupAffinityFilter`

This proposed change would add:
* **AggregateInstanceExtraSpecsFilter**
* **DifferentHostFilter** 
* **SameHostFilter**

...while removing: 
* **ComputeCapabilitiesFilter**
* **ImagePropertiesFilter**
* **RetryFilter**
* **ServerGroupAffinityFilter**
* **ServerGroupAntiAffinityFilter**

erhudy:
@kelvk can you clarify the reasoning behind the add/remove changes listed above, just so that I understand the rationale?

kelvk:
The scheduler defaults filters in this PR are what we have internally today, but never committed to upstream. The AggregateInstanceExtraSpecs filter enables use of host aggregates (i.e. maintenance group for hypervisors), while Different/Same host filters give tenants more control over scheduling.

We've not had to use other filters. Parameterizing this provides flexibility should our requirements change. If we prefer to keep attributes in sync with Openstack's defaults, I can provide overrides via the environment.

erhudy:
With this change merged, I was able to launch a pile of instances without Nova getting angry at me.

PR #604, opened 2015-05-22
Fixes #596 - pre-creates .rgw.buckets.extra for RADOSgw

### LINKED ISSUE #596: S3 multipart uploads over radosgw broken in 5.x
With 5.x, I am seeing timeout failures in multipart uploads against `radosgw` over S3 via boto.

`radosgw` does not seem to be responding to the initial request to initiate the multipart upload.  An HTTP request like:

```
POST /bosh-blobstore/concourse-0.50.0.tgz?uploads HTTP/1.1
Host: 192.168.100.5
Accept-Encoding: identity
Date: Fri, 22 May 2015 14:40:30 GMT
Content-Length: 0
Authorization: AWS 2U0B5RZU86DTQGBKTDC1:I+pxTghji29oWtpjHma84I2whFE=
User-Agent: Boto/2.8.0 (darwin)

```

times out with a 500 error.

It should respond with an XML response container of InitiatedMultipartUploadsResult per http://ceph.com/docs/v0.80.5/radosgw/s3/objectops/#initiate-multi-part-upload

The s3put.py from boto needs some hacks to reproduce it.  If it is hard for others to reproduce, I can post a cleaned up version of my s3put.py somewhere.

(FWIW, I first triggered this via setting BOSH to use an S3 endpoint.)

erhudy:
RADOSgw uses .rgw.buckets.extra to store metadata when accepting multipart file uploads. If this pool is not already created, it tries to create it with crush_ruleset 0, size 3, min_size 2, and so the PGs will never go active because crush_ruleset 0 is not one that exists in our setup. This adds that pool to the list of pools to be pre-created.

### DIFF
--- a/cookbooks/bcpc/recipes/ceph-rgw.rb
+++ b/cookbooks/bcpc/recipes/ceph-rgw.rb
@@ -62,7 +62,7 @@ rgw_optimal_pg = power_of_2(get_ceph_osd_nodes.length*node['bcpc']['ceph']['pgs_
 
 rgw_crush_ruleset = (node['bcpc']['ceph']['rgw']['type'] == "ssd") ? node['bcpc']['ceph']['ssd']['ruleset'] : node['bcpc']['ceph']['hdd']['ruleset']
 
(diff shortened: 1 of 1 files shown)


jerenkrantz:
LGTM.

PR #611, opened 2015-05-27
Add support for custom flavors

caiush:
Ad configured this will delete all flavors installed by default (if present) and create a single new one `b1.tiny`. It is expected that any real cluster would add its own flavors in.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -392,7 +392,4 @@ default['bcpc']['rally']['user'] = 'ubuntu'
 ###########################################
 
 default['bcpc']['flavors']['deleted'] = []
-default['bcpc']['flavors']['enabled'] = { 
-  "b1.tiny" => { "vcpus" => 1,
-                 "disk_gb" => 0}
-}
+default['bcpc']['flavors']['enabled'] = {}
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -385,7 +385,6 @@ default['bcpc']['bootstrap']['mirror_path'] = "/ubuntu"
 # None needed at this time
 default['bcpc']['rally']['user'] = 'ubuntu'
 
-
 ###########################################
 #
 # Openstack Flavors
@@ -397,6 +396,3 @@ default['bcpc']['flavors']['enabled'] = {
   "b1.tiny" => { "vcpus" => 1,
                  "disk_gb" => 0}
 }
-                 
-                                         
-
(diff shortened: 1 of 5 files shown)


jerenkrantz:
I am very much against doing this by default.  A lot of public OpenStack examples rely upon `m1` flavors.  Creating new flavor names just for the sake of it is likely to be very counter-productive.

I would be okay with redefining the built-in flavors by default to alternate values (for example, that may better suit a laptop, for example), but deleting them and putting in new names out-of-the-box does not sit well.

cloudm2:
I know that Rally and Tempest look for default values for flavors among other default items that I have seen. I sure we can add the ability (if not present somewhere) to support it but a number of tools default to the "out of the box" flavors.

caiush:
modification of flavours actually, under the hood, deletes and remakes so you can;t actually modify flavors, (short of fudging the db).

caiush:
I removed the removal. Honestly dont really care. The import point is that one can add new flavors to clusters without going in an bumbling around with some hinky cli. Lots of clusters have different flavors and not having that in chef is dangerous. 

I would be shocked if testing tools dont allow you to specify flavors, but perhaps I should know better than to be surprised by now.

kamidzi:
I think this is good, as well as using UUIDs for ids, but we run into a consistency problem when the default flavors remain:
```
$ nova flavor-list
+--------------------------------------+-----------+-----------+------+-----------+------+-------+-------------+-----------+
| ID                                   | Name      | Memory_MB | Disk | Ephemeral | Swap | VCPUs | RXTX_Factor | Is_Public |
+--------------------------------------+-----------+-----------+------+-----------+------+-------+-------------+-----------+
| 1                                    | m1.tiny   | 512       | 1    | 0         |      | 1     | 1.0         | True      |
| 2                                    | m1.small  | 2048      | 20   | 0         |      | 1     | 1.0         | True      |
| 3                                    | m1.medium | 4096      | 40   | 0         |      | 2     | 1.0         | True      |
| 4                                    | m1.large  | 8192      | 80   | 0         |      | 4     | 1.0         | True      |
| 5                                    | m1.xlarge | 16384     | 160  | 0         |      | 8     | 1.0         | True      |
| bbfdd3c5-2432-4b02-ae06-600715f67bb7 | b1.tiny   | 512       | 0    | 0         |      | 1     | 1.0         | True      |
+--------------------------------------+-----------+-----------+------+-----------+------+-------+-------------+-----------+
```

mihalis68:
osflavor.rb has a wrong title in its comment block, identifies itself as cephconfig.

Also, I think the attributes introduced for this support an optional id field, but then the invocation of nova flavor-create never uses it as far as I can tell.

mihalis68:
reviewed again. Now LGTM

PR #612, opened 2015-05-27
Default quotas under control of chef

caiush:
Configure default quotas through chef.


caiush:
Should have said.. this is an easy upgrade, just chef it in. It will modify all "Standard" project quotas.

kamidzi:
This looks good, but needs to be rebased to merge cleanly.

PR #613, opened 2015-06-01
Relocate zabbix server to monitoring role

kelvk:
Other monitoring solutions may deprecate this in future but a basic and functional availability and alerting solution is needed to provide visibility, at least in the short-term. Therefore, this is purely a move of Zabbix server from head nodes to the separate monitoring role and enabling agents across the cluster.


kamidzi:
Strangely, I hit the following:
```
Recipe Compile Error in /var/chef/cache/cookbooks/bcpc/recipes/zabbix-agent.rb

ArgumentError
-------------
wrong number of arguments (0 for 1..3)

Cookbook Trace:
---------------
  /var/chef/cache/cookbooks/bcpc/libraries/utils.rb:84:in `search_nodes'
  /var/chef/cache/cookbooks/bcpc/recipes/zabbix-agent.rb:71:in `block in from_file'
  /var/chef/cache/cookbooks/bcpc/recipes/zabbix-agent.rb:66:in `from_file'
```

I'll investigate further

erhudy:
Above problem fixed in #617.

PR #617, opened 2015-06-02
Fixes a failure that would occur on the initial Chef of a node

erhudy:
This fixes a problem that was introduced by a change to `search_nodes` in #591. On the first Chef run on a node, the call to `results.include?(node)` would not short-circuit the entire comparison and Chef would evaluate `node.run_list.expand.recipes.include?("bcpc::#{value}")`. `expand` requires at least one parameter (the Chef environment), and so compilation would fail. However, the Chef server would then be updated with the run list and any subsequent Chef runs would short circuit on the comparison and never reach the call to the `expand` method.


kelvk:
The problem appeared masked when the second chef-run was automatically invoked during the build process. Having removed the second chef run, I could see the failure more clearly and it is reproducible. Fwiw, tested this OK, but another pair of eyes will be great since this was not spotted in #591!

PR #618, opened 2015-06-02
Release 5 non vagrant

mihalis68:
Fix up the non-vagrant build path following the move to Trusty/Kilo. 

Also add another check to cluster-check.sh to spot a nasty issue that cropped up in such a build where default routes in table main disappeared and led to very odd behavior.

### DIFF
--- a/chefit.sh
+++ b/chefit.sh
@@ -24,19 +24,25 @@ if [[ -z "$CHEF" ]]; then
     $SCPCMD finish-head.sh    /home/ubuntu
 
     if [[ -n "$(source proxy_setup.sh >/dev/null; echo $PROXY)" ]]; then
-	PROXY=$(source proxy_setup.sh >/dev/null; echo $PROXY)
-	echo "setting up .wgetrc's to $PROXY"
-	$SSHCMD "echo \"http_proxy = http://$PROXY\" > .wgetrc"
+	    PROXY=$(source proxy_setup.sh >/dev/null; echo $PROXY)
+	    echo "setting up .wgetrc's to $PROXY"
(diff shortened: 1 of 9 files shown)


chefit.sh
```diff
@@ -24,19 +24,25 @@ if [[ -z "$CHEF" ]]; then
...
-	PROXY=$(source proxy_setup.sh >/dev/null; echo $PROXY)
-	echo "setting up .wgetrc's to $PROXY"
-	$SSHCMD "echo \"http_proxy = http://$PROXY\" > .wgetrc"
+	    PROXY=$(source proxy_setup.sh >/dev/null; echo $PROXY)
+	    echo "setting up .wgetrc's to $PROXY"
+	    $SSHCMD "echo \"http_proxy = http://$PROXY\" > .wgetrc"
+
+        # possibly set up a proxy for apt too
+        if [[ -n "$APTPROXY" ]]; then
+	        echo "Acquire::http::Proxy \"http://${APTPROXY}\";" > /tmp/apt.conf
```
kamidzi: Is this intentional to blow away whatever else might have been configured for APT?

erhudy: This also won't work if you're combining a proxy configuration with a local mirror that isn't accessible via the proxy, because apt will need an `Acquire::http::Proxy::x.x.x.x "DIRECT";` entry (where `x.x.x.x` is the hostname or IP address of the local mirror).

mihalis68: That is a fair point, however this is only for the pre-Chef stage. Once we can get Chef to run we'll configure things from Chef recipes. I need a little bit of apt goodness before I can run Chef since we zap the ceph disks first.

PR #626, opened 2015-06-09
Hammer Release

cloudm2:


cookbooks/bcpc/attributes/default.rb
```diff
@@ -190,6 +206,7 @@
 default['bcpc']['repos']['gridcentric'] = "http://downloads.gridcentric.com/packages/%s/%s/ubuntu"
 default['bcpc']['repos']['elasticsearch'] = "http://packages.elasticsearch.org/elasticsearch/1.5/debian"
 default['bcpc']['repos']['erlang'] = "http://packages.erlang-solutions.com/ubuntu"
+default['bcpc']['repos']['ceph'] = "http://ceph.com/debian-hammer"
```
erhudy: Do we want to use the standalone Ceph repo instead of Canonical's packages from the Ubuntu cloud archive?

environments/Test-Laptop.json
```diff
@@ -21,7 +21,7 @@
         "cpu_allocation_ratio": "20.0",
         "nova_single_cpu": true
       },
-      "domain_name" : "bcpc.example.com",
+      "domain_name" : "bcpc.local.lan",
```
erhudy: This change shouldn't be part of this PR.

cookbooks/bcpc/attributes/default.rb
cloudm2: It does not use the default repo. In ceph-common.rb it creates the ceph.list in sources for the ceph repo because it's the only way to get the latest version.

environments/Test-Laptop.json
cloudm2: Residual left over. I'll correct

cookbooks/bcpc/attributes/default.rb
erhudy: The Ubuntu cloud archive provides both Firefly and Hammer; the reason I added the version pin for Ceph was because it was defaulting to installing the 0.94.1 packages. Or do you mean it's newer in terms of bugfixes/errata? I can understand that if that's the case.

cloudm2: When I simply used the new version info (assume ubuntu cloud) it didn't like it and ceph version is always updated first. Anyway, if it needs to change then exactly what and where I will be happy to do so.

cloudm2: Also, don't want to pin anything and sometimes vendor repo is not as up to date on all of the dependencies etc.

mihalis68:
As discussed internally : When chef-bcpc moved to Trusty and Kilo we stopped using ceph.com to get the ceph binaries and dropped ceph-extras (at least from the mirrors I make for disconnected cluster building in the data center). We get ceph from the Canonical Trusty repository.

mihalis68:
This looks good to me now

caiush:
I'll say in on behalf of contributors that are out. The commit history here is a bit rough. Not a reason to hold up pr, but worth just taking note of

caiush:
Other than that, huzzah hammer! Nice

cloudm2:
Yep, commits would look better if squashed.

erhudy:
Suffered the following failure when recheffing the head node:
```
  * ruby_block[initialize-ceph-common-config] action run
    - execute the ruby block initialize-ceph-common-config
  * ruby_block[write-ceph-mon-key] action run (skipped due to not_if)
  * template[/etc/ceph/ceph.conf] action create
    - update content in file /etc/ceph/ceph.conf from a57326 to fd903d
    --- /etc/ceph/ceph.conf	2015-06-10 23:24:07.955368346 +0000
    +++ /tmp/chef-rendered-template20150610-17597-11tf1vf	2015-06-10 23:47:43.873672353 +0000
    @@ -48,5 +48,11 @@
         rgw frontends = civetweb port=8088
         debug rgw = 0/0
     
    +    rgw keystone url = 10.0.100.5:35358
    +    rgw keystone admin token = ACcmWf4zkVo7G3sTjI4X
    +    rgw keystone accepted roles = admin Member _member_
    +    rgw keystone token cache size = 1000
    +    rgw keystone revocation interval = 1200
    +    rgw s3 auth use keystone = true
         rgw dns name = s3.bcpc.example.com
  * bcpc_cephconfig[paxos_propose_interval] action set
    
    Error executing action `set` on resource 'bcpc_cephconfig[paxos_propose_interval]'
    
    RuntimeError
    ------------
    Ceph target "/var/run/ceph//ceph-client.radosgw.gateway.9958.139873634926992.asok" doesn't have the config value paxos_propose_interval
    
    Cookbook Trace:
    ---------------
    /var/chef/cache/cookbooks/bcpc/providers/cephconfig.rb:47:in `block (2 levels) in class_from_file'
    /var/chef/cache/cookbooks/bcpc/providers/cephconfig.rb:28:in `each'
    /var/chef/cache/cookbooks/bcpc/providers/cephconfig.rb:28:in `block in class_from_file'
    
    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/bcpc/recipes/ceph-common.rb
    
     86: bcpc_cephconfig 'paxos_propose_interval' do  
     87:   value node["bcpc"]["ceph"]["rebalance"] ? "60" : "1"
     88:   target "ceph-*"
     89: end
     90: 
    
    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/bcpc/recipes/ceph-common.rb:86:in `from_file'
    
    bcpc_cephconfig("paxos_propose_interval") do
      action :set
      retries 0
      retry_delay 2
      declared_type :bcpc_cephconfig
      cookbook_name "bcpc"
      recipe_name "ceph-common"
      value "1"
      target "ceph-*"
      path "/var/run/ceph/"
    end
```

cloudm2:
I have seen this in two cases: 1. Existing and running ceph-mon etc 2. Existing ceph directories. That provider looks for certain things I believe. Remove those things and do again or start the process again. The recipes are not idempotent like they should be.

cloudm2:
Updated the commits and squashed down.

erhudy:
It appears that the provider was breaking because of the regex it was using to parse Ceph's output. Since Ceph output is JSON, I rewrote the provider to use the JSON parser instead of the more brittle regex.

Since this PR is coming from your fork rather than this repo and I can't update it, I've pasted the revised provider in a Gist here: https://gist.github.com/erhudy/6fe13d5b8ff218f83994

erhudy:
After updating the provider, I'm able to Chef everything up.

erhudy:
RADOSgw tests need updated:
```
vagrant@bcpc-vm1:~$ sudo check
+-----------+-------+--------+------------------------------------------------+
|    name   | state |  time  |                      msg                       |
+-----------+-------+--------+------------------------------------------------+
|   apache  | ERROR | 0.202  |            Status rgw returned 500             |
|           |       |        |                                                |
|    nova   |  OKAY | 22.386 |                                                |
| float_ips |  OKAY | 1.306  |                                                |
|   mysql   |  OKAY | 0.203  |                                                |
|    rgw    | ERROR | 69.147 |    BotoServerError: 503 Service Unavailable    |
|           |       |        |  <html><body><h1>503 Service Unavailable</h1>  |
|           |       |        | No server is available to handle this request. |
|           |       |        |                 </body></html>                 |
|           |       |        |                                                |
+-----------+-------+--------+------------------------------------------------+
```

cloudm2:
Just the port needs to change. I left the default of the haproxy rgw port alone. At THIS time both apache and civetweb are installed with civetweb answering on 8088 instead of 80 which the radosgw_check sets it to (based on the haproxy rgw port). After testing and comparing we can make a final change. I will change the port (I just did it manually for now).

I will also apply the gist for the provider. If the environment is perfect the old provider worked but put in any variances it would have heartburn.

erhudy:
After updating from the new changes and building a fresh local cluster, everything worked first time. I will probably throw this on an existing cluster tomorrow to see how badly it goes, but upgradeability is probably not a precondition for this.

erhudy:
I'm still having issues with the RADOSgw checks. I think that it is likely problematic that it has been removed from Apache, because Apache was used for SSL termination, and Civetweb does not do SSL.

cloudm2:
Yep on SSL. Both Apache config for radosgw and civetweb are present so for testing we can simply change the ceph.conf and restart rgw. Once 100% sure of load we can simply remove the apache test for rgw and the configs from the recipes.

erhudy:
If this is going to break RADOSgw functionality by merging, then we should hold off on merging this until it can be fixed.

cloudm2:
It's not a break in functionality (SSL term will be handled another way). I want to test the same configs with Apache and Civetweb just so we know. Others are just going ahead with the change but I want to be able to supply numbers to the Ceph days and other events we're doing this summer. For now we should keep it like it is unless it's major.

erhudy:
That's fine, my main concern is that we should keep the default mechanism as Apache rather than Civetweb for the test environment so it works out of the box.

cloudm2:
I'm actually changing the condition to make it where it can just be flipped automatically.

mihalis68:
This change removes rgw recipe from work nodes, putting it only in the head node role, instead of in the work node role (which head nodes share). I presume this is deliberate?

cloudm2:
We had rgw on basically every node. In the new design it will be on it's on or mixed with a mon. Only thought about the smaller number (head nodes) but we can do it on worker instead. Thoughts?

cloudm2:
Updated pull request. Added new role, BCPC-Rgwnode. Added new role to BCPC-Worknode and since it's called by BCPC-Headnode means it's as before, on all nodes, for now. Updated repo for the ability to pull in patches quickly if needed or wanted. Also added the ability to set the 'frontend' attribute to "" or anything other than "civetweb" to enable Apache in the event we need or want to roll over to Apache. The attribute default['bcpc']['ports']['radosgw'] will need to be changed to 80 or 8080 for Apache and 8088 for civetweb.

erhudy:
Checks pass for me now.

However, Object Store within Horizon is now broken for me. Interaction with the Swift API has always been janky and I don't know how much (if at all) this particular part of Horizon is relied upon, so this may not be a game changer.

caiush:
my 2 cents, fix it or turn off the interface.

erhudy:
It seems to be pretty broken with the Keystone LDAP changes as well, so it would probably be better to just turn it off.

erhudy:
Tried it on master and it's broken there too. I will work up a PR independent of this to disable the Object Store panel in Horizon.

erhudy:
Made my PR for fixing the above in #643. I am satisfied with the state of this branch now.

erhudy:
Question: on the work nodes, in `/var/log/ceph`, there is a `ceph-osd.0.log` that stops being written to once the OSDs are all provisioned. This implies that there's a temporary state where we have duplicate OSD IDs floating around. Is this something to be concerned about?

cloudm2:
No. It's fine.

erhudy:
Reviewing the Civetweb vs Apache stuff, there will be some more work to do there to clarify what's really going on. It's not really Civetweb vs Apache, it's Civetweb vs FastCGI, because of the requirement for SSL termination.

We need to use Apache in either case, and so switching `['bcpc']['ceph']['frontend']` should configure/deconfigure Civetweb in `ceph.conf` and also change the Apache configuration from using FastCGI to mod_proxy. Civetweb also doesn't need to listen on all interfaces if Apache is proxying requests to it, and allowing it to do so may be a security issue.

mihalis68:
I have an on-metal cluster to build and would like this to be as good as possible when I start. Would this break if we bailed out of civetweb after building a cluster with this merged as is, for example?

mihalis68:
After more internal discussion this is good for now. There may be some further fettling after

PR #628, opened 2015-06-10
Avoid MySQL deadlocks with Zabbix

kelvk:
Make only one MySQL server operational at any one time to avoid MySQL deadlocks with Zabbix.


erhudy:
This worked fine for me in testing.

PR #631, opened 2015-06-11
Bring policy control management under chef

semenovroman:
Bring nova, glance, cinder, keystone and heat policies (policy.json) under chef.
Please note, this PR does not change any policies, just templatizes them.


erhudy:
There's a section in the Horizon configuration file (`/etc/openstack-dashboard/local_settings.py`) that appears to relate to policies as well. Does this also need to be updated?
```
# The Horizon Policy Enforcement engine uses these values to load per service
# policy rule files. The content of these files should match the files the
# OpenStack services are using to determine role based access control in the
# target installation.

# Path to directory containing policy.json files
#POLICY_FILES_PATH = os.path.join(ROOT_PATH, "conf")
# Map of local copy of service policy files
#POLICY_FILES = {
#    'identity': 'keystone_policy.json',
#    'compute': 'nova_policy.json',
#    'volume': 'cinder_policy.json',
#    'image': 'glance_policy.json',
#}
```

mihalis68:
A few of us cast our eyes over this and it's good.

PR #632, opened 2015-06-11
Prioritize Ceph OSD and monitor processes higher

erhudy:
This PR creates a simple upstart script that renices Ceph OSD processes to -10 after they have all been started. Some load pressure testing showed that re-nicing the OSDs kept responsiveness higher under high to extreme load conditions (hypervisor load averages of 10 or higher, tested up to 50).

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -121,6 +121,10 @@ default['bcpc']['ceph']['hdd']['ruleset'] = 2
 # See wiki for further details. 
 default['bcpc']['ceph']['rebalance'] = false
 
+# Set the default niceness of Ceph OSD and monitor processes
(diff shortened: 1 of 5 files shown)


kelvk:
Were Ceph mons reasonably responsive during your testing? Sluggish Ceph mons could result in VIP jumps and cascade into various failure situations, perhaps something we want to protect as well.

erhudy:
I didn't have problems with the responsiveness of the monitors but that thought did occur to me. This could be adapted pretty easily for the mons.

kelvk:
I think Ceph mons deserve the same priority as OSDs even if experiments suggest that it is less than necessary. Could we extend the PR to include mons as well?

erhudy:
Updated PR with a second service for head nodes to renice monitors via a separate attribute (also -10 by default).

kelvk:
Tested fine on my build.

PR #638, opened 2015-06-12
Reverts the nova-single-cpu patch hackery

erhudy:
This hack is no longer necessary because the underlying bug was fixed
upstream in OpenStack bug #1439280, and the updated libvirt package has
been released to the cloud archive. Building a fresh local cluster off
master with current packages from the cloud archive works without this
patch under VirtualBox, and I verified the original behavior functions
correctly by inspecting virsh dumpxml.

If you have an existing local cluster that you want to preserve, you can
revert the patch by uncommenting the 6 commented-out lines starting
around line 1835 of
/usr/lib/python2.7/dist-packages/nova/virt/libvirt/config.py and
restarting OpenStack services via hup_openstack.


kelvk:
LGTM

PR #639, opened 2015-06-12
Moving the dhcp-server patch from nova-setup.rb to nova-work.rb

### LINKED ISSUE #573: Multiple tenants broken in 5.x due to dhcp_server nova_network bug

erhudy:
This is a minor change to move this DHCP server patch that was originally created to fix #573 from `nova-setup.rb` to `nova-work.rb`, so that it applies to all compute hosts and not just head nodes.

### DIFF
--- a/cookbooks/bcpc/recipes/nova-setup.rb
+++ b/cookbooks/bcpc/recipes/nova-setup.rb
@@ -20,37 +20,6 @@
 include_recipe "bcpc::keystone"
 include_recipe "bcpc::nova-head"
 
-#  _   _  ____ _  __   __  ____   _  _____ ____ _   _
-# | | | |/ ___| | \ \ / / |  _ \ / \|_   _/ ___| | | |
-# | | | | |  _| |  \ V /  | |_) / _ \ | || |   | |_| |
-# | |_| | |_| | |___| |   |  __/ ___ \| || |___|  _  |
-#  \___/ \____|_____|_|   |_| /_/   \_\_| \____|_| |_|
-# this patch resolves OpenStack issue #1456321 and BCPC issue #573 -
-# fixes DHCP server assignment so that each fixed IP subnet gets its gateway
-# address as its DHCP server by default instead of all subnets getting the
-# gateway of the lowest subnet
-cookbook_file "/tmp/nova-network-dhcp-server.patch" do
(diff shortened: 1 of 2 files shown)


kelvk:
LGTM

kelvk:
Spun up VMs on multiple tenancies and they had their default gateway set correctly.

PR #643, opened 2015-06-12
Adding a way to hide Horizon panels via the Chef environment

erhudy:
This PR adds a mechanism via attributes for disabling panels in Horizon via the Django slug (see `openstack_dashboard/dashboards/project/dashboard.py` for the full list), and disables the Object Store panel, because the feature doesn't work for us.


mihalis68:
I reviewed this for safety and seems to be a no-op unless you define new attributes, so in that case is very safe. Confirmed this with the author, so this LGTM.

PR #644, opened 2015-06-14
Optionally bootstrap 1-3 monitoring nodes

kelvk:
As head nodes do not run monitoring, this updates the bootstrap process to optionally deploy additional VM(s) for the monitoring role. Monitoring bootstrapping is disabled by default, as anecdotally, a functional Openstack cluster is generally the requirement.

### DIFF
--- a/bootstrap/vagrant_scripts/vagrant_configure_chef.sh
+++ b/bootstrap/vagrant_scripts/vagrant_configure_chef.sh
@@ -47,20 +47,19 @@ do_on_node bootstrap "mkdir -p \$HOME/.chef && echo -e \"chef_server_url 'https:
 do_on_node bootstrap "$KNIFE ssl fetch"
 do_on_node bootstrap "$KNIFE bootstrap -x vagrant -P vagrant --sudo 10.0.100.3"
 
-# install and bootstrap Chef on cluster nodes
-for vm in vm1 vm2 vm3; do
-  do_on_node $vm "sudo dpkg -i \$(find /chef-bcpc-files/ -name chef_\*deb -not -name \*downloaded | tail -1)"
-done
-do_on_node bootstrap "$KNIFE bootstrap -x vagrant -P vagrant --sudo 10.0.100.11"
-do_on_node bootstrap "$KNIFE bootstrap -x vagrant -P vagrant --sudo 10.0.100.12"
-do_on_node bootstrap "$KNIFE bootstrap -x vagrant -P vagrant --sudo 10.0.100.13"
+# Initialize VM lists
+vms="vm1 vm2 vm3"
+if [ $MONITORING_NODES -gt 0 ]; then
+  i=1
+  while [ $i -le $MONITORING_NODES ]; do
+    mon_vm="vm`expr 3 + $i`"
+    mon_vms="$mon_vms $mon_vm"
+    i=`expr $i + 1`
+  done
+fi
 
-# install the knife-acl plugin into embedded knife and generate the actor map
+# install the knife-acl plugin into embedded knife
 do_on_node bootstrap "sudo /opt/opscode/embedded/bin/gem install /chef-bcpc-files/knife-acl-0.0.12.gem"
-do_on_node bootstrap "cd \$HOME && $KNIFE actor map"
-
-# using the actor map, set bootstrap and vm1 as admins so that they can write into the data bag
-do_on_node bootstrap "$KNIFE group add actor admins bcpc-bootstrap.$BOOTSTRAP_DOMAIN && $KNIFE group add actor admins bcpc-vm1.$BOOTSTRAP_DOMAIN"
 
 # rsync the Chef repository into the non-root user (vagrant)'s home directory
 do_on_node bootstrap "rsync -a /chef-bcpc-host/* \$HOME/chef-bcpc"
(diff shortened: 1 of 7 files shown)


erhudy:
I shall give this a test. One minor thought before I start: right now this uses the configure block for all cluster nodes, which allocates 4 extra volumes for Ceph per VM. These aren't used by the monitoring role, right?

The reason I ask is I have a to-do item to consolidate the two configure blocks into one down the road since there's a lot of duplicated material between the two and I'd like to revise it to be a single configuration block with appropriate guards depending on the VM's role, but this quickly becomes a more extensive refactoring of how the Vagrant bootstrap process creates VMs and determines their role, and implies that we would want to bring in something like `cluster.txt` to handle this whole process in a manner consistent with hardware clusters.

kelvk:
Correct. The monitoring role uses just the first device. I agree we would need to move on to a cluster definition approach as roles are increasingly fragmented. There had been occasions when I did not need work nodes. In the meantime, this PR hopefully will make testing monitoring more convenient.

erhudy:
I agree, I think that as long as everything works here this can be merged as-is and then we'll come back later to try to make it all consistent.

bootstrap/vagrant_scripts/vagrant_configure_chef.sh
```diff
@@ -81,29 +80,61 @@ do_on_node bootstrap "$KNIFE cookbook upload apt bcpc chef-client cron logrotate
...
   # run on head node one last time to update HAProxy with work node IPs
   do_on_node vm1 "sudo chef-client"
   # HUP OpenStack services on each node to ensure everything's in a working state
-  for vm in vm1 vm2 vm3; do
+  for vm in $vms; do
     do_on_node $vm "sudo hup_openstack || true"
   done
+  # Run chef on each mon VM before assigning next node for monitoring.
+  for vm in $mon_vms; do
+    do_on_node bootstrap "$KNIFE node run_list set bcpc-$vm.$BOOTSTRAP_DOMAIN 'role[BCPC-Monitoring]'"
```
erhudy: Can you migrate this section up to where the other roles are set? My initial reasoning behind the `BOOTSTRAP_CHEF_DO_CONVERGE` variable was that it should stop just before converge, and all the roles/environments should be assigned out and ready to go.

erhudy:
So, I tested with 3 nodes and 0 nodes and it works as expected in both cases.

One problem I have found that is really a Vagrant issue (you can argue whether it's a bug or not), is that Vagrant reads the Vagrantfile every time you run a command instead of referencing metadata in the `.vagrant` folder, should the folder exist. What this means is that Vagrant effectively loses track of the extra monitoring VMs unless `MONITORING_NODES` is set every time you run a Vagrant command. Observe:
```
➜  vagrant_scripts git:(09922a9) vagrant status          
Current machine states:

bootstrap                 not created (virtualbox)
vm1                       not created (virtualbox)
vm2                       not created (virtualbox)
vm3                       not created (virtualbox)

This environment represents multiple VMs. The VMs are all listed
above with their current state. For more information about a specific
VM, run `vagrant status NAME`.
➜  vagrant_scripts git:(09922a9) MONITORING_NODES=3 vagrant status
Current machine states:

bootstrap                 not created (virtualbox)
vm1                       not created (virtualbox)
vm2                       not created (virtualbox)
vm3                       not created (virtualbox)
vm4                       not created (virtualbox)
vm5                       not created (virtualbox)
vm6                       not created (virtualbox)

This environment represents multiple VMs. The VMs are all listed
above with their current state. For more information about a specific
VM, run `vagrant status NAME`.
```
Basically what this means is that Vagrant will lose track of the extra monitoring VMs unless that environment variable is set, which can lead to some weirdly counterintuitive behavior (Vagrant not destroying all VMs if you reduce the number of monitoring nodes between cluster rebuilds, for example, or not being able to `vagrant ssh` to VMs that you know are running).

bootstrap/vagrant_scripts/vagrant_configure_chef.sh
kelvk: Probably, but likely not without reworking some recipes. I attempted to and found out that uninitialized node['bcpc']['management']['ip'] of second/third nodes would break during execution of haproxy/mysql/etc recipes. We used to knife bootstrap with role, so this hadn't been an issue until now.

erhudy:
The above comment wasn't meant to be a reason why this shouldn't be merged, just an observation/caution about counterintuitive Vagrant behavior that will be encountered after the merge if you start adding monitoring nodes.

bootstrap/vagrant_scripts/vagrant_configure_chef.sh
erhudy: That section only sets the role on the Chef server, it doesn't perform any convergence (it's so that you can go to the node and just run `chef-client`). If somebody runs it out of order, it's their own fault.

kelvk:
Yes, that's rather annoying and I think the presence of a cluster definition file, if it is then used to generate the Vagrant file (thinking aloud here), could lay this to rest.

kelvk:
Added a note about this behavior in `FAQ.md`. `vagrant_clean.sh` is also updated to ensure mon VMs are always destroyed.

bootstrap/vagrant_scripts/vagrant_configure_chef.sh
erhudy: Talked to @kelvk, rescinding this request.

erhudy:
Built fine with 3 monitoring nodes, doing one last build with no monitoring nodes and then will merge if successful.

erhudy:
Opened #648 to track the issue discussed above, merging.

PR #645, opened 2015-06-15
Extending Keystone wait timeout and making it an attribute

erhudy:
Apparently 30 seconds isn't always enough time to wait for Keystone to get ready. :water_buffalo:


mihalis68:
I've had this problem also when a build is slow (e.g. when not using optimal VM flags). This looks good to me, or at least I agree.

kelvk:
Tested fine on my build.

PR #633, opened 2015-06-11
Rework Zabbix host discovery for auto host registration

kelvk:
This reworks Zabbix host registration that existed in RC2/3. 

Zabbix will discover hosts every 10 minutes in the management network by default, then associates them to appropriate template and hostgroup (BCPC-Headnode, BCPC-Worknode, BCPC-Monitoring). These templates will likely be evolving as there appears room for checks/triggers expansion. For reviewer(s), I'd say the committed xml files can be safely ignored.

PR #629, opened 2015-06-11
Configure Zabbix templates

kelvk:
At some point between RC2 and RC3 the configuration of Zabbix templates appeared to had been removed from Chef. This reinstates that, with following changes:

1. Bypassing of proxy for zabbix_config (a utility to import templates) requests to the monitoring VIP
2. Update of Zabbix BCPC templates to reflect 3 distinct roles (Head, Work, Monitoring)

The templates will likely be evolving as there seems to be plenty of room for checks/triggers expansion. For reviewer(s), I'd say the committed xml files can be safely ignored.

PR #647, opened 2015-06-16
Make monitoring hostnames work and provide overrides

kelvk:
Directing requests to monitoring sites based on host header never seemed functional, so this fixes it by implementing it on haproxy correctly. IP-based requests will continue to work as usual.

In addition, there may be occasions where monitoring functions are delegated to an external cluster. FQDN overrides are provided to at least maintain working hyperlinks in the landing page.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -990,6 +990,9 @@ default['bcpc']['cpupower']['ondemand_up_threshold'] = nil
 #
 ###########################################
 #
+# Graphite Server FQDN
+default['bcpc']['graphite']['fqdn'] = "graphite.#{node['bcpc']['domain_name']}"
+#
 # Default retention rates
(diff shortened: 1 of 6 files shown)


erhudy:
I deployed this locally with 3 monitoring nodes; everything built cleanly out of the box without requiring any intervention. `graphite`/`zabbix`/`kibana` all resolve correctly, the links from the landing page work (though out of the box the links with the domain names won't work unless you change your desktop to use the PowerDNS resolver on the management VIP, obviously).

Anything else necessary to test for this? I observed one failing check in Zabbix saying RADOSGW tests in HAProxy were not succeeding; it seems unrelated but just mentioning for clarity.

kelvk:
That is sufficient, thanks. The check failure is unrelated. While the Zabbix templates have been cleaned up significantly in #629 / #633, further refinement of checks is necessary. This will be a separate task.

erhudy:
Okay, unless you have any further concerns I will merge.

Issue #514, opened 2015-03-10
Enable RBD cache on libvirt 


caiush:
In firefly this is off by default.

caiush:
here is the docs for firefly:http://ceph.com/docs/v0.80.5/rbd/rbd-config-ref/

kelvk:
I found the following statement in the docs important, and probably something tenants should know about.

"Since the cache is local to the client, there’s no coherency if there are others accessing the image. Running GFS or OCFS on top of RBD will not work with caching enabled."

kelvk:
Fixed in #653

PR #653, opened 2015-06-18
nova RBD updates

### LINKED ISSUE #651: Guests booted from image are not using RBD-backed disk
The disk should be backed by RBD in the "vms" pool. nova.conf needs to be updated to reflect Kilo configuration changes per http://ceph.com/docs/master/rbd/rbd-openstack/.

We should also verify Glance and Cinder configurations.

kelvk:
This fixes #651 and enables writeback caching per #514.


kelvk:
Besides using http://ceph.com/docs/master/rbd/rbd-openstack/ as reference for the changes, I also discovered more required changes in our nova.conf according to http://docs.openstack.org/kilo/config-reference/content/list-of-compute-config-options.html, such as using compute_driver instead of connection_type and moving of use_virtio_for_bridges to libvirt section.

PR #655, opened 2015-06-18
Re-adding some hostname-related patches that were lost in the move to 5.x

erhudy:
This PR re-adds some patches (modified) that were forgotten when we moved to 5.x. Specifically, this alters the behavior of nova-network and nova-api so that DHCP leases are provided with hostnames in the `ip-x-x-x-x` AWS style, and the metadata server will now return `ip-x-x-x-x.bcpc.example.com` for the internal hostname and `public-y-y-y-y.bcpc.example.com` for the external hostname.

### DIFF
--- a/cookbooks/bcpc/files/default/nova_api_metadata_base.patch
+++ b/cookbooks/bcpc/files/default/nova_api_metadata_base.patch
@@ -1,0 +1,55 @@
+diff --git a/nova/api/metadata/base.py b/nova/api/metadata/base.py
+index 382378f..369e2c9 100644
+--- a/nova/api/metadata/base.py
++++ b/nova/api/metadata/base.py
+@@ -14,6 +14,8 @@
+ #    License for the specific language governing permissions and limitations
+ #    under the License.
+ 
++# THIS FILE PATCHED BY BCPC
++
+ """Instance Metadata information."""
+ 
+ import base64
+@@ -223,6 +225,11 @@ class InstanceMetadata(object):
+         fixed_ips = self.ip_info['fixed_ips']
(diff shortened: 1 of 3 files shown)


erhudy:
Note for testing: CirrOS appears to interpret cloud-init differently from Ubuntu/CentOS 6.6, or maybe dropbear overrides cloud-init. CirrOS will report the bare hostname for both `hostname` and `hostname -f`. CentOS 6.6 and Ubuntu 14.04 both report the FQDN for `hostname -f` and the bare hostname for `hostname` as expected, so this seems to be CirrOS weirdness.

mihalis68:
Patched this into a vagrant-built cluster and rechefed, the rebuilt a testing Cirros instance. 
```
$ hostname -f
ip-1-127-0-4
```
Which is as expected when the patched worked.

kelvk:
I associated a floating IP after the VM is booted and saw dns_fill.py doing its thing. However, the metadata server still returns a hostname containing the fixed ip. Is this expected?

ubuntu@ip-1-127-0-3:~$ curl http://169.254.169.254/latest/meta-data/public-hostname
public-1-127-0-3.bcpc.example.com

erhudy:
No, that looks incorrect. Let me look.

erhudy:
That error is now fixed in the current patch (force push). It will now use the lowest floating IP address assigned (that could possibly differ, but the important thing is that it doesn't flip out the instance has multiple floats assigned). If no floating IP is assigned, public-hostname on the metadata server will be blank.

kelvk:
LGTM.

PR #656, opened 2015-06-18
Improve resilience if one of more memcached servers are down

erhudy:
This PR reduces the timeouts that Nova and Keystone will wait for a reply from memcached before marking that instance down. It also copies the memcached server list into another spot in nova.conf, since apparently it's supposed to be specified twice.

(Note: the option names in both keystone.conf and nova.conf are different, despite describing the same thing in both cases, because of course they are. I am pretty sure they are correct for each file, but if you would like to verify, the config file references are at http://docs.openstack.org/kilo/config-reference/content/section_keystone.conf.html and http://docs.openstack.org/kilo/config-reference/content/list-of-compute-config-options.html.)

### DIFF
--- a/cookbooks/bcpc/templates/default/keystone.conf.erb
+++ b/cookbooks/bcpc/templates/default/keystone.conf.erb
@@ -102,11 +102,11 @@ memcache_servers = <%=@servers.map{|x| x['bcpc']['management']['ip'] + ":11211"}
 <% else %>
 #memcache_servers =
 <% end %>
-#memcache_dead_retry = 300
-#memcache_socket_timeout = 3
+memcache_dead_retry = 60
+memcache_socket_timeout = 1
 #memcache_pool_maxsize = 10
 #memcache_pool_unused_timeout = 60
-#memcache_pool_connection_get_timeout = 10
+memcache_pool_connection_get_timeout = 1
 
 [catalog]
 #template_file = default_catalog.templates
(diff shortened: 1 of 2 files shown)


kelvk:
I can't really verify its behavior since the Vagrant build is limited to one head node. Good to merge if you or anyone else has tested it on multi-head cluster.

erhudy:
I did a local build with three head nodes and took down one and two memcached instances. It occurs to me now that I did not test behavior if all memcached instances are down, so I will do a local build when I get home and see how it reacts to all instances being unavailable.

erhudy:
So, no memcached is a bit bad, because it's where Keystone expects to be able to store tokens. If all memcached instances are unavailable, everything grenades pretty harshly with 500s. I don't know if we can defend against this, though.

erhudy:
On a local cluster with 3 head nodes, taking 1 memcached out of rotation caused pretty noticeable slowdown in API requests, and at 2 out of rotation API calls failed more often than not. Given the limitations of the tunables in this PR, I don't know how much I can do about that. When testing on hardware, removing memcached instances had almost no effect, though. I will check tomorrow to verify that everything was set up the same way on the hardware, because it's a huge difference in behavior. (However, if it does behave better on hardware than on local virtualized clusters, that's a good thing, because the whole point of this PR is to allow flexibility in removing head nodes for maintenance without requiring tedious reconfiguration of other head nodes.)

erhudy:
Tested locally again after giving each node more memory and everything worked more sanely. I think this is safe to merge.

PR #657, opened 2015-06-18
Update policy files

semenovroman:
This PR brings policy files up to date with Kilo release.


erhudy:
Will test locally.

erhudy:
Instead of explicitly specifying every line to output into the JSON, can I suggest just iterating over the different key-value pairs in the attribute dictionary in a block? You could generate the JSON in the recipe and pass it through to the template in a string. That way it's not necessary to specify everything in two places, and adding new attributes will automatically update the JSON.

erhudy:
Actually, you wouldn't even need to do that. You can marshal the attribute dictionary to JSON in the recipe with
```
require 'json'
keystone_policy_json = JSON.dump node['bcpc']['keystone']['policy']
```
and then pass that into the policy.json template.

caiush:
Yeah I prefer that solution, It will make for more maintainable policy templates and allow us to more flexibility in defining new policies on the API.

erhudy:
Built this and the JSON looks fine and everything seems fine from a functionality standpoint. I did not test anything to do with whether the policies themselves are sane, but I'm assuming this has already been vetted and/or these are the OpenStack default policies.

semenovroman:
Policy files were checked out from 'kilo' branch from @openstack

erhudy:
I would like to merge this today. Does anyone else have any comments?

PR #661, opened 2015-06-21
Disconnected build actually works now!

erhudy:
I figured out what was causing the problem with setuptools; any call to `setuptools.setup_requires()` will invoke easy_install. This PR fixes that by configuring easy_install with similar configuration options to pip, so that it will not attempt to talk to PyPI. I tested this on both Linux and OS X by pulling down all the dependencies, mirroring everything locally, providing NTP/DNS services from the build machine, and then disconnecting the network interface entirely (shutting off WiFi/pulling the network cable).

This PR also adds documentation and a sample apt-mirror configuration file and updates a few of the build scripts to allow their inputs/outputs to be more easily overridden by setting certain environment variables.

### DIFF
--- a/bootstrap/common_scripts/common_build_bins.sh
+++ b/bootstrap/common_scripts/common_build_bins.sh
@@ -4,9 +4,10 @@ set -e
 # This script must be invoked from the root of the repository (e.g., as
 # bootstrap/common_scripts/common_build_bins.sh).
 
-FILES_ROOT=/chef-bcpc-files
+if [[ -z $FILES_ROOT ]]; then FILES_ROOT=/chef-bcpc-files; fi
+if [[ -z $BUILD_DEST ]]; then BUILD_DEST=cookbooks/bcpc/files/default/bins; fi
 
-pushd cookbooks/bcpc/files/default
+pushd $BUILD_DEST
 # Define the appropriate version of each binary to grab/build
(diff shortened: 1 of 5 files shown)


bootstrap/common_scripts/common_build_bins.sh
```diff
@@ -165,6 +165,11 @@ if [ ! -f rally-pip.tar.gz ] || [ ! -f rally-bin.tar.gz ]; then
   # We install rally and a few other items here. Since fpm does not resolve dependencies but only lists them, we
   # have to force an install and then tar up the dist-packages and local/bin
   PIP_INSTALL="pip install --no-cache-dir --disable-pip-version-check --no-index -f $FILES_ROOT/rally"
+  # this kludge is to prevent easy_install from trying to go out to PyPI:
+  # rally calls setuptools.setup_requires(), which uses easy_install to
+  # install any packages listed there; this forces easy_install to use
+  # the same mechanism as we are telling pip to use in $PIP_INSTALL
+  echo -e "[easy_install]\nallow_hosts = ''\nfind_links = file://$FILES_ROOT/rally/" > $HOME/.pydistutils.cfg
```
erhudy: This is the change that reconfigures easy_install.

kelvk:
Tested that it does not break normal build.

PR #665, opened 2015-06-23
Replaces compiled Zabbix with official Zabbix packages

erhudy:
This PR replaces the old compiled Zabbix setup with upstream packages. It includes migration resources to remove compiled Zabbix if present, so it should be tested both as an upgrade to an existing cluster and as a fresh installation. This also upgrades Zabbix from 2.2.2 to 2.2.9.

NOTE: if testing the upgrade on an existing cluster, update the run list for the **BCPC-Monitoring** role (`knife role from file roles/BCPC-Monitoring.json`). I renamed the `zabbix` cookbook to `zabbix-server` to more clearly differentiate it from the agent.


erhudy:
Force pushed a change to this branch that fixes a mistake I made with not updating the paths to the Zabbix database schemas that caused a fresh build to fail. In-place upgrades would have succeeded because the schema was already in place and upgraded by Zabbix.

kelvk:
The agent part seems to work fine. Checks are still running as expected.

Without recheffing the monitoring VM yet, it appears that removal of the zabbix-server upstart job will break migration of Zabbix-server via keepalived (https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/templates/default/upstart-zabbix-server.conf.erb#L9-L12) and Zabbix-server startup safeguarding.

erhudy:
Is working around that by augmenting `vip_won` and `vip_lost` just for the monitoring role to start/stop the Zabbix server an acceptable solution?

kelvk:
Partially, with some work. For example, `vip_change` in `vip_won`, isn't necessary. We could probably just pass `/etc/init.d/zabbix-server <stop|start>` as keepalived notify commands as well.

The other is the upstart safeguard. I would like that in place because it prevents any accidental startup of zabbix-server on non-VIP-holding node.

erhudy:
Okay, I'll give it some thought.

caiush:
the reason for the emit was that the zabbix start/stop can take a while. The emit was an easy way to just let upstart deal with the termination/startign of zabbix. I hit a lot of problems because it was taking zabbix too long to actually return from service stop.

erhudy:
Pushed the WIP so I can come back to this later. Tagged as do not merge, because this is definitely a bit broken now (transitioning things over from the sysvinit to upstart script is not going smoothly).

kelvk:
I've updated the branch with my changes to use upstart correctly, amongst other minor changes, such as removal of user/directory resources since the Zabbix packages now take care of that.

erhudy:
I was able to build a 6-node cluster with 3 monitoring nodes both directly from the branch and by recheffing the branch over master (which has not diverged significantly from 5.3.0, so I think that that serves as sufficient assurance that it will upgrade cleanly from 5.3.0).

Unless anyone brings up any further concerns, I will merge this on Monday.

PR #666, opened 2015-06-23
Updates default security group mechanism for Kilo

erhudy:
This was a major oversight on my part. In previous versions of OpenStack, a security group named **default** was treated as the default security group. At some point, the default security group mechanism was moved off to its own thing, but recipes were not updated appropriately. This updates our recipes to create the default security group rules in the expected format.

### DIFF
--- a/cookbooks/bcpc/recipes/nova-setup.rb
+++ b/cookbooks/bcpc/recipes/nova-setup.rb
@@ -28,14 +28,14 @@ bash "wait-for-nova-to-become-operational" do
 end
 
 
-bash "nova-default-secgroup" do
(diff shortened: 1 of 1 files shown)


erhudy:
To test, create some more tenants and a user in one of the tenancies, log in, and launch an instance without doing anything else (e.g., going to the security tab). Assign a floating IP, and you should be able to ping and SSH. This was broken previously unless you visited the security group rules first, and possibly you also had to tweak the rules manually to get them to take effect.

erhudy:
Note: this can be safely cheffed onto an existing cluster without any modifications.

caiush:
+1

PR #667, opened 2015-06-23
Bug fix and metadata config extension.

caiush:


cookbooks/bcpc/templates/default/nova.conf.erb
```diff
@@ -38,7 +38,7 @@ osapi_compute_extension=nova.api.openstack.compute.contrib.standard_extensions
 osapi_compute_extension=cobalt.nova.osapi.cobalt_extension.Cobalt_extension
 <% end %>
 <% if not node["bcpc"]["vendordata_driver"].nil? %>
-vendordata_driver = node["bcpc"]["vendordata_driver"].
+vendordata_driver = <%=node["bcpc"]["vendordata_driver"] %>
```
acaiafa: you want to add a space here 
```ruby
<%= node["bcpc"]["vendordata_driver"] %>
```

cookbooks/bcpc/templates/default/nova.conf.erb
```diff
@@ -38,7 +38,7 @@ osapi_compute_extension=nova.api.openstack.compute.contrib.standard_extensions
 osapi_compute_extension=cobalt.nova.osapi.cobalt_extension.Cobalt_extension
 <% end %>
 <% if not node["bcpc"]["vendordata_driver"].nil? %>
```
acaiafa: could make this 
```ruby
 <% unless node["bcpc"]["vendordata_driver"].nil? %>
```

cookbooks/bcpc/templates/default/nova.conf.erb
caiush: Why, its not the convention we have used everywhere else. If we want to switch to that (perhaps a good thing), we should do that in a separate PR, i.e. change the spacing convention.

acaiafa: ah ok just saw it and figured to point out a common pattern that is widely used. I noticed the rest of the templates are like this so my apologies.

caiush: There is a lot of, cough, "cruft" in the recipes that need to be cleaned up and made more standard. We can at that to the list.

cookbooks/bcpc/templates/default/nova.conf.erb
```diff
@@ -182,3 +182,12 @@ inject_partition=-2
 live_migration_flag="VIR_MIGRATE_UNDEFINE_SOURCE,VIR_MIGRATE_PEER2PEER,VIR_MIGRATE_LIVE,VIR_MIGRATE_PERSIST_DEST,VIR_MIGRATE_TUNNELLED"
 virt_type=<%=node['bcpc']['virt_type']%>
 use_virtio_for_bridges=True
+
+
+<% unless node["bcpc"]["vendordata_config"].nil? %>
+[metadata]
```
erhudy: If `['bcpc']['vendordata_config']` is present but empty, there will be a `[metadata]` section without anything in it. Will that cause problems?

caiush: Provides a way to add in custom metadata driver config into the BCPC cookbooks. Should be a no-op for the laptop env.

erhudy:
What's the right way to test that this fixed something in terms of functionality? Obviously looking at the template output to verify it was actually replaced with a thing is something, but what should be tested in terms of the functionality of the metadata service?

PR #670, opened 2015-06-24
Make Graphite webapp work properly with SSL termination

kelvk:
The URL produced by "short URL" button in Graphite web composer does not load as the request is redirecting to HTTP instead of HTTPS. This PR makes Django behave correctly behind SSL terminating proxy and updates the header name to comply with the Forwarded HTTP extension guidelines.


erhudy:
After applying this patch, I still get redirected to HTTP. e.g., the short URL https://10.0.100.6:8888/S/D tries to redirect to http://10.0.100.6:8888/render/?width=586&height=308

erhudy:
I ensured that both Apache and HAProxy were restarted, though Chef should have already restarted them due to the template updates.

kelvk:
Updated commit. I was testing against the port 443 https frontend.

erhudy:
Okay, that worked for me now.

PR #671, opened 2015-06-24
Hacks in a fix on the trousers postinst script so that it installs

erhudy:
This fixes (I hope) an issue with getting TPM set up on hardware. I regret that this exists.

### DIFF
--- a/cookbooks/bcpc/recipes/tpm.rb
+++ b/cookbooks/bcpc/recipes/tpm.rb
@@ -2,7 +2,7 @@
 # Cookbook Name:: bcpc
 # Recipe:: tpm
 #
-# Copyright 2014, Bloomberg Finance L.P.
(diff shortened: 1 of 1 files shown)


mihalis68:
best branch name ever

mihalis68:
I'll test this

erhudy:
Amended to reorder trousers ahead of the other packages, since tpm-tools tries to install it as a dependency.

mihalis68:
This seems to work for me. The only nodes I am having problems with had a failed install before. My fix is this:

```
$ sudo apt-get purge tpm-tools rng-tools
$ sudo chef-client
```
and this works with the above manually merged in and loaded onto my chef server.

mihalis68:
Has now worked on about 25 real machines. Is good.

PR #672, opened 2015-06-24
reorders Erlang packages to install in dependency order

erhudy:
Erlang/OTP 18.0 is out. This PR reorders the list of Erlang packages so that as they install, each dependency has already been satisfied, and apt won't go out and get some Erlang 18 package, install that, and then blow up on dependencies.


erhudy:
This may still be malfunctioning, continuing to test.

erhudy:
Reorganized the list again and it works properly now (`erlang-nox` is apparently a bit of a weird package that was causing everything else to get upgraded to 18.0).

PR #673, opened 2015-06-24
Updates rng-tools to work properly with Trusty

erhudy:
It was too much to hope that all the crypto stuff would _really_ be working correctly the first time through. This PR should fix everything for real this time by ensuring that the `tpm_rng` module is loaded into the kernel prior before doing anything, and adds it to `/etc/modules` so that `/dev/hwrng` is available on boot and **rng-tools** will start without grumpiness. (This also installs `linux-extra-image-virtual`, which is a virtual package that pulls in the appropriately versioned kernel extras, which is where `tpm_rng` comes from.)

### DIFF
--- a/cookbooks/bcpc/recipes/tpm.rb
+++ b/cookbooks/bcpc/recipes/tpm.rb
@@ -17,8 +17,23 @@
 # limitations under the License.
 #
 if node['bcpc']['enabled']['tpm'] then
-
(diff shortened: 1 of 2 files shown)


mihalis68:
testing this now

mihalis68:
this chefed in and worked. Now trying a reboot, which is slow

PR #678, opened 2015-06-27
Added ephemeral storage based on host aggregates

cloudm2:
1. Ephemeral and Persistent storage
2. Added --property hw_scsi_model=virtio-scsi --property hw_disk_bus=scsi to glance image-create and nova.conf [libvirt] hw_disk_discard=unmap so that purged file space can be reclaimed with fstrim. This needs to apply to all images
3. Changed ceph.conf to get rid of health warning on too many pgs on small clusters - good thing
4. Set stage for next PR on LVM


cloudm2:
Modifying disk.local so closing for now.

PR #679, opened 2015-06-27
Added ephemeral storage based on host aggregates

cloudm2:
1. Added ephemeral and persistent host aggregates
2. Corrected ceph.conf on too many pgs on small clusters
3. Setup ephemeral roles. If the host is an ephemeral role then no rbd
4. No way to split disk, disk.local and snapshots so as to mount them on different configs. Blueprint outstanding on this.
5. All snapshots (ephemeral and persistent) go to rbd not local for ephemeral (maybe a config setting but I didn't see it)
6. Because of #4 disk (root) and disk.local (ephemeral) are on same volume. Disk (root) can not go to rbd while disk.local (ephemeral) goes to local volume
7. LVM will be in subsequent PR


erhudy:
Same request here as on #679; please restore the roles to their original state and create new **-Ephemeral** roles that can be used for testing.

erhudy:
I am deploying this with slightly modified roles on hardware to begin testing.

PR #680, opened 2015-06-29
LVM for ephemeral storage added

cloudm2:
*Had to updated the haproxy version to 1.5.13 because 1.5.12 in the repo we use changed it on sat.

Depends on ephemeral PR. The environment file(s) will need to be updated to support ephemeral and drives just like the ceph portion does.


bootstrap/vagrant_scripts/Vagrantfile
```diff
@@ -268,14 +268,26 @@ Vagrant.configure("2") do |config|
...
         ('b'..'e').each_with_index do |disk, idx|
           disk_file = File.join(vm_dir, "vm#{i}-#{disk}.vdi")
           unless File.exist?(disk_file)
             vb.customize ["createhd", "--filename", disk_file, "--size", disk_size]
             vb.customize ["storageattach", :id, "--storagectl", "SATAController", "--device", "0", "--port", "#{idx+1}", "--type", "hdd", "--medium", disk_file]
           end # File.exist
+          last_port = idx+1
         end # b..e each
+
+        unless i == 2 or i == 3
```
erhudy: This will go a bit wrong if monitoring nodes are added.

bootstrap/vagrant_scripts/Vagrantfile
```diff
@@ -268,14 +268,26 @@ Vagrant.configure("2") do |config|
         rescue
           fail "Unable to locate VM #{vm_name} on disk, terminating"
         end        
-        
+
+        last_port = 0
         ('b'..'e').each_with_index do |disk, idx|
```
erhudy: What you might want to do instead of duplicating this section below is to create the range based on whether `i` is 1 or something else (i.e., we're working on the head node or another node). Something like:
```
disk_range = (i == 1) ? ('b'..'h') : ('b'..'e')
disk_range.each_with_index do |disk, idx|
```
This sets `disk_range` to `('b'..'h')` if `i` is 1 (i.e., it's the head node), and `('b'..'e')` for everything else. This also eliminates `last_port` because you don't need to remember where you were.

erhudy: Actually, since @kelvk and I had discussed this a bit already in the context of not creating extra disks for monitoring nodes at all, this would probably be an ideal situation for a case statement:
```
disk_range = case i; when 1 then ('b'..'h'); when 2..3 then ('b'..'e'); else []; end
```
(Of course this whole scenario is a bit crap in the first place because we're making assumptions about what each node will be used for based entirely on the index of the VM in the Vagrantfile, but that's an issue for another day.)

cookbooks/bcpc/recipes/nova-ephemeral.rb
```diff
@@ -0,0 +1,48 @@
...
+
+template "/tmp/nova-ephemeral.sh" do
+  user 'root'
+  source "nova-ephemeral.sh.erb"
+  owner "root"
+  group "root"
+  mode 0755
+end
+
+bash "nova-ephemeral-create" do
```
erhudy: Now that I understand what's going on maybe, maybe it would be more straightforward to use the convention from the rest of the cookbook and source `/root/adminrc` prior to executing the command, since `/root/adminrc` already contains most of the needed environment variables (and avoids the problem with `get_config`). That would rewrite the command in this resource as
```
. /root/adminrc; /tmp/nova-ephemeral.py --os-compute-api-version=2 --persistent --availability-zone="#{node['bcpc']['region_name']}" --hosts-persistent="#{node['hostname']}"
```

cookbooks/bcpc/recipes/nova-persistent.rb
```diff
@@ -0,0 +1,48 @@
...
+
+template "/tmp/nova-persistent.sh" do
+  user 'root'
+  source "nova-persistent.sh.erb"
+  owner "root"
+  group "root"
+  mode 0755
+end
+
+bash "nova-persistent-create" do
```
erhudy: Comment from the **nova-ephemeral** recipe also applies here.

cloudm2:
The above suggestions are being tested and an updated PR will show in the timeline soon.

roles/BCPC-Headnode.json
```diff
@@ -27,7 +27,7 @@
       "recipe[bcpc::powerdns-nova]",
       "recipe[bcpc::heat]",
       "recipe[bcpc::horizon]",
-      "role[BCPC-Worknode]",
+      "role[BCPC-Worknode-Ephemeral]",
```
erhudy: What's the rationale for changing head nodes to provide ephemeral storage by default? Was this a testing thing?

roles/BCPC-Worknode-Ephemeral.json
```diff
@@ -0,0 +1,35 @@
...
+      "recipe[bcpc::checks-work]",
+      "recipe[bcpc::zabbix-agent]"
+    ],
+    "description": "A functional compute node in a BCPC cluster with ephemeral storage",
+    "chef_type": "role",
+    "override_attributes": {
+      "bcpc": {
+        "nova": {
+          "persistent": false,
+          "ssd_disks": ["sdf", "sdg", "sdh"]
```
kelvk: Is it expected that a node can run both BCPC-Worknode-Ephemeral and BCPC-Worknode-Persistent, hence the need to create additional disks?

I think `attributes/default.rb` contains (or at least, intended to) full list of configurable attributes, so having ssd_disks in there would be handy.

erhudy: My reading is that a node is expected to be either ephemeral or persistent, in which case I don't think the extra disks would be necessary.

cookbooks/bcpc/templates/default/ceph.conf.erb
```diff
@@ -11,6 +11,10 @@
     cluster network = <%= @node["bcpc"]["storage"]["cidr"] %>
     keyring = /etc/ceph/$cluster.$name.keyring
     mon host = <%= @servers.map{|s| s["bcpc"]["storage"]["ip"] + ":6789"}.join(', ') %>
+    mon_pg_warn_max_per_osd=0
+    mon_osd_full_ratio=.85
+    mon_osd_nearfull_ration=.70
+    osd_backfill_full_ratio=.70
```
kelvk: The parameters should not have underscores, according to http://ceph.com/docs/master/rados/configuration/mon-config-ref/. mon_osd_nearfull_ration is also misspelt.

What is the rationale for not using the defaults?

erhudy:
Also, how is this PR differentiated from #679? It is difficult to determine what needs to be evaluated when the two PRs have a high degree of overlap.

roles/BCPC-Headnode.json
cloudm2: For testing as of now. It can be moved anywhere we want in another PR. This is a good discussion point on overall design of cluster.

roles/BCPC-Worknode-Ephemeral.json
cloudm2: Caius and I discussed breaking it out into environment for a given node. It can be another good design topic for discussion. @erhudy, it's not that a node is either ephemeral or not, it's which node nova "thinks" is ephemeral or not. Meaning, ceph can run on any of them like it does now. On the hand, another good topic for discussion on overall design of the cluster.

cloudm2:
LVM was simply branched off of ephemeral.

cookbooks/bcpc/recipes/nova-ephemeral.rb
cloudm2: Done on ephemeral but not persistent since worker nodes do not use adminrc.

roles/BCPC-Headnode.json
erhudy: Can you restore the existing head node role back to its previous state and create a separate role for ephemeral headnode (e.g., **BCPC-Headnode-Ephemeral**)? I would like for it to be an opt-in feature at the moment so that we're only testing ephemeral if we specifically want to.

PR #682, opened 2015-06-29
Dropping parts of the repo that are vestigial/unmaintained

erhudy:
This PR is a work in progress and should not be merged at this time. However, I am opening it now to let people comment on proposed deletions and changes.


caiush:
+100, 

btw: coblat recipes can go

@jerenkrantz will be sad removing vmware.

erhudy:
I would like to re-add VMware stuff on Vagrant eventually. I tinkered with it a bit when doing the initial Vagrant re-do and found that it was a lot less trivial than one might hope:
* sourcing upstream Vagrant boxes that support the VMware provider is annoying
* we use a number of calls to VBoxManage directly in the Vagrantfile to do things that Vagrant doesn't natively support (mostly creating the extra disks)
* Vagrant VMware provider has different behavior from the VirtualBox provider in some areas (e.g., `vagrant destroy` on a nonexisting VM is a no-op with VirtualBox but blows up with VMware)
* the two different networking setups operate somewhat differently

As for VMware on the non-Vagrant path, I don't know if it's worth spending time continuing to maintain it. In my view ( @mihalis68 may disagree) the chief purpose of the non-Vagrant build path is to vet the functionality of the Cobbler/PXE setup and various scripts that interact with the hardware clusters and `cluster.txt`. Whether that happens on VMware or VirtualBox is immaterial.

mihalis68:
As discussed just now:

I'm in favor of removing CentOS support.

Not in favor of removing VMware support, but fine with vmware not being an option for the non-vagrant path. The non-vagrant path is, as @erhudy says, useful for pretending we are dealing with hardware for the most part.

erhudy:
Rewrote the commits to make it clearer what's being removed/rewritten.

kelvk:
ceph-yum.rb looks safe to remove.

erhudy:
Feel free to push a commit with that. I'm keeping everything in separate commits right now while we review the changes, and then before merge I will crush it all into one commit.

jerenkrantz:
I'd prefer we keep the vmware directory - if we don't delete it, I'll
submit a PR to update it in the coming weeks to match the new Vagrant
model.  (Most of the issues @erhudy mentions are resolvable via Vagrant
plugins and other tweaks.)

erhudy:
The https://github.com/bloomberg/chef-bcpc/tree/feature/vmware-support-for-vagrant branch is where I am working on the VMware support. The main issue remaining to be resolved there is setup and teardown of additional disk images; setup is easy but teardown is harder because Vagrant does not surface the UUID of the VM at any point, so there is no way to keep the VMDKs with the VM directory and rely on them getting deleted at VM teardown like we do with Vagrant. I'm going to solve that with vagrant-triggers.

First, though, I am working on the existing Vagrant+VirtualBox setup on https://github.com/bloomberg/chef-bcpc/tree/feature/vagrant-bootstrap-0.2 to resolve a number of issues that didn't make it for the first cut of the new Vagrant stuff, plus it resolves some other issues that we've found using it. My OOO here is to finish this branch first, then rebase and finish the VMware branch (and then after that do another update that integrates Packer into the build process to generate Vagrant boxes for both providers and also VirtualBox OVFs for non-Vagrant builds, so that all build paths are using the same base image).

PR #685, opened 2015-06-29
Modify Nova policy to allow only supported actions

semenovroman:
This switches *default* policy from 'allow everyone' to 'allow only admin'. Only a small subset of rules, the ones we officially support, was set to "rule:admin_or_owner", or "".

please note: most of the file (all policies with "role:admin") can be safely erased since default policy is "role:admin". I left those policies to allow easier changes in the future and preference for explicit declarations.


erhudy:
Will test.

erhudy:
Behavior is weirdly counterintuitive when trying to create an image in Horizon, but there may not be much we can do about that. The **Create Image** button is still present in Horizon and lets you fill the fields out, but as soon as you commit it goes to status *Killed*.

erhudy:
Problem: cannot boot an image from a volume with error **Error: Policy doesn't allow compute:create:attach_volume to be performed. (HTTP 403)**

erhudy:
Can create volumes, just can't attach them.

semenovroman:
This PR does not include restrictions related to images, only nova. I will create a new PR for glance policy with appropriate restrictions. Will fix attach_volume.

erhudy:
Okay, `attach_volume` seems fine now when creating new instances. I can't manage attachments for volumes still, so for example you can't attach/detach another volume you've created to/from an instance.

erhudy:
```
vagrant@bcpc-vm1:~$ nova volume-attach eee gggggg
ERROR (Forbidden): Policy doesn't allow compute_extension:volumes to be performed. (HTTP 403)
```

erhudy:
Attaching/detaching seems to work fine now.

erhudy:
I think everything is good here now; looking for one more confirmation before merging.

+1

caiush:
I had a go at it. Looked good, okay to merge.

semenovroman:
Should I squash those 3 commits into 1, before merge happens?

erhudy:
That would be ideal.

PR #687, opened 2015-06-30
Ephemeral + LVM

erhudy:
Opening on behalf of @cloudm2 to consolidate two previous PRs.


cookbooks/bcpc/files/default/nova-ephemeral.py
```diff
@@ -0,0 +1,270 @@
...
+        LOG.error('Hosts list is empty and option --hosts-all=False (default value).')
+        return None
+
+    if verbose:
+        LOG.info('Starting create function.')
+    ret_value = False
+    id = None
+
+    # Step 1: Gather list with IDs and check for existence.
+
```
erhudy: Code style nit: this would be better rendered as the list comprehension `[agg.name for agg in nova.aggregates.list()]`.

cookbooks/bcpc/files/default/nova-ephemeral.py
```diff
@@ -0,0 +1,270 @@
...
+            # Already exists but still passed earlier check
+            LOG.warn('Aggregate %s already exists.' % aggregate_name)
+    else:
+        for agg in aggregate_list:  # Find agg by name
+            if agg.name == aggregate_name:
+                break
+
+    if not hosts:
+        # Build host list from hypervisor list
+        host_list = nova.hypervisors.list()
```
erhudy: Same as above: this would be better rendered as the list comprehension `hosts = [host.hypervisor_hostname for host in nova.hypervisors.list()]` (though there is a verbose statement in the for loop that would be difficult to render inside a list comprehension).

cookbooks/bcpc/recipes/nova-ephemeral-lvm.rb
```diff
@@ -0,0 +1,79 @@
...
+
+devices = device_list.join(' ')
+
+bash "setup-lvm-pv" do
+  user "root"
+  code <<-EOH
+    pvcreate #{devices}
+  EOH
+  not_if "pvdisplay | grep '/dev'"
+end
```
erhudy: I ran into a weird issue with this resource when testing; I specified 7 drives in `['bcpc']['nova']['ssd_disks']`, but only 3 had PVs created and added to the volume group. I need to dig a little further.

cookbooks/bcpc/files/default/nova-ephemeral.py
cloudm2: Yep I know. Had logging in for debugging and took them out but did not refactor. Short list.

roles/BCPC-Headnode.json
```diff
@@ -27,7 +27,6 @@
       "recipe[bcpc::powerdns-nova]",
       "recipe[bcpc::heat]",
       "recipe[bcpc::horizon]",
-      "role[BCPC-Worknode]",
```
erhudy: In the interests of keeping PRs focused, I'm going to revert this change, because we shouldn't decide to remove work node capabilities from head nodes here (though I am in favor of the change itself).

cookbooks/bcpc/files/default/nova-ephemeral.py
cloudm2: Yep.

cookbooks/bcpc/recipes/nova-ephemeral-lvm.rb
erhudy: Found the cause; there was an `override_attributes` in the role left over from debugging. Removed.

erhudy:
Tested this PR on virtualized without creating any ephemeral nodes to look for obvious regressions, but everything seemed to be in order there. Will expand to ephemeral next.

cookbooks/bcpc/files/default/nova-ephemeral.py
```diff
@@ -0,0 +1,270 @@
...
+            if verbose:
+                LOG.info('hosts value not passed. Building hosts list: %s' % host)
+
+    new_hosts=[]
+    # Attempt to add using hosts list.
+    for host in hosts:
+        try:
+            tmp = nova.aggregates.add_host(agg, host)
+        except exceptions.Conflict, e:
+            LOG.warn('Host aggregate %s for %s aggregate already exists (1).' % (host, aggregate_name))
```
erhudy: This didn't seem to fire for me (assuming that this is the exception that gets thrown if the host is already in another host aggregate). I got a pretty general error about the server not being able to perform the requested function or some such thing if the host was already in another agg.

erhudy:
General request for feedback: do we agree that nodes are either for OSDs or for ephemeral disks, or is the plan something else? There are a few recipes that should be cleaned up to get rid of a couple of assumptions if it's an either/or proposition (chiefly a few recipes that include **ceph-work** even though it's already in the role run list).

cookbooks/bcpc/files/default/nova-ephemeral.py
cloudm2: Conflict exception is fired when the host is already present in the same aggregate. Hosts can belong to different aggregates.

erhudy: Okay, it seems like it got a bit grumpy there then. All the hosts were in the persistent aggregate, and I kept getting the aforementioned error until I removed the host I was working on from that persistent aggregate.

cloudm2:
IMHO, head nodes should not have any compute or OSDs and if possible ceph mons. They should be only controllers. That said, I understand the point but ephemeral can be on any compute node regardless of OSDs existing on the same node if the nodes have capacity. If we don't have the extra drives then we will have to reduce the number of OSDs on the node or remove them completely so that the drives can be used for ephemeral usage. How are the new "compute" project nodes going to be configured?

mihalis68:
Just for reference, I want the ability to (continue to) run OSDs and compute on head nodes in small cluster context, both VM and "office desk".

caiush:
For now I'm okay with having the either or option, all OSD node or all ceph. When we have more experience on how people want to use this we can consider adding some capability to mix back in. For now we can go 100% either way. 

I agree that (long term) we should remove compute and OSDs front he head nodes, but for now lets leave them as regular OSD nodes. If nothing else it makes it much easier to bootstrap the cluster.

cloudm2:
No need to keep this open any longer since they both have been rewritten.

PR #688, opened 2015-06-30
Stuck carbon-relay fix

kelvk:
carbon-relay sometimes remain stuck even when destinations (i.e, carbon-cache) have recovered from lag/unresponsiveness. This patch applies Graphite Carbon PR 400 which fixes a race condition in carbon client queue signalling.

### DIFF
--- a/cookbooks/bcpc/files/default/graphite-carbon-client.patch
+++ b/cookbooks/bcpc/files/default/graphite-carbon-client.patch
@@ -1,0 +1,18 @@
+--- graphite/lib/carbon/client.py.orig	2014-12-31 16:49:45.000000000 +0000
++++ graphite/lib/carbon/client.py	2015-07-07 15:07:50.844870514 +0000
+@@ -1,3 +1,5 @@
++# THIS FILE PATCHED BY BCPC
++
+ from twisted.application.service import Service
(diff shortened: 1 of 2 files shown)


erhudy:
One suggestion on the method of detecting whether the patch has been applied: for newer patches I've been adding a comment line like `# THIS FILE PATCHED BY BCPC` to the top of the file, and using the presence/absence of that line to determine whether to apply the patch. This has the benefit of notifying us if a package upgrade replaces the patched file, since Chef will attempt to reapply the patch and likely fail if that file itself was changed. It will also simply reapply the patch if the file was restored back to its original state.

erhudy:
More holistically, we should probably write a patch LWRP to handle patching in a nice way, since right now we're just copying and pasting around the same block of shell script.

kelvk:
How about uploading patched files and calling `cookbook_file`?

erhudy:
Testing.

erhudy:
Bounced carbon-cache repeatedly and carbon-relay reconnected as expected each time.

kelvk:
Talked with @erhudy offline that we will test for BCPC signature in target file. Commit updated.

erhudy:
Re-testing with changed patch format.

PR #689, opened 2015-06-30
Parameterize Java options for Elasticsearch

kelvk:
Garbage collection logging will be enabled by default. No change to default heap size (256m).


erhudy:
Is 256MB enough heap size for Elasticsearch to do anything useful? In my experience it requires considerable amounts of memory.

kelvk:
The value will almost certainly be overridden in production deployments (bare-metal). I am only keeping it low by default for local virtual environments.

erhudy:
Testing.

erhudy:
Looks fine to me.

PR #693, opened 2015-07-08
Making all Keystone backends configurable

erhudy:
This PR is to make it easy to swap Keystone backend drivers out (mainly to swap between SQL and LDAP for specific drivers).


erhudy:
I built my own PR without issue (and it's also being actively used on current hardware, which is why I originally wrote this to accommodate some use cases on test hardware). It's mostly a no-op because all of the settings in the attribute defaults are what Keystone uses by default, *except* for the role driver which is `None` by default, but roles seem to work fine.

kelvk:
Keystone discovery, tenancy/user creation and role assignments work as expected.

PR #694, opened 2015-07-09
Tighten cinder policy

semenovroman:
Disallow cinder backup since it is not supported.


erhudy:
Testing.

erhudy:
Works as promised.

PR #697, opened 2015-07-10
Extend iptables for monitoring

kelvk:
On clusters that monitor (aka collect logs/metrics and alert on) clients outside of the cluster, current iptables rules are insufficient. As number of clients may be large, and if their IP addresses are not contiguous, `ipset` will be used here to prevent bloated iptables rulesets.


erhudy:
Seems to apply fine for me on an existing cluster build.

erhudy:
Stepped through things affected by the modified firewall rules and verified that the Zabbix server was still getting data from agents, new log information is making it into Elasticsearch and viewable in Kibana, and new metrics are showing up in Graphite.

PR #700, opened 2015-07-14
New host aggregates LWRP

caiush:
Create a general_compute host aggregate, which *does not* include headnodes. 

This PR lays the ground work for a several things:
* Clearer way to have ephemeral nodes, with their own flavors 
* Creating an AZ per failure domain, rather than one whooping lump of compute
* The ability to have ring fenced resources for some tenants 

Also I noted that the filters were way out of date, so updated them.

I had to create a new role which is designed to be a base role. Chef uses addition for merging attributes, so if BCPC-Headnode is a derivative of BCPC-Worknode, there is no way to clobber the aggregate membership array. This is a tad ugly, but on balance I think the overall ugliness is less.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -536,7 +536,8 @@ default['bcpc']['nova']['live_migration_patch'] = false
 # Nova debug toggle
 default['bcpc']['nova']['debug'] = false
 # Nova scheduler default filters
-default['bcpc']['nova']['scheduler_default_filters'] = ['AggregateInstanceExtraSpecsFilter', 'AvailabilityZoneFilter', 'RamFilter', 'ComputeFilter', 'DifferentHostFilter', 'SameHostFilter']
+default['bcpc']['nova']['scheduler_default_filters'] = ['AggregateInstanceExtraSpecsFilter', 'RetryFilter', 'AvailabilityZoneFilter', 'RamFilter', 'ComputeFilter', 'ComputeCapabilitiesFilter', 'ImagePropertiesFilter', 'ServerGroupAntiAffinityFilter', 'ServerGroupAffinityFilter']
+
 default['bcpc']['nova']['quota'] = {
   "cores" => 4,
   "floating_ips" => 10,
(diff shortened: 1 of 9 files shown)


cookbooks/bcpc/attributes/default.rb
```diff
@@ -536,7 +536,8 @@
 # Nova debug toggle
 default['bcpc']['nova']['debug'] = false
 # Nova scheduler default filters
-default['bcpc']['nova']['scheduler_default_filters'] = ['AggregateInstanceExtraSpecsFilter', 'AvailabilityZoneFilter', 'RamFilter', 'ComputeFilter', 'DifferentHostFilter', 'SameHostFilter']
+default['bcpc']['nova']['scheduler_default_filters'] = ['AggregateInstanceExtraSpecsFilter', 'RetryFilter', 'AvailabilityZoneFilter', 'RamFilter', 'ComputeFilter', 'ComputeCapabilitiesFilter', 'ImagePropertiesFilter', 'ServerGroupAntiAffinityFilter', 'ServerGroupAffinityFilter']
```
erhudy: New filters added here:
* ComputeCapabilitiesFilter
* ImagePropertiesFilter
* RetryFilter
* ServerGroupAffinityFilter
* ServerGroupAntiAffinityFilter

caiush: Indeed the ones we had are deprecated. Also I think the RetryFilter is def worth having and the ServerGroup, but that takes some user education to be used correctly.

cloudm2:
Same as https://github.com/bloomberg/chef-bcpc/pull/687 except for the lvm portion. The main difference being ruby vs python I guess. You can set individual hosts to whatever you want or match existing persistent. I used python and structured it so it could easily be added as a ansible module or ran as cli or from any tool.

roles/BCPC-Worknode.json
```diff
@@ -4,24 +4,13 @@
...
-      "recipe[bcpc::tpm]",
-      "recipe[bcpc::checks-work]",
-      "recipe[bcpc::zabbix-agent]"
+      "role[BCPC-Compute]"
     ],
     "description": "A functional compute node in a BCPC cluster",
     "chef_type": "role",
     "override_attributes": {
+      "bcpc" : {
+            "aggregate_membership" : ["general_compute"]
```
erhudy: I assume that it's intentional that this flag is set on the Worknode role instead of the Compute role to decrease compute load on the head nodes?

caiush: aye

erhudy:
Spun it up a couple of different times. This time I was able to create another flavor without the `extra_specs` key without the previously witnessed explosion.

erhudy:
Modifying a flavor in the environment after its initial creation doesn't work because the provider only checks for the name. In practice this probably isn't a big deal.

erhudy:
That is, it doesn't update the flavor in-place, but it doesn't cause a crash.

caiush:
It was actually by design. I consider that a v2 ability, its not hard to add but what we have is a good first step.

erhudy:
Yes, I agree that it's not problematic as is. I haven't had issues with this PR and am fine with merging.

+1

caiush:
I should explain. openstack doesnt allow modification of flavors (other than extra specs). You have delete them and recreate them. So Its not actually limiting functionality.

erhudy:
I see, that makes sense, in an **ಠ⸏ಠ OpenStack ಠ⸏ಠ** kind of way.

PR #701, opened 2015-07-14
Install logtail for Zabbix checks

kelvk:


kelvk:
Logtail is used in https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/templates/default/zabbix_openstack.conf.erb#L19-L24.

PR #702, opened 2015-07-14
New recipe for dumping dev only tools in

caiush:
too old to learn vi now...


erhudy:
do we have enough disk space for a full emacs install

PR #703, opened 2015-07-14
Numerous updates to Vagrant bootstrap process

erhudy:
- allow Vagrant to access bootstrap configuration even when not run from
  BOOT_GO.sh
- move default configuration variables out of BOOT_GO.sh
- regularize certain shared scripts in advance of future work on
  non-Vagrant build path
- consolidate multiple Vagrant configuration blocks into one
- improve validation of additional CA certificates in Vagrantfile
- verify proxy configurations from within bootstrap node
- updates to documentation and example apt-mirror configuration


erhudy:
The diff didn't correctly capture that everything under `bootstrap/common_scripts` was moved to `bootstrap/shared` and also had their names changed to begin with `shared_` for consistency (e.g., `bootstrap/common_scripts/common_build_bins.sh` was renamed to `bootstrap/shared/shared_build_bins.sh`).

erhudy:
I have validated the modifications to the Vagrant process by building:
- without proxies, with local mirror
- with proxies, with local mirror
- with proxies, without local mirror

erhudy:
Since it was asked, this PR doesn't change over to using `cluster.txt` to figure out how to build the cluster. That will come at some indeterminate future date when @mihalis68 and I have agreed on a way forward for consolidating the Vagrant and PXE boot paths.

kelvk:
Tested successfully with proxies and without local mirror. +1 from me.

mihalis68:
I read through this and it all looks good to me.

PR #707, opened 2015-07-15
Reinstate nova and rgw functional checks

kelvk:
Reinstate Nova and RadosGW functional checks so we have a "service-oriented" view of the cluster health via Zabbix. The repetitive deletes invoked by nova check is removed as it should not be necessary for Kilo.

I have tested this on a fresh cluster. The crontab should run on the head node(s) like so:

Jul 15 16:00:01 vagrant-ubuntu-trusty-64 CRON[19824]: (zabbix) CMD (zabbix_sender -c /etc/zabbix/zabbix_agentd.conf --key 'check.nova' --value `check -f timeonly nova` 2>&1 | /usr/bin/logger -p local0.notice)
Jul 15 16:00:01 vagrant-ubuntu-trusty-64 CRON[19826]: (zabbix) CMD (zabbix_sender -c /etc/zabbix/zabbix_agentd.conf --key 'check.rgw' --value `check -f timeonly rgw` 2>&1 | /usr/bin/logger -p local0.notice)

Once the cron jobs run after the monitoring node is up, the results should be visible in Zabbix dashboard.

### DIFF
--- a/cookbooks/bcpc/files/default/checks/nova
+++ b/cookbooks/bcpc/files/default/checks/nova
@@ -57,15 +57,11 @@ class TestNovaCompute(object):
 
 
     def _delete_server(self, server):
-        """Nassty hack, current stack requires >1 delete"""
-        start = time.time()
-        while time.time() - start < self.config["timeout"]:
-            try:
-                server.delete()
-                time.sleep(2)
-            except novaclient.exceptions.NotFound:
-                return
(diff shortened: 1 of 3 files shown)


erhudy:
Built out with this PR, confirmed that the cron job is there and results for **Functional check error: nova** and **Functional check error: rgw** are present in the Zabbix overview.

PR #708, opened 2015-07-15
Fix Zabbix agent check errors

kelvk:
First stab at fixing various errors reported in /var/log/zabbix/zabbix_agentd.log.

The removal of Zabbix configuration items is not automatic, as `deleteExisting` rule for Zabbix configuration import is only available from [2.4](https://www.zabbix.com/documentation/2.4/manual/api/reference/configuration/import).


erhudy:
Only error I see in `zabbix_agentd.log` now is complaining about the Ceph admin socket, which we already know about. Is this the expected result?

kelvk:
Yes. This PR is a step towards a complete cleanup.

Issue #683, opened 2015-06-29
Break out hypervisor domain and cluster VM domain into separate attributes


erhudy:
After some confusion regarding whether `['bcpc']['domain_name']` was meant to represent the DNS domain of cluster nodes as well as VMs running on the cluster. The conclusion was it was not, and so:

* there should be a second attribute set up for the purpose of representing the baremetal domain (e.g., `['bcpc']['metal_domain']` or `['bcpc']['hypervisor_domain']`)
* `['bcpc']['domain_name']` should probably be renamed to something that makes it clearer it is meant to be the domain name for cluster services and VMs rather than cluster members

The cluster domain should be used in places like the hosts file on each server.

erhudy:
@mihalis68 and I are testing a possible solution in https://github.com/bloomberg/chef-bcpc/tree/bugfix/separating-cluster-and-hypervisor-domains. If everything works out, I will PR it.

PR #709, opened 2015-07-16
Switch to using fqdn endpoints

semenovroman:
This PR makes keystone list API endpoints using fqdn instead of ip addresses (makes ssl setup significantly cleaner).

### DIFF
--- a/cookbooks/bcpc/templates/default/hosts.erb
+++ b/cookbooks/bcpc/templates/default/hosts.erb
@@ -13,8 +13,9 @@ ff00::0 ip6-mcastprefix
 ff02::1 ip6-allnodes
 ff02::2 ip6-allrouters
 
+<%=node['bcpc']['management']['vip']%> openstack.<%=node['bcpc']['cluster_domain']%>
 <% @servers.each do |server| -%>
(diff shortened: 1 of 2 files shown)


erhudy:
Have you tried to see whether this works when doing a local cluster build?

semenovroman:
So far I just tried re-chefing, will be rebuilding from scratch in couple hours.

mihalis68:
This depends on https://github.com/bloomberg/chef-bcpc/issues/683 which is marked as an enhancement but I'd call a bug. So I would prefer not to merge this until 683 is resolved.

erhudy:
I suspect this will not work as is for local builds because it will require a hostfile entry for `openstack.#{node['bcpc']['hypervisor_domain']}` to be added to `/etc/hosts`, since we're fudging DNS locally. It should probably be added in any event since API calls are so frequent and would save occasional trips out to DNS.

erhudy:
Tagging **Do not merge** until the cluster/hypervisor domain split is complete.

semenovroman:
@erhudy you are right, it does not work without `/etc/hosts` entry. Works fine with the entry (https://github.com/semenovroman/chef-bcpc/commit/e9eecd47258ee2ff67f42626b52dd1b0c0fd73f8)

semenovroman:
Updated PR to use 'cluster_domain'\'hypervisor_domain'.

semenovroman:
Closing this one, will create a new one based on newer master.

PR #711, opened 2015-07-16
Parameterize innodb_buffer_pool_size

kelvk:
This is a no-op on existing cluster.


mihalis68:
I reviewed this and see that as @kelvk says this will be a no-op unless a cluster builder adds a variable to the chef attributes, in which case they presumably know what they are doing. This is also very straightforward, so... looks good to me.

PR #714, opened 2015-07-17
Use post 0.50 cephx config

caiush:
the auth_support config has been deprecated since 0.50. 

There is a bunch of things not right/fishy about ceph auth in bcpc, will work through them.


erhudy:
Do you want this PR to reviewed and merged as is or do you plan on adding additional commits to fix other parts of the config first?

caiush:
No this can be reviewed and merged, the other fixes are in #720

erhudy:
I cheffed this onto an existing cluster and everything seemed fine.

+1

PR #715, opened 2015-07-17
Fixes #713 - no longer requires config overrides file to exist

### LINKED ISSUE #713: BOOT.sh expects overrides file
When running a new install:
```
There was an error loading a Vagrantfile. The file being loaded
and the error message are shown below. This is usually caused by
a syntax error.

Path: /Users/xaius/github/bloomberg/chef-bcpc/bootstrap/vagrant_scripts/Vagrantfile
Message: No such file or directory - /Users/xaius/github/bloomberg/chef-bcpc/bootstrap/config/bootstrap_config.sh.overrides
```

erhudy:
Additional semi-related change in this PR is to require that the defaults file exist in `BOOT_GO.sh`, so that it won't try to plow ahead when required environment variables are unconfigured.


mihalis68:
Looks good to me

PR #718, opened 2015-07-18
tag all routes with proto static

mihalis68:
According to http://www.policyrouting.org/iproute2.doc.html#ss9.16
routes should be tagged with protocol type - perhaps this will make
our routes more stable


mihalis68:
I tested this with a VM cluster - seems to be working fine.

mihalis68:
I am interested in @pchandra's opinion on this

pchandra:
This all makes sense and is a good thing to add to pin those routes as static. One small side point is that I don't think you need to mark the 3 routes added to the 'main' table as static since that script removes them at the end, but it certainly shouldn't hurt anything, so it's OK to leave this PR as is.

mihalis68:
I actually thought about that! I couldn't decide between consistency (ALL routes we create are static) and the thought that pinning ones we then kill shortly afterwards is sort of misleading, obscures the ones we intend to keep.

PR #719, opened 2015-07-20
Reinstate Diamond 4.x

kelvk:
DIamond was upgraded to 4.x (#614) but it appears to have regressed.


bootstrap/shared/shared_prereqs.sh
```diff
@@ -83,7 +83,7 @@ download_file cirros-0.3.4-x86_64-disk.img http://download.cirros-cloud.net/0.3.
 download_file kibana-4.0.2-linux-x64.tar.gz https://download.elastic.co/kibana/kibana/kibana-4.0.2-linux-x64.tar.gz
 # unfortunately GitHub ZIP files do not contain the actual Git index, so we must use Git to clone here
 if [[ ! -f $BOOTSTRAP_CACHE_DIR/diamond_downloaded ]]; then
-  git clone https://github.com/BrightcoveOS/Diamond $BOOTSTRAP_CACHE_DIR/diamond
+  git clone https://github.com/python-diamond/Diamond $BOOTSTRAP_CACHE_DIR/diamond
   touch $BOOTSTRAP_CACHE_DIR/diamond_downloaded
```
erhudy: Since the source is changing, I recommend changing this file as well to something new (otherwise the caching script will see `diamond_downloaded` exists, skip the checkout, then the install will blow up later on because the requested revision will be missing).

erhudy: This also blows up if the diamond path already exists, so it may need to be cloned into a new directory, or some mechanism needs to be added to test if it's the old one or the new and then remove the old directory.

kelvk:
Commit updated.

kelvk:
While we are doing this, bump Diamond version by a few notches to fix a RabbitMQ collector bug. For in-place upgrade, remove `cookbooks/bcpc/files/default/bins/diamond.deb` and re-run `cookbooks/bcpc/files/default/build_bins.sh`.

erhudy:
Tested and seems fine; installed Diamond 4.0.181.

PR #725, opened 2015-07-21
Time how long it takes to deploy a cluster

semenovroman:
...
Thanks for using BCPC!
Finished in 2854 seconds


bootstrap/vagrant_scripts/BOOT_GO.sh
```diff
@@ -85,3 +85,5 @@ $REPO_ROOT/bootstrap/shared/shared_configure_chef.sh
 
 # Dump out useful information for users.
 $REPO_ROOT/bootstrap/vagrant_scripts/vagrant_print_useful_info.sh
+
+echo "Finished in $SECONDS seconds"
```
erhudy: I was going to ask where `$SECONDS` is coming from, until Google revealed to me that it's built into bash. Clever!

I usually time how long a build run takes by running `BOOT_GO.sh` with `time`, but this will save a little extra typing.

PR #727, opened 2015-07-22
Update rabbitmq-server to version 3.5.4-1

semenovroman:
It does not install with version 3.5.3-1, failing with 'not found' error.


mihalis68:
As it happens I hit the same thing and saw 3.5.4-1 hence I have tested this.

PR #720, opened 2015-07-20
Two CephX changes: Cinder/Glance CephX clients and perms on admin sockets

caiush:
This PR does two things:
* Previously we were using the admin key for every ceph operation. This key has the following perms:
```
	caps mds = "allow"
	caps mon = "allow *"
	caps osd = "allow *"
```
For various security reasons this seems like a bad idea. In a future PR I will remove the admin key from the compute nodes. However, there may be monitoring consequences as you will be unable to do things like `ceph health` without passing your own key. If indeed this is a problem we can create a readonly key for mons. Monitoring should however be able to get everything it needs through the admin sockets.  @kelvk will have to comment.
* Secondly, ceph was configured to create an admin socket for *all* ceph clients. This includes the admin client (which *should* just be used to do admin tasks on the cluster, see above). Now we break down the configuration of the clients based on type. So cinder and glance clients create sockets, but things like `ceph health` does not. I found that this functionality was actually broken for any of the useful clients anyhow. So we weren't even creating admin sockets for cinder clients (no matter which cephx client they use). This is a simple fix to apparmor config.  Fixes #652 

This should be able to be cheff-ed in to an already running BCPC cluster, perhaps bouncing openstack services will be required.

PR #728, opened 2015-07-22
Organize Ceph and RGW Zabbix checks

kelvk:
The following changes are introduced with this PR:
1. Increase RGW logging so HTTP return codes are appropriately logged. This will be in place until http://tracker.ceph.com/issues/12432 is resolved.
2. Monitor Civetweb logs if it's the RGW frontend in use.
3. Monitor `ceph health` on head nodes only.
4. Remove unnecessary Zabbix items (i.e. HTTP metrics) as these are better suited to be counted and presented in Kibana/Graphite.

In Zabbix [2.2.6](http://www.zabbix.com/rn2.2.6.php), `logtimefmt` field is added to all exported items. Since `zabbix_bcpc_templates.xml` was generated while we were on 2.2.2, the field addition is expected and can be safely ignored.

For the updated Zabbix templates to take effect on a running cluster, one needs to drop `BCPC-Headnode` and `BCPC-Worknode` templates (ensuring `delete selected with linked elements` is checked), then re-associate hosts with the appropriate templates. This is clearly not ideal. I am increasingly tempted to upgrade Zabbix to [2.4](https://www.zabbix.com/documentation/2.4/manual/introduction/whatsnew240#option_to_remove_missing_resources_when_importing) to address this shortcoming. Something for another PR.

### DIFF
--- a/cookbooks/bcpc/files/default/zabbix_bcpc_templates.xml
+++ b/cookbooks/bcpc/files/default/zabbix_bcpc_templates.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <zabbix_export>
     <version>2.0</version>
-    <date>2015-06-10T21:11:39Z</date>
+    <date>2015-07-22T19:49:26Z</date>
     <groups>
         <group>
             <name>Templates</name>
@@ -18,6 +18,9 @@
             </groups>
             <applications>
                 <application>
+                    <name>Ceph</name>
+                </application>
+                <application>
                     <name>Functional Checks</name>
                 </application>
                 <application>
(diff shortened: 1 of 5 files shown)


kamidzi:
Build's taking forever. Testing with #732

kelvk:
Expanding on test detail as suggested by @kamidzi:

1. Most of the testing/verification is by eyeballing https://10.0.100.6/zabbix/latest.php. `ceph health` should only be associated with the head nodes.
2. A bunch of RGW "Bucket*" and RGW log items no longer exist for all nodes (exact items are in the diff)
3. `/etc/zabbix/zabbix_agentd.d/zabbix-rgw.conf` now exists on all RGW nodes (head+work). Log checks are added if Ceph frontend is Civetweb, which is the BCPC default. Monitoring Apache RGW logs is excluded as it is expected to be deprecated per #626.
4. RGW will log HTTP status codes in `/var/log/ceph/radosgw.log` like so, due to above-mentioned Ceph bug:
<pre>
2015-07-29 15:06:06.815257 7f22367fc700  1 ====== req done req=0x39402c0 http_status=200 ======
</pre>
5. This PR cleans up various Ceph/RGW errors in `/etc/zabbix/zabbix_agentd.log`. If #720 is merged as well, the log should be totally clean.

kamidzi:
Looks good. Thanks.

PR #729, opened 2015-07-23
Parameterize PHP settings for Zabbix

kelvk:


erhudy:
It looks like these settings are also exposed in `/etc/apache2/conf-enabled/zabbix.conf`. Do we need to manage that file and parameterize things there as well?

kelvk:
Reorganized this and pushed an update.

erhudy:
Tested out and looks good.

PR #731, opened 2015-07-23
Restrict image upload

semenovroman:
Modify glance policy to restrict image upload to admins only.


erhudy:
Verified that image creation/uploading is locked out through the API and that the Create Image button is unavailable in Horzion.

PR #732, opened 2015-07-24
Disable metrics that are unlikely to be consumed

kelvk:


kamidzi:
Build's taking forever, so I'll finish up later

PR #733, opened 2015-07-24
default sec group rules for admin tenant

mihalis68:
Replicate the default secgroup rules into the admin tenancy since the
default rules that we establish here don't get applied to pre-existing
tenancies.

### DIFF
--- a/cookbooks/bcpc/recipes/nova-setup.rb
+++ b/cookbooks/bcpc/recipes/nova-setup.rb
@@ -28,24 +28,42 @@ bash "wait-for-nova-to-become-operational" do
 end
 
 
-bash "nova-configure-default-secgroup-rules" do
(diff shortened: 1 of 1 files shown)


erhudy:
Dare I ask if this branch could be overloaded to fix the guards on the new resource and also the default rules resource above to verify that both the ICMP and TCP rules exist?

mihalis68:
Sure they would be good to have. Will do.

mihalis68:
I've updated this as suggested and retested on a new build.

erhudy:
Verified that rules are in place for the admin tenant, that VMs launched immediately after cluster build without any Horizon monkeying are pingable and SSH-to-able, and that the guards prevent adding the rules over and over again.

PR #736, opened 2015-07-25
Ephemeral storage option

caiush:
Basically a tidied up version of #687, but with one important difference. The volumes are directly creating in LVM. The previous PR actually used LVM to create a file system then put a file system over that. The LVM drive is considerably more performant (from what I have read).

I also explicitly create the flavors using the environment and the osflavor LWRP and give the ephemeral disk extra scratch space. 

Future interesting features:
* I believe that you can tell nova to format the ephemeral block dev, so it would appear mounted and available as scratch. This is probably worth doing, or at least finding out if cloud-init can be configured to do it. 
* Cinder often needs a kick after the head node run, I think this is an issue with cephx auth and will probably be fixed by #720.


cloudm2:
I know they claim it but I find it hard to believe that nova lvm is more performant. Nothing in openstack is more performant than native.

caiush:
the LVM driver doesn't have the overhead of the file system, thats all.  It has to be faster, if its less stable.. TBD... but its pretty mature, been in openstack since E for F (don't recall from  memory).

cloudm2:
You still have the normal pvcreate and vgcreate. The only thing missing in the recipe is lvcreate which I would believe the nova "driver" (not a real driver I'm sure - haven't looked into the code) will call since nothing in Openstack is really performant. Not throwing a wet blanket but I don't understand why spent more cycles on this when not really needed.

caiush:
I rebased from master and tested. awaiting independent test.

erhudy:
On a 6-node local build (1 head node, 1 RBD work node, 1 ephemeral work node, 3 monitoring nodes), Ceph looks pretty unhappy:
```
    cluster bb93c222-d3e4-4643-8f09-8396e6e9c28e
     health HEALTH_WARN
            2290 pgs degraded
            2290 pgs stuck degraded
            704 pgs stuck unclean
            2290 pgs stuck undersized
            2290 pgs undersized
            recovery 112/183 objects degraded (61.202%)
            recovery 7/183 objects misplaced (3.825%)
            too many PGs per OSD (624 > max 300)
     monmap e1: 1 mons at {bcpc-vm1=172.16.100.11:6789/0}
            election epoch 2, quorum 0 bcpc-vm1
     osdmap e166: 8 osds: 8 up, 8 in; 206 remapped pgs
      pgmap v4654: 2496 pgs, 18 pools, 40164 kB data, 61 objects
            454 MB used, 143 GB / 143 GB avail
            112/183 objects degraded (61.202%)
            7/183 objects misplaced (3.825%)
                1792 active+clean+undersized+degraded
                 498 active+undersized+degraded
                 206 active+remapped
```

erhudy:
Suffered a failure when cheffing the **BCPC-EphemeralWorknode** role onto a physical node with 7 available SSDs:
```
Recipe: bcpc::lvm
  * apt_package[lvm2] action install
    - install version 2.02.98-6ubuntu2 of package lvm2
  * bash[setup-lvm-pv] action run
    
    Error executing action `run` on resource 'bash[setup-lvm-pv]'
    
    Expected process to exit with [0], but received '5'
    ---- Begin output of "bash"  "/tmp/chef-script20150730-2000-18only2" ----
    STDOUT: 
    STDERR: Device /dev/sdb not found (or ignored by filtering).
      Device /dev/sdc not found (or ignored by filtering).
      Device /dev/sdd not found (or ignored by filtering).
      Device /dev/sde not found (or ignored by filtering).
      Device /dev/sdf not found (or ignored by filtering).
      Device /dev/sdg not found (or ignored by filtering).
      Device /dev/sdh not found (or ignored by filtering).
    ---- End output of "bash"  "/tmp/chef-script20150730-2000-18only2" ----
    Ran "bash"  "/tmp/chef-script20150730-2000-18only2" returned 5
    
    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/bcpc/recipes/lvm.rb
    
     23:   bash "setup-lvm-pv" do
     24:     user "root"
     25:     code <<-EOH
     26:     pvcreate #{ node['bcpc']['nova']['ephemeral_disks'].join(' ') }
     27:   EOH
     28:     not_if "pvdisplay | grep '/dev'"
     29:   end
     30: 
    
    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/bcpc/recipes/lvm.rb:23:in `from_file'
    
    bash("setup-lvm-pv") do
      action "run"
      retries 0
      retry_delay 2
      command "setup-lvm-pv"
      backup 5
      returns 0
      user "root"
      code "    pvcreate /dev/sdb /dev/sdc /dev/sdd /dev/sde /dev/sdf /dev/sdg /dev/sdh\n"
      interpreter "bash"
      declared_type :bash
      cookbook_name "bcpc"
      recipe_name "lvm"
      not_if "pvdisplay | grep '/dev'"
    end
```

erhudy:
The disks are available:
```
# ls -l /dev/sd*
brw-rw---- 1 root disk 8,   0 Jul 30 09:36 /dev/sda
brw-rw---- 1 root disk 8,   1 Jul 30 09:36 /dev/sda1
brw-rw---- 1 root disk 8,   2 Jul 30 09:36 /dev/sda2
brw-rw---- 1 root disk 8,  16 Jul 30 09:55 /dev/sdb
brw-rw---- 1 root disk 8,  32 Jul 30 09:55 /dev/sdc
brw-rw---- 1 root disk 8,  48 Jul 30 09:55 /dev/sdd
brw-rw---- 1 root disk 8,  64 Jul 30 09:55 /dev/sde
brw-rw---- 1 root disk 8,  80 Jul 30 09:55 /dev/sdf
brw-rw---- 1 root disk 8,  96 Jul 30 09:55 /dev/sdg
brw-rw---- 1 root disk 8, 112 Jul 30 09:55 /dev/sdh
```

erhudy:
Needed to completely destroy GPT/MBR partition tables on the disks with gdisk (install gdisk, then for each disk, do `gdisk /dev/sdx`, **x**, **z**, answer Y to both questions). Not sure if gdisk is scriptable, but it's faster than zeroing the entire disk to destroy all copies of the GPT.

erhudy:
Better way: `sgdisk -Z /dev/sdx`

erhudy:
Re-opening so that I can continue to test and comment here. I think some of the issues that were being had previously were because of conflicts with some other unfinished code.

erhudy:
Behavioral note: if you boot from image with an ephemeral flavor, both disks will be on LVM. If you boot from volume, you will get your root volume from Ceph and the ephemeral volume from LVM.

erhudy:
Rebased branch off current master and pushed a fix for calculating Ceph PGs, so that the recipe is only counting nodes that actually have OSDs on them.

erhudy:
Here are some informal results of quick and dirty testing on hardware (note that this is only using 21 OSDs, 7 per system, because of re-dedicating some hardware to ephemeral).

Test conditions
---
* e1.medium running Ubuntu 14.04.2 LTS with Ceph root partition on `/dev/vda` and LVM ephemeral partition on `/dev/vdb` (same as m1.medium, plus ephemeral disk)
* formatted `/dev/vdb` as ext4
* rest of cluster was quiesced to minimize traffic from anywhere else (so no noisy neighbor testing)
* readahead set to 2048KB on all block devices via `blockdev --setra`
* using `bonnie++`, ran 5 times with default settings as `ubuntu` (8GB test payload)

Results
---
* ephemeral is faster for reads: ~500MB/sec for ephemeral versus ~400MB/sec for Ceph, though Ceph has turned in slightly better results with larger Ceph installations
* Ceph is faster for block writes by a considerable margin: ~200MB/sec for ephemeral versus ~300MB/sec for Ceph, and Ceph again has turned in better results with larger installations
* most curiously, Ceph completely clowns ephemeral in terms of latency, which is counter-intuitive and is leading me to doubt the test methodology somewhat
  * ~125ms block writes for Ceph versus ~300ms for ephemeral
  * ~250ms block rewrites for Ceph versus ~850ms for ephemeral
  * ~9ms block reads for Ceph versus ~14ms for ephemeral
* re-testing ephemeral with the smaller default block size of 256KB makes it considerably more performant for both reads and writes, though with unusual effects on latency
  * from ~200MB/sec to ~260MB/sec for block writes
  * from ~500MB/sec to ~550MB/sec for block reads
  * block write latency decreases to a level roughly comparable to Ceph
  * block read latency increases to nearly 40ms
  * block rewrite latency accordingly gets absolutely terrible

Thoughts
---
* Better performance might be seen by using a RAID 0 device from onboard RAID or mdadm instead of an LVM JBOD, but that would increase the size of the failure domain and also make us dependent either on hardware RAID again or mdadm (and mdadm frustrates me).
* Ceph only has 3 nodes in this cluster, and so its performance numbers are likely to improve in larger clusters.
* While running the tests, the hypervisor showed unexpectedly high load average and iowait that would hover in the 10% range, occasionally spiking to 25%.
* Ephemeral volumes are apparently extremely sensitive to readahead settings, and so I will continue to test to try to find the optimal setting.

semenovroman:
out of curiosity, can you post `hdparm -t /dev/vdx` results?

erhudy:
`/dev/vda` is Ceph, `/dev/vdb` is ephemeral:
```
ubuntu@ip-1-100-0-132:/mnt/bonnie$ sudo hdparm -t /dev/vdb

/dev/vdb:
 Timing buffered disk reads: 1332 MB in  3.00 seconds = 443.89 MB/sec
ubuntu@ip-1-100-0-132:/mnt/bonnie$ sudo hdparm -t /dev/vda

/dev/vda:
 Timing buffered disk reads: 986 MB in  3.01 seconds = 328.03 MB/sec
```

erhudy:
Changing readahead both on the hypervisor and VM to 4096KB seems to produce the best results in general, though it does not get back to the block read latency results seen originally. I am testing outside the VM by running bonnie against an LV created on the hypervisor.

erhudy:
After testing JBOD for a bit, I switched over to using RAID 5 from the hardware RAID controller (6 data disks, 1 parity disk) and placed the LVM volume group over top of that single block device in Linux. It is _considerably_ faster: block read speeds jump to 1.5GB/sec and block write speeds to 900MB/sec; latency also improves. This definitely seems like the way we should be offering ephemeral volumes in order to set them apart from RBD volumes performance-wise.

cloudm2:
Interesting results. I suspected nothing through Nova would be faster. The hardware RAID controller is usually always faster and the only downside would be tying yourself closer to a specific vendor wouldn't it?

erhudy:
Also worth pointing out that CPU load on the hypervisor drops significantly and iowait drops to near 0% with hardware RAID (which you would expect).

erhudy:
Merging this because the basic concepts have been proven out successfully and the remaining questions all have to do with performance and aren't relevant to the content of this PR specifically.

PR #737, opened 2015-07-27
Enabled isolcpus

caiush:
restrict the kernel scheduler to a subset of the cpus on the host and force libvirt to use the others. 

Really we should enable the cpu pinning filter and do the full monte here, especially for deployments that are not doing overcommit, because the performance boost is supposed to be quite impressive.

Note I wasn't able to get this to work on virtual box(something about the boot loader), but was able to get it working on a hardware clusters. To deploy on hardware, use `virsh capabilities` to grab the cpu tags and the numa groups. then split up the cpus however you like,,, 

Note, currently there is no way to set isolcpus outside of the boot cmd params, so  *reboot is required*

### DIFF
--- a/cookbooks/bcpc/recipes/nova-work.rb
+++ b/cookbooks/bcpc/recipes/nova-work.rb
@@ -266,3 +266,17 @@ bash "patch-for-ip-hostnames-networking" do
     notifies :restart, "service[nova-compute]", :immediately
     notifies :restart, "service[nova-network]", :immediately
 end 
+
+#
+# Pin all host processes (including OSDs) to the following cpus
+#
+if not node['bcpc']['nova']['host_cpus'].nil?
+bash "set-cpu-isolation" do
+    user "root"
+    code <<-EOH
+        echo GRUB_CMDLINE_LINUX_DEFAULT=\\\"\\$GRUB_CMDLINE_LINUX_DEFAULT isolcpus=#{ node['bcpc']['nova']['host_cpus'].join(',')}\\\" >> /etc/default/grub
+        update-grub
+    EOH
+    not_if "grep 'isolcpus=#{ node['bcpc']['nova']['host_cpus'].join(',')}' /etc/default/grub"
+end
+end
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -518,6 +518,9 @@ default['bcpc']['keystone']['policy'] = {
 default['bcpc']['nova']['ram_allocation_ratio'] = 1.0
 default['bcpc']['nova']['reserved_host_memory_mb'] = 1024
 default['bcpc']['nova']['cpu_allocation_ratio'] = 2.0
+
+default['bcpc']['nova']['host_cpus'] = [0]
+default['bcpc']['nova']['vm_cpus'] = [1]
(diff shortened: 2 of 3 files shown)


caiush:
Just for discussion and for other people to play with, not ready to be merged.

semenovroman:
I am wondering why would there be a performance boost? I can see a small latency improvement on isolated cpus\cores, but I suppose anything latency sensitive would not be virtualized in the first place. On the other hand, by restricting available cpu subset, we limit raw "horsepower" (in cases where kernel cpus are idle) and interfere with turbo-boost (since kernel can't schedule cpu intensive task on a cooler cpu\core that can be turbo boosted while hot one is cooling down). p.s: this is my speculation that was not tested on real life scenario.

cookbooks/bcpc/recipes/nova-work.rb
```diff
@@ -266,3 +266,17 @@
...
+# Pin all host processes (including OSDs) to the following cpus
+#
+if not node['bcpc']['nova']['host_cpus'].nil?
+bash "set-cpu-isolation" do
+    user "root"
+    code <<-EOH
+        echo GRUB_CMDLINE_LINUX_DEFAULT=\\\"\\$GRUB_CMDLINE_LINUX_DEFAULT isolcpus=#{ node['bcpc']['nova']['host_cpus'].join(',')}\\\" >> /etc/default/grub
+        update-grub
+    EOH
+    not_if "grep 'isolcpus=#{ node['bcpc']['nova']['host_cpus'].join(',')}' /etc/default/grub"
```
erhudy: If `['bcpc']['nova']['host_cpus']` is changed, the resource will append another `isolcpus=` line to the GRUB configuration. Will GRUB tolerate this?

erhudy:
I can see there being a performance boost if there is no CPU overcommit, because the system can bypass the entire waiting-for-CPU step. We would need to test it under something replicating actual production hypervisor conditions, e.g., 50+ VMs on a host all not doing much to determine whether any possible performance gains there offset the possibility of being more heavily negatively impacted by a noisy neighbor if two VMs are sharing the same core pin and one of them goes highly active.

caiush:
So there is two things here. First off by isolating osds to a subset of cores not shared with vms the OSDs should always get cycles. This is less of an issue now we nice up osds but this maybe more effective. Second, in the future we may wish to support NUMA grouping on none over committed clusters. Others have reporting some pretty good perf boosts for mutlisocket nodes. I'll believe it when I see it. But the first point I think is kinda interesting. 

Against its just a pr to have a play with on hard are at some point see what it can give us.

cookbooks/bcpc/recipes/nova-work.rb
caiush: Meh, don't care. This is not intended as something to go into master at this time. Just enough to test the feature out. If it's interesting then we can fiddle with the guard

erhudy:
http://redhatstackblog.redhat.com/2015/05/05/cpu-pinning-and-numa-topology-awareness-in-openstack-compute/ was a blog post from Red Hat that this was probably based on.

erhudy:
Closing this PR because it's languished for a while. I am filing an issue to investigate isolcpus for future use.

caiush:
Actually I got it from the cern blog (or at least that started the thought process) about their numa work.

Issue #738, opened 2015-07-27
HTTPS proxy test during pre-flight build check fails when behind MITM proxy


kamidzi:
If attempting to build behind a MITM proxy, the HTTPS request test at https://github.com/bloomberg/chef-bcpc/blob/master/bootstrap/shared/shared_proxy_setup.sh#L14 may fail due to failed certificate validation.

kamidzi:
This shouldn't be a problem if the proper certs are added to the proper store.

PR #739, opened 2015-07-27
Also include certs in BOOTSTRAP_ADDITIONAL_CACERTS_DIR for HTTPS prox…

kamidzi:
…y test

Fixes bloomberg/chef-bcpc#738


kamidzi:
This was put on hold because SSL cert verification may also be required during the prerequisite download phase. An obvious solution  is to repeat the --cacert argument passing, but that seems a bit...hokey. Another, probably better approach is to define the CAPath and always pass this to curl. This implies processing ADDITIONAL_CACERTS_DIR with c_rehash, which has its own implications wrt the tests in the Vagrantfile for file extensions. Its not immediately clear whether update-ca-certifcates will mind about the symlinks

erhudy:
The presumption is that the system somebody's building on will already have any additional necessary CA certificates in their system store, and that curl will be using this store.

PR #724, opened 2015-07-21
FQDN endpoints in catalog, configs and .rc files

semenovroman:
Tested by creating a new cluster.

PR #740, opened 2015-07-27
Add ability to specify custom SSL keypair

semenovroman:
To test (pull fqdn branch first https://github.com/bloomberg/chef-bcpc/pull/724 , then merge this one):
Put SSL certificate and key (for example `ssl_cert.pem` and `ssl.key`) into `cookbooks/bcpc/files/default` directory, modify `attributes/default.rb` - set:
`default['bcpc']['ssl_certificate'] = "ssl_cert.pem"`
`default['bcpc']['ssl_private_key'] = "ssl.key"`

// Make sure certificate has cluster_domain as 'CN' or one of alternative names set (otherwise installation will fail with ssl vertification error in keystone test).


cookbooks/bcpc/recipes/certs.rb
```diff
@@ -26,6 +26,25 @@
     mode 00644
 end
 
+ruby_block "read-ssl-certificate" do
+    block do
+      begin
+        require 'openssl'
+        unless node['bcpc']['ssl_certificate'].nil? or node['bcpc']['ssl_private_key'].nil?
```
erhudy: `unless`+`or` is a little hard to mentally parse, because the negation only applies to the first part of the conditional.

erhudy:
Testing now.

erhudy:
Update with current wackity OpenStack bug that @semenovroman is looking at: when using proper certificates, Keystone works fine but anything that tries to use Keystone to get a token to access another service (Nova, Glance, Cinder, etc.) fails out with the error message **Invalid OpenStack Identity credentials**. Replacing the certs with self-signed certs makes the problem go away.

semenovroman:
Rewrote and '--force' pushed this branch. It is ready for testing now (and up to date with a master).

erhudy:
Worked fine on a 6-node local build, testing with real certs Elsewhere.

erhudy:
One request: please update the landing page template (index.html.erb) so that hostname-based URLs are shown everywhere for the **Web Interfaces** and **API Endpoints** sections.

erhudy:
Merging this now. There are a few outstanding issues that I will address in a separate branch.

PR #742, opened 2015-07-28
Changes Cinder RBD driver behavior with RAW image->volume

erhudy:
This patch alters the behavior of the Cinder RBD driver to greatly speed
operations where a user is creating a volume from a RAW image. Normally
the image is always piped through the API, but if the type of the source
image is RAW, it will instead execute rbd clone and rbd flatten
commands against Ceph and bypass the APIs. This is considerably faster
and more efficient than transferring the image using the APIs (in tests,
about 5-6 times faster).

This requires that the Ceph user Cinder is using to access the pools
have read capability to the images pool. I have added another
configuration item as well to make Cinder aware of the name of the
images pool, since that information was not previously available to
Cinder.

### DIFF
--- a/cookbooks/bcpc/files/default/cinder_volume_drivers_rbd.py.patch
+++ b/cookbooks/bcpc/files/default/cinder_volume_drivers_rbd.py.patch
@@ -1,0 +1,89 @@
+diff --git a/cinder/volume/drivers/rbd.py b/cinder/volume/drivers/rbd.py
+index 7dae2ec..d129daf 100644
+--- a/cinder/volume/drivers/rbd.py
++++ b/cinder/volume/drivers/rbd.py
+@@ -13,6 +13,8 @@
+ #    under the License.
+ """RADOS Block Device Driver"""
+ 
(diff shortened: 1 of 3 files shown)


erhudy:
To test this PR:
* create volumes from images using `cinder create --image-id` or via Horizon
* launch instances using *Boot from image (creates a new volume)*
* verify that image->volume clone operations are happening entirely in Ceph and are not going via the API (if things are working you will see high `rbd` CPU usage on the node where the Cinder API is running, if things are not working you will see high `glance-api` and `haproxy` CPU usage because the image is being piped through the API)
* do various volume operations (create, delete, resize, etc.)

The only thing this PR actually changes is the mechanism used to clone new volumes from images, but it won't hurt to validate general functionality.

caiush:
I tested this and everything functioned just fine and as designed/

erhudy:
Hold off on merge while investigating something to do with the Glance API version Cinder is using.

erhudy:
Closing this PR and will open a new one for Glance API version change.

PR #745, opened 2015-07-29
Enable Zabbix alert scripts path

kelvk:
No-op change. Enable the path where alert scripts are placed for [custom alerting](https://www.zabbix.com/documentation/2.2/manual/config/notifications/media/script).


kamidzi:
Not a major deal, but what about parameterizing this?

kelvk:
If you mean toggling this on/off, I think there is limited value because one still needs to define a script in Zabbix to make use of this. For the path, `/usr/lib/zabbix/alertscripts` is provided by the `zabbix-server-mysql` so it seems like a fine place for scripts to be stored in.

kamidzi:
sounds good to me.

PR #746, opened 2015-07-29
Glance API version 2 upgrade

erhudy:
This PR supersedes #742 and should be tested in the same fashion.

* update service catalog to v2 URL
* update Cinder to use v2
* no change to Nova, which apparently does not care or even have an explicit way to set the API version
* does NOT update Horizon to use v2, because Horizon does not support Glance API v2 (see https://blueprints.launchpad.net/horizon/+spec/horizon-glance-v2)

**Test strategy:**

* create volumes from images using `cinder create --image-id` and via Horizon
* launch instances using **Boot from image (creates a new volume)**
* verify that RAW image to volume clone operations are happening entirely in Ceph and are not going via the API (QCOW2 image to volume clones will still need to go through the API and be converted, and `cinder create` from a QCOW2 image may result in something unusable)
  * if things are working you will see basically no activity on the node where the Cinder API is running
  * if things are not working you will see high `glance-api` and `haproxy` CPU usage because the image is being piped through the API
  * check `rbd ls -l volumes-ssd` to verify that volumes created are COW snapshots of the image (the volumes will show the image as their parent if COW)
* do various volume operations (create, delete, resize, etc.)

### DIFF
--- a/cookbooks/bcpc/recipes/glance.rb
+++ b/cookbooks/bcpc/recipes/glance.rb
@@ -182,7 +182,7 @@ bash "glance-cirros-image" do
     code <<-EOH
         . /root/adminrc
         qemu-img convert -f qcow2 -O raw /tmp/cirros-0.3.4-x86_64-disk.img /tmp/cirros-0.3.4-x86_64-disk.raw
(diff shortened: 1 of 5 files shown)


erhudy:
Obviously be sure to test with a sufficiently large image, which you can generate with `qemu-img` and the Ubuntu cloud image to get something like a 20GB RAW image.

caiush:
pulling this down and testing now.

PR #748, opened 2015-07-30
Randomize host selection from scheduler

caiush:
It drives me potty that if you keep starting and stopping an instance it always lands on the same host. This PR randomizes the selection a bit.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -529,6 +529,8 @@ default['bcpc']['keystone']['policy'] = {
 default['bcpc']['nova']['ram_allocation_ratio'] = 1.0
 default['bcpc']['nova']['reserved_host_memory_mb'] = 1024
 default['bcpc']['nova']['cpu_allocation_ratio'] = 2.0
+# select from between this many equally optimal hosts when launching an instance
+default['bcpc']['nova']['scheduler_host_subset_size'] = 3
 # "workers" parameters in nova are set to number of CPUs
 # available by default. This provides an override.
 default['bcpc']['nova']['workers'] = 5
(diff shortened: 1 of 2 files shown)


erhudy:
What's the right way to test this? I launched some instances and ran `nova stop` and `nova start` on them repeatedly, but they always restarted on the same hypervisor each time.

erhudy:
(This was with multiple available hypervisors as far as I remember, so I don't _think_ I was doing anything stupid like trying to test it alongside the ephemeral stuff on a local build, where there would only be 1 general compute and 1 ephemeral compute hypervisor. I will double-check, though.)

kelvk:
I have the same experience as erhudy. I made sure I added bcpc-vm1 and bcpc-vm3 to `general_compute` aggregate group and saw the m1.tiny instance come up on bcpc-vm1. Despite repeated nova stop/start and removal of bcpc-vm1 from the aggregate group, the VM still comes up on the same hypervisor.

caiush:
It randomized the output of the scheduler. Thus the instance has to pass through the scheduler for it to work. I caused confusion by sayin start/stop, should have said start terminate. It's a real issue in clusters that have a low level of churn. When a customer hits a node with a problem, terminates his instance and tries to rebuild and gets me exactly the some result. This would at least give them a shot of being able to schedule

erhudy:
After doing a full delete-terminate, it is working as expected. With this setting set to 3, over 100 loops of launching an instance and terminating it, I got a roughly equal distribution between 3 hypervisors (actually 33/39/28, but this would've probably evened out with more launches). With the setting commented out (i.e., set to 1), the instance kept launching on the same hypervisor.

I am going to amend this commit to be parameterized with a default of 3 and then merge it.

erhudy:
Recheffed with the change to parameterized and it looks good, merging.

PR #751, opened 2015-07-30
Remove building from Zabbix source

kelvk:
Building from source is no longer required as Zabbix is deployed from upstream packages.


mihalis68:
I'll test this now

mihalis68:
Worked for me in a non-vagrant build

PR #755, opened 2015-07-31
reconciling fluentd gems between the build_bins scripts

erhudy:
This should fix an issue reported by @mihalis68 where some gems were not made available to the non-Vagrant build path.


mihalis68:
This doesn't work because in the non-vagrant path it rise to install from a file called (for example) "bins/fluentd-plugin-elasticsearch.gem" - the old method renamed that plugin to drop the version number from the filename.

erhudy:
In the fluentd recipe there's a gate on the attribute `use_bootstrap_v2` to account for the fact that this wasn't updated. Since these are now reconciled, I'll take that out and push an update.

mihalis68:
I tested this on a non-vagrant build and it worked.

PR #744, opened 2015-07-29
Run Zabbix API requests against local instance

kelvk:
`zabbix_config` and `zabbix-api-auto-discovery` in `zabbix-server.rb` sometimes fail to complete successfully on first monitoring node's Chef-run due to delayed apache and haproxy instantiation of zabbix site. This tends to have noticeable impact if you are only bootstrapping one monitoring node.

I have tested this on a fresh build (one monitoring node) and verified that the following are true:
1. `BCPC-Headnode`, `BCPC-Worknode` and `BCPC-Monitoring` templates exist in https://10.0.100.6/zabbix/templates.php.
2. Discovery rule `10.0.100.0/24` is in https://10.0.100.6/zabbix/discoveryconf.php.
3. bcpc-vm* VMs are associated with appropriate templates in https://10.0.100.6/zabbix/hosts.php.

PR #752, opened 2015-07-30
Apt-mirror cleanup

kelvk:
Apt-mirror doesn't currently work when run behind a proxy. This also removes repos (cassandra/hadoop) that are not known to be used and adds Zabbix repo, since Zabbix is now installed from packages.

Test steps:
1. If bootstrap VM needs a proxy, define `proxy` key in https://github.com/bloomberg/bcpc/blob/master/environments/Test-Laptop-Vagrant.json#L52 using `knife environment edit ...`. Value should be a full URL (i.e. `http://<proxy fqdn>:<proxy port>`)
2. Associate apt-mirror and apache-mirror recipes with Bootstrap role and rechef.
3. Execute `apt-mirror`. To speed things up, you may wish to remove ubuntu/ceph/percona/etc `deb` definitions from `/etc/apt/mirror.list`, leaving just the ones of interest (elasticsearch/treasure-data/zabbix).
4. Check downloaded debs are in `/var/spool/apt-mirror/mirror`.
5. Check apt repos are available:
http://10.0.100.3/zabbix/
http://10.0.100.3/elasticsearch/
http://10.0.100.3/fluentd/

PR #759, opened 2015-08-05
Upgrade Zabbix to 2.4

kelvk:
A motivation for upgrading Zabbix now is the new [ability to remove missing resources](https://www.zabbix.com/documentation/2.4/manual/introduction/whatsnew240#option_to_remove_missing_resources_when_importing) when importing Zabbix configuration. As monitoring checks are expected to continually evolve, this provides additional convenience. #728 is an example. For more comprehensive list of changes, see https://www.zabbix.com/documentation/2.4/manual/introduction/whatsnew240.

This also requires upgrading pyzabbix and actually pull it from upstream, instead of copying into our repo which I mistakenly thought it was our code when I reinstated it from our old branches.

This should be tested with #744 as otherwise, errors (though harmless on multi-node cluster) may be thrown during API requests.

Upgrade procedure on existing monitoring cluster:
* Stop Zabbix server and Apache on all monitoring nodes
<pre>
stop zabbix-server
service apache2 stop
</pre>
* Run chef on each node. On the first node, it will upgrade the zabbix packages _and_ database schema. I began the upgrade on the monitoring VIP holder.

Some verification steps to run:
1. Review https://10.0.100.6/zabbix/tr_status.php for errors. The list should only list Zabbix agent version changes and a Ceph health warning.
2. Check agent and server logs in `/var/log/zabbix` for errors.
3. Fail over Zabbix server instance to other monitoring nodes to ensure they are just as functional. Do this by stopping keepalived on the VIP holder.
4. Check that each Zabbix frontend (http://10.0.100.14:7777, http://10.0.100.15:7777, http://10.0.100.16:7777) loads the login page.


kelvk:
Pushed changes from #744 here as they cannot be directly merged.

cookbooks/bcpc/files/default/zabbix_config
```diff
@@ -15,6 +15,8 @@ groups = ["BCPC-Headnode",
           "BCPC-Monitoring"]
 
 zapi = ZabbixAPI(config["ip"])
+# Disable SSL certificate verification
+zapi.session.verify = False
```
erhudy: To remain consistent with the way Roman has been doing the rest of the SSL certificate stuff, you might want to consider gating the certificate verification and also whether the `ZabbixAPI` class is instantiated via IP or FQDN on whether `node['bcpc']['ssl_certificate']`is set (which implies that a real SSL certificate is being provided and that it is okay to fully verify the CA chain).

I don't think fixing this is critical, but it might be good to push a policy of full SSL verification. In fact, I think local builds are now including the self-signed SSL cert in the CA stores of the VMs, so certificate verification should even work there.

kelvk: +1 full SSL verification everywhere. Opened #763 to track this separately.

erhudy: Sounds good, I will go ahead and test applying this over an existing local cluster.

erhudy:
Error messages seen in `/var/log/zabbix/zabbix_server.log` after upgrade:
```
 12963:20150805:205443.770 error reason for "bcpc-vm3:ceph.rgw.haproxy" changed: Unsupported item key.
 12963:20150805:205443.770 error reason for "bcpc-vm3:ceph.rgw.http500" changed: Unsupported item key.
```

erhudy:
I checked the same items on the other work node (bcpc-vm2) and they were enabled. It only appears to have a problem with the checks on bcpc-vm3 for whatever reaso.

erhudy:
Mystery solved: bcpc-vm3 is an ephemeral work node by default and does not have RADOSgw running.

erhudy:
Everything else about upgrade was uneventful and seemed fine to me. I think this is fine to merge with a commit crush.

kelvk:
Thanks! I think the commits are rather distinct but let me know if I should squash them anyway.

erhudy:
Fair point. Do you want to update the apt-mirror template and the example apt-mirror configuration in `docs/` on this PR?

kelvk:
I will update #752.

kelvk:
Bootstrapping a new cluster does not create auto-registration actions correctly (HostMetadata conditional is omitted). I've pushed an updated `zabbix_api_auto_discovery.erb` for Zabbix 2.4 API compatibility.

erhudy:
Re-testing with a fresh build.

erhudy:
Everything seems fine on a fresh build, no errors or fussing that I see.

PR #760, opened 2015-08-05
Parameterize worker counts for Glance, Cinder and Heat

erhudy:
The worker counts for Glance/Cinder were fixed at 1, and Heat was creating as many workers as CPU cores times 3. This parameterizes all worker counts (and also adjusts the default MySQL connection count calculation upwards a bit, because MySQL connections are cheap).

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -979,6 +979,7 @@ default['bcpc']['nova']['policy'] = {
 #  Cinder Settings
 #
 ###########################################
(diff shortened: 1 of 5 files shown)


erhudy:
This should be merged before cutting a 5.5.0 release. I have already tested this both locally and on lab hardware with no ill effects (and in fact several positive effects, because I was running into intermittent errors due to overloading workers and running out of database connections that are no longer occurring).

kelvk:
LGTM. I see additional workers spawned and MySQL `max_connections` increased after recheffing.

PR #761, opened 2015-08-05
Rework the landing page in several ways

erhudy:
**Bugfixes**:
- fixes Glance API URL to actually be v2 in the href
- fixes API FQDN URLs to have `openstack.` at the beginning

**Feature additions**:
- print FQDN at top of page next to IP
- move content into tables
- remove PHPLDAPAdmin listing
- add warning about other API versions not being supported
- print SSL intermediate CA cert if there is one


erhudy:
This should be merged prior to a 5.5.0 cut. I have tested both locally and in the lab with real DNS.

semenovroman:
Currently rebuilding\testing cluster with this branch included.

semenovroman:
Looks great!

PR #762, opened 2015-08-05
Allow tenants to modify instance metadata

semenovroman:
Resolves https://github.com/bloomberg/bcpc/issues/753

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -582,6 +582,10 @@ default['bcpc']['nova']['policy'] = {
   "compute:reboot" => "rule:admin_or_owner",
   "compute:delete" => "rule:admin_or_owner",
   "compute:unlock_override" => "rule:admin_api",
+  
+  "compute:get_instance_metadata" => "rule:admin_or_owner",
+  "compute:update_instance_metadata" => "rule:admin_or_owner",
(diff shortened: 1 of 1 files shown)


erhudy:
What's the right way to test this? Without this patch applied, I'm able to launch instances as a regular non-admin user in a new tenancy and pass in metadata via `nova boot --meta`, as well as user data via `--user-data`.

jerenkrantz:
I think that Fog changes the metadata *after* instance creation.

erhudy:
I see. `nova meta` does get angry in expected ways, then:
```
$ nova meta test set meep=bleep
ERROR (Forbidden): Policy doesn't allow compute:update_instance_metadata to be performed. (HTTP 403)
$ nova meta test delete meep=bleep
ERROR (Forbidden): Policy doesn't allow compute:get_instance_metadata to be performed. (HTTP 403)
```

semenovroman:
@erhudy  After this PR is applied you should be able to set\modify existing instance metadata without any errors. `nova meta test set meep=bleep` is expected to work fine with this patch.

erhudy:
After patch works as expected:
```
$ nova meta test set meep=bleep
$ nova meta test delete meep=bleep
$ nova meta test delete moop=bloop
ERROR (NotFound): Metadata item was not found (HTTP 404)
```

PR #765, opened 2015-08-06
Fixes bug where Ceph keys in data bag could beome desynced

erhudy:
This fixes a bug where, if you destroyed and re-Cheffed a cluster without clearing the Cinder/Glance keys out of the data bag, those keys wouldn't get updated from Ceph auth when new keys were generated and nova-compute wouldn't be able to talk to Ceph. The data bag key is now updated when the new value is generated, and the solution has been generalized to adding a `force` keyword on `make_config()` to allow overwriting values.

(Technically only the Cinder value from the data bag is used, Glance one is not, but this fixes both for the sake of consistency.)

### DIFF
--- a/cookbooks/bcpc/libraries/utils.rb
+++ b/cookbooks/bcpc/libraries/utils.rb
@@ -48,9 +48,9 @@ def init_config
     end
 end
 
-def make_config(key, value)
+def make_config(key, value, force=false)
     init_config if $dbi.nil?
-    if $dbi[key].nil?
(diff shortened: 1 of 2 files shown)


erhudy:
I tested this on an existing cluster and also on a new one. On the existing cluster, I tested by removing the keys completely and recheffing, and also modifying the keys and then recheffing. In both cases Chef replaced the missing/wrong keys with the right ones.

PR #747, opened 2015-07-30
Added support for AZs in compute

caiush:
Currently users can use server groups to make sure they don't land on the same host. If however, you have a bigger failure domain, e.g. racks, there is no way to allow users to spread VMs over these. 

This PR introduces availability zones (AZ). By default the AZ will be one per rack, however, you can create your own AZ by setting `node['bcpc']['availability_zone']`.

In the laptop environment we give each node its own AZ, just for fun.

PR #768, opened 2015-08-08
Availability zones

caiush:
I was unable to repo @erhudy's issue in #747 but I suspect its the cinder zone setting that was wrong and must have got mangled in the commit. 

Anyhow, this PR has been rebased against master and allows the specification of AZs. It would prob be a good plan to call 1+ racks an AZ (depending on the size of your cloud).

To test I have to change bcpc-vm3's role back to a non ephemeral worknode. I did also chef this in over a running 5.5 cluster (with running instances) with success.

### DIFF
--- a/cookbooks/bcpc/recipes/host-aggregates.rb
+++ b/cookbooks/bcpc/recipes/host-aggregates.rb
@@ -16,11 +16,8 @@
 # limitations under the License.
 #
 
-az_number = if node['hostname'] =~ /bcpc\-vm/
-  node['bcpc']['node_number']
-else
-  node['bcpc']['rack'].nil? ? 1 : node['bcpc']['rack']
-end 
+parsed_rack_number = node['bcpc']['rack_name'].match(/^rack-(\d+)/)
+az_number = parsed_rack_number.nil? ? 1 : parsed_rack_number.captures[0].to_i
 availability_zone = (node['bcpc']['availability_zone'].nil? ) ? node['bcpc']['region_name'] + "-" + az_number.to_s : node['bcpc']['availability_zone'].to_s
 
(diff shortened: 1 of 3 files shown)


erhudy:
I am testing on a slightly hacked up 6-node cluster where the last 3 nodes are work nodes and not monitoring nodes. Each node is its own AZ.

cookbooks/bcpc/recipes/host-aggregates.rb
```diff
@@ -16,6 +16,10 @@
 # limitations under the License.
 #
 
+parsed_rack_number = node['bcpc']['rack_name'].match(/^rack-(\d+)/)
+az_number = parsed_rack_number.nil? ? 1 : parsed_rack_number.captures[0].to_i
```
erhudy: I rewrote the calculation of the AZ number here to derive it from `['bcpc']['rack_name']` instead of `['bcpc']['rack']`, which doesn't exist. This presupposes that the format of the attribute will be `rack-###`, so if that changes this will "break" (it will fail in a non-ridiculous way and just set everything to live in a single availability zone rather than exploding in a Bad Way).

erhudy:
Tested under lab conditions with real hardware, putting one node in a separate "rack". Things seem to work as expected - if you pick an AZ all the instances launch in it, assuming the appropriate resources are there. Cinder and Glance seem fine. Turning over to @caiush for some experimentation with server groups.

erhudy:
Tested AZ interaction with server groups and it seemed fine. Going to re-test locally with my changed code.

erhudy:
Seems fine, a single AZ is created called Test-Laptop-Vagrant-1 (or whatever your environment is named). Instance launches work as expected.

PR #769, opened 2015-08-10
Add fallback firewall rule for out-of-cluster Zabbix-agent communication

kelvk:
This creates a rule to permit inbound Zabbix-agent connectivity. This is necessary for nodes monitored by a VIP that lives outside of its management subnet.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -160,7 +160,6 @@ default['bcpc']['management']['netmask'] = "255.255.255.0"
 default['bcpc']['management']['cidr'] = "10.17.1.0/24"
 default['bcpc']['management']['gateway'] = "10.17.1.1"
 default['bcpc']['management']['interface'] = nil
-default['bcpc']['management']['monitoring']['vip'] = "10.17.1.16"
 # if 'interface' is a VLAN interface, specifying a parent allows MTUs
 # to be set properly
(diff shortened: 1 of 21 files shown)


erhudy:
The expectation is that the IP address of the external Zabbix would be in `["bcpc"]["management"]["monitoring"]["vip"]`?

kelvk:
It is. I think I will change the attribute `['bcpc']['monitoring']['vip']` since external cluster monitoring will likely be the primary/only use case in future. Hold off review until I push a new update.

kelvk:
Commit updated.

erhudy:
One suggestion: if `["bcpc"]["management"]["monitoring"]["vip"]` is set and ["bcpc"]["monitoring"]["vip"]` is not, intentionally fail the Chef run so that the user becomes aware of the need to change the environment file.

erhudy:
Rebuilt and it seems to work as expected - new firewall rules are in place on the work nodes and monitoring nodes, Zabbix checks are working.

erhudy:
Retesting with additional change and will merge if all is well.

kelvk:
Hmm, in-line comments don't seem to generate email notifications. Anyway, I've updated commit with fix.

erhudy:
It caught the error this time. Going to test cheffing on an existing cluster and will merge if good.

erhudy:
Tested on existing cluster with success.

PR #770, opened 2015-08-13
Changes Vagrant behavior so that all nodes but bootstrap get extra disks

erhudy:
This is a small behavioral change that has Vagrant create the 4 extra disks for all non-bootstrap nodes rather than just the first 3. Since they're thin provisioned, the extra space take is like 3MB, and it allows for easily re-using the monitoring nodes as Ceph/ephemeral compute nodes.

### DIFF
--- a/bootstrap/vagrant_scripts/Vagrantfile
+++ b/bootstrap/vagrant_scripts/Vagrantfile
@@ -269,9 +269,7 @@ Vagrant.configure("2") do |config|
         vb.customize ["modifyvm", :id, "--hwvirtex", "on"]
         vb.customize ["modifyvm", :id, "--ioapic", "on"]
 
-        # this attaches extra disks to the first 3 non-bootstrap nodes (additional cluster
-        # nodes are assumed to be monitoring, which do not need extra disks)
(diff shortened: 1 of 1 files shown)


mihalis68:
Nice!

Sent from Rivendell

----- Original Message -----
From: Edmund Rhudy <reply@reply.github.com>
To: chef-bcpc@noreply.github.com
At: 13-Aug-2015 13:08:00


This is a small behavioral change that has Vagrant create the 4 extra disks for all non-bootstrap nodes rather than just the first 3. Since they're thin provisioned, the extra space take is like 3MB, and it allows for easily re-using the monitoring nodes as Ceph/ephemeral compute nodes.You can view, comment on, or merge this pull request online at:
  https://github.com/bloomberg/chef-bcpc/pull/770Commit Summary
*changes Vagrant behavior so that all nodes but bootstrap get extra disksFile Changes
*M     bootstrap/vagrant_scripts/Vagrantfile     (4)   Patch Links:
*https://github.com/bloomberg/chef-bcpc/pull/770.patch
*https://github.com/bloomberg/chef-bcpc/pull/770.diff

PR #771, opened 2015-08-13
Switch catalog backend to SQL

semenovroman:
Plan is to write a chef service\endpoint provider in the next iteration and modify this code to use it (will get rid off 'jq')

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -306,7 +306,7 @@ default['bcpc']['keystone']['wait_for_keystone_timeout'] = 120
 # The driver section below allows either 'sql' or 'ldap' (or 'templated' for catalog)
 # Note that not all drivers may support SQL/LDAP, only tinker if you know what you're getting into
 default['bcpc']['keystone']['drivers']['assignment'] = 'sql'
-default['bcpc']['keystone']['drivers']['catalog'] = 'templated'
+default['bcpc']['keystone']['drivers']['catalog'] = 'sql'
 default['bcpc']['keystone']['drivers']['credential'] = 'sql'
 default['bcpc']['keystone']['drivers']['domain_config'] = 'sql'
 default['bcpc']['keystone']['drivers']['endpoint_filter'] = 'sql'
--- a/cookbooks/bcpc/recipes/keystone.rb
+++ b/cookbooks/bcpc/recipes/keystone.rb
@@ -67,7 +67,7 @@ template "/etc/keystone/keystone.conf" do
     group "keystone"
     mode 00600
     variables(:servers => get_head_nodes)
-    notifies :restart, "service[apache2]", :delayed
+    notifies :restart, "service[apache2]", :immediately
 end
 
 template "/etc/keystone/default_catalog.templates" do
--- a/cookbooks/bcpc/recipes/keystone.rb
+++ b/cookbooks/bcpc/recipes/keystone.rb
@@ -43,6 +43,11 @@ package 'keystone' do
     action :upgrade
 end
 
+# needed to parse openstack json output
+package 'jq' do
+    action :upgrade
+end
+
 # do not run or try to start standalone keystone service since it is now served by WSGI
 service "keystone" do
     action [:disable, :stop]
@@ -190,7 +195,7 @@ bash "keystone-create-admin-tenant" do
     code <<-EOH
         . /root/adminrc
         . /root/keystonerc
-        keystone tenant-create --name "#{node['bcpc']['admin_tenant']}" --description "Admin services" 
+        keystone tenant-create --name "#{node['bcpc']['admin_tenant']}" --description "Admin services"
     EOH
     not_if ". /root/keystonerc; . /root/adminrc; keystone tenant-get '#{node['bcpc']['admin_tenant']}'"
 end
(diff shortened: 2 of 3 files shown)


kamidzi:
Is this intended to be applied to an existing cluster? I ran into the following problem on a freshly build, already operational cluster:
```
Error executing action `run` on resource 'bash[keystone-create-identity-endpoint]'

Expected process to exit with [0], but received '1'
---- Begin output of "bash"  "/tmp/chef-script20150817-17219-6blsu2" ----
STDOUT:
STDERR: ERROR: openstack No service with a type, name or ID of 'identity' exists.
---- End output of "bash"  "/tmp/chef-script20150817-17219-6blsu2" ----
Ran "bash"  "/tmp/chef-script20150817-17219-6blsu2" returned 1
```

semenovroman:
@kamidzi this should work on an existing cluster, as well as on a new build. What does `keystone service-list` give you on a headnode (looks like it is trying to create keystone endpoint, but keystone service is not defined)?

kamidzi:
When I had attempted this, a second chef-run seemed to complete cleanly. For `keystone service-list`, do you mean before or after this change is applied? Either way, I'm waiting on a clean build to try it again.

semenovroman:
Error you have seen might be related to a bug I just found when doing a brand new build. Looking into it right now.

semenovroman:
Updated code to work cleanly on a new cluster deploy (force-pushed).

semenovroman:
@kamidzi and I found an issue, that affects this PR, with keystone WSGI returning different results for subsequent runs (problem goes away if `processes=5` in `/etc/apache2/sites-enabled/wsgi-keystone.conf` is set to `1`)

erhudy:
Where would these different results manifest? Running `keystone catalog` and `keystone endpoint-list` repeatedly seems to consistently return the same thing each time.

semenovroman:
@erhudy In most cases (installations) it works just fine. However, sometimes, running `keystone` commands will give different results on subsequent executions (setting `processes=1` fixes this behavior). It is unknown at the moment why this behavior appears on some installations but not others.

erhudy:
I see. My selection of `processes=5` was an arbitrary pick when I moved Keystone from eventlet to WSGI, but we probably need to have multiple Keystone somethings available for servicing requests, whether it's processes or threads, since every API call requires Keystone. Does the problem surface with something like `threads=10 processes=1`?

semenovroman:
@kamidzi do you still have that build where `keystone ...` was returning different results on each run? If you do, can we try setting `threads=10 processes=1` as advised by @erhudy?

kamidzi:
Unfortunately, no. Though I can tell you that it was built from 7a685a6 with 6 vms total, 1 headnode.

erhudy:
Processes vs threads in WSGI is like prefork vs worker for Apache backends. I find it highly weird that you would see broken results when we were using multiple single-threaded processes, where everything should be isolated. I have no expectation that switching to threads instead of processes will magically fix the problem (in fact I expect the opposite), but it's worth a shot just to see if something different happens.

semenovroman:
I will try building 6vm cluster and see if I can reproduce keystone issue

kamidzi:
@semenovroman and I sat down together and confirmed this does not chef in cleanly.

semenovroman:
Issue that his PR surfaced was with keystone caching. I am currently finishing testing a fix. Will update PR once I confirm solution works in all possible scenarios.

kamidzi:
This is what I'm consistently experiencing: 
```
2015-08-26 14:49:41.893 13637 TRACE keystone.common.wsgi DBReferenceError: (IntegrityError) (1452, 'Cannot add or update a child row: a foreign key constraint fails (`keystone`.`endpoint`, CONSTRAINT `fk_endpoint_region_id` FOREIGN KEY (`region_id`) REFERENCES `region` (`id`))') 'INSERT INTO endpoint (id, legacy_endpoint_id, interface, region_id, service_id, url, enabled, extra) VALUES (%s, %s, %s, %s, %s, %s, %s, %s)' ('b58911bd6a3847689480d1561dba9407', 'd26a74eaf9fb43d7b7c918e7b43c13cc', 'admin', 'Test-Laptop-Vagrant', '84f6bedad9444e9c9fc76a2424c16244', 'https://openstack.bcpc.example.com:35357/v2.0', 1, '{}')
```

but looking a bit, seems like a red herring. The real error is indeed the lack of a service catalog. The following non-fatal error precedes the above:

```
2015-08-26 16:37:07.273 13646 WARNING keystone.common.wsgi [-] Could not find service: identity
```

kamidzi:
In my testing this will only work if keystone catalog caching is off. I think it makes sense to update the default environment as well with ``bcpc.keystone.enable_caching = false``. I would do this myself, but this branch is not public. Otherwise, LGTM.

kamidzi:
Actually, it seems like that attribute's effect isn't localized to catalog caching...it may be better to only turn the keystone catalog cache off.

kamidzi:
@semenovroman, I believe I have a--perhaps unideal--solution. Let me know if this sits with you: 66cd3905a814a8c2369b1d49b247d477bda3faa1

You test the migration by
1.  Build master
2.  Upload the updated cookbook with these changes
3.  Run chef-client ( `bcpc::keystone` ) should be sufficient

Then to to go back
1.  Clear `keystone.service`, `keystone.endpoint` tables
2.  Upload the original cookbook from master
3.  Run chef-client ( `bcpc::keystone` ) should be sufficient

Only caveat is that persons will now need to remember to pass the `@catalog_caching` variable to the `template[/etc/keystone/keystone.conf]` (This is already similar case with `@servers`).

kamidzi:
This may require an additional wait on glance before `bash[glance-cirros-image]`...

semenovroman:
Latest patch is working fine with caching enabled for me. Lets try going over `re chefing` tomorrow together and try to figure out what exactly is going wrong.

kamidzi:
Seems like the phenomena I observed was isolate to my testing environment.

PR #773, opened 2015-08-14
Breaking up the monolithic BCPC cookbook

erhudy:
I am taking on the task of breaking up the BCPC cookbook into smaller more manageable cookbooks, with the following goals:
* switching from roles to role cookbooks for versioning purposes
* isolating concerns within cookbooks (I've already found a number of areas where recipes were executing other seemingly unrelated recipes, like networking interface setup installing and configuring Apache)
* migrating all binary files into a **bcpc-binary-files** cookbook
* improving composability of the cookbooks
* improving Chef cookbook compile and convergence time

This PR is not intended to be merged for some time, but I'm opening it to track progress on the work done.


kamidzi:
So the cookbook versions intend to be synchronized with the BCPC release, or...? What's the significance of starting at 6.0.0?

erhudy:
I have not considered the cookbook versioning strategy with any particular level of detail. My expectation is that the work here would eventually form the basis for a release 6, hence picking 6.0.0 to start (and the branch name), but it shouldn't be taken as an indication that I think it's the final form. Everything in this branch is definitely in high flux right now.

erhudy:
Closing this to re-open with a copy of the branch on this repo.

PR #774, opened 2015-08-14
Power DNS stats page

caiush:
Enable the power DNS web page for interesting stats,

This would chef into to any recent cluster just fine.


kelvk:
This chefs in fine. Few things noted:
1. The web auth accepts any username
2. Inbound port 8081 is not permitted in iptables, so access will be restricted to within the cluster. Traffic will also be in plain-text. However, as I understand it, the page consumer will live inside the cluster, therefore these are acceptable.

I will withhold merge until Wednesday should there be any further comments.

caiush:
yeah its internal only at this point. Not sure there is anything really super sensitive on this page anyhow, so if we want to open it up we can consider it.

PR #775, opened 2015-08-18
Deploy (and upgrade) Kibana using apt

kelvk:
Tested with existing and fresh build. https://10.0.100.6/kibana/#/settings/about should indicate 4.1.1 when built.


mihalis68:
I've been looking at this. It provides a welcome amount of tidying. Are there also functional changes when switching from our locally packaged version? I'm trying to see if we would want to merge this into existing BCPC release 5.X clusters.

kelvk:
Yes, though I don't expect them to affect or change our usage today.

https://www.elastic.co/guide/en/kibana/4.1/whats-new.html

PR #776, opened 2015-08-18
Adds remaining ephemeral flavors up to e1.2xlarge

erhudy:
40GB max root disk size is intentional.


mihalis68:
This sets some attributes on non-ephemeral flavors also. Is that deliberate?

erhudy:
It doesn't do anything new there, I just shuffled the list around so that non-ephemeral flavors came first. The diff is a bit misleading. The extra_specs stuff is the same as before.

mihalis68:
Ah ok, I see. In that case, LGTM

PR #778, opened 2015-08-20
Remove gratuitous patches. Fixes #777

### LINKED ISSUE #777: Patches do not apply cleanly on 2015.1.1
A patch release for Kilo (2015.1.1) has been released to kilo-proposed on the Ubuntu cloud archive that causes builds to fail due to patches no longer applying cleanly.

kamidzi:
Removes patches that are breaking the build. It's unclear whether `horizon-swift-content-length.patch` is also still required.


kamidzi:
See #779.

PR #779, opened 2015-08-20
Remove gratuitous patches. Fixes #777

### LINKED ISSUE #777: Patches do not apply cleanly on 2015.1.1

kamidzi:
Removes patches that are breaking the build. It's unclear whether horizon-swift-content-length.patch is also still required.


erhudy:
I do not think the patches should be removed entirely, because they will be needed for installations using 2015.1.0 still. They should consider the SHA of the file to be patched to determine whether the patch is needed.

kamidzi:
Since when have we supported older versions? It's always been upgrade or die.

erhudy:
Current production clusters are deployed on 2015.1.0, and we will need to support that version as well as 2015.1.1 until upgrades can be scheduled and executed.

erhudy:
Superseding this pull request with #783, which selectively applies the patches in question for 2015.1.0 and skips for 2015.1.1.

PR #780, opened 2015-08-20
Make mon_pg_warn_max_per_osd configurable (300 is upstream default)

erhudy:
Adding this in to make it possible to suppress the Ceph "too many PGs per OSD" warning, which sets the Ceph cluster status at HEALTH_WARN.


mihalis68:
Looks good to me. I verified that it is indeed a threshold number, not a boolean (other blogs report success with setting it to either a very big or to 0 to suppress the warning).

PR #781, opened 2015-08-20
Parameterize Openstack logging verbosity option

kelvk:
Log level remains unchanged (WARN by [default](http://docs.openstack.org/kilo/config-reference/content/list-of-compute-config-options.html)). Log level post-change can be verified by inspecting `/var/log/{nova,glance,cinder}/*.log`.


erhudy:
Tested in lab and all was well, toggled true/false as expected when set in environment.

PR #783, opened 2015-08-21
Guards for conflicting patch files to only patch 2015.1.0 versions of files

erhudy:
Rewrite of #779 so that these 3 patches will continue to apply to 2015.1.0 installations, but will not apply for 2015.1.1.

### DIFF
--- a/cookbooks/bcpc/recipes/glance.rb
+++ b/cookbooks/bcpc/recipes/glance.rb
@@ -60,7 +60,7 @@ bash "patch-backport-for-glance-v2-null-description" do
       exit $rv
     fi
   EOH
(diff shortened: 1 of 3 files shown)


erhudy:
I have tested this against 2015.1.1 upstream, and a local mirror copy of the OpenStack packages taken before the release of 2015.1.1. Also tested the functionality in 2015.1.1 that was affected by these 3 packages and verified that it remained operational:
* Inspecting images via Glance API v2 with possible nulls in metadata
* Nova DHCP server gateway address behavior
* Glance image uploads via Horizon

kamidzi:
I think it makes sense to anchor those regex's

erhudy:
Redoing this PR off the release branch.

PR #785, opened 2015-08-21
Checking for SHAs of 2015.1.0 files to decide whether to patch

erhudy:
This PR supersedes #783.


kamidzi:
Waiting on build to complete, then we should be good to release

Issue #786, opened 2015-08-21
PowerDNS record insertion into MySQL can stall


erhudy:
On larger record insertions, MySQL can stall out and get stuck on the resource forever. I believe this only affects the `powerdns-load-fixed-records` resource.

PR #789, opened 2015-08-22
Workaround for #786 - Only reload PowerDNS data into MySQL if it changes

erhudy:
This doesn't fix the underlying issue, but should at least decrease the frequency of its occurrence by only reloading the data into MySQL when necessary (i.e., when the template contents change).


mihalis68:
Looks good to me.

kelvk:
Works as expected.

mihalis68:
Thanks! :)

reply@reply.github.com At: Aug 28 2015 10:48:01" data-digest="From: reply@reply.github.com At: Aug 28 2015 10:48:01" style="">

PR #795, opened 2015-08-25
Fixes #772 - Relocates RADOSgw log for logrotate

### LINKED ISSUE #772: Ceph log rotation breaks radosgw logging
There are two log rotation scripts, `/etc/logrotate.d/ceph` and `/etc/logrotate.d/radosgw`, but only the former comes into effect because radosgw for unknown reason, does not log to `/var/log/radosgw`. When ceph log rotation runs, it greedily rotates `/var/log/ceph/radosgw.log` as well. Since this is unintended, the radosgw daemon never received a HUP signal to release the log.

When log is broken:
<pre>
radosgw 47936 root    3w   REG                8,5  4558962 15212973 /var/log/ceph/radosgw.log.1 (deleted)
</pre>

After HUP-ing it:
<pre>
radosgw 47936 root    3w   REG                8,5     2688 15213286 /var/log/ceph/radosgw.log
</pre>

This is particularly an issue because the log file is monitored by Zabbix.

erhudy:
This fixes #772 by relocating the RADOSgw log to `/var/log/radosgw`, where `/etc/logrotate.d/radosgw` expects it to be, and updates the Zabbix agent template appropriately.

NOTE: if cheffing onto an existing cluster, you will need to roll RADOSgw manually with `sudo service radosgw-all restart`, as changes to `ceph.conf` intentionally do not trigger service restarts due to the large number of services controlled by the configuration file and the potential Extreme Badness of Ceph daemons being restarted haphazardly.

### DIFF
--- a/cookbooks/bcpc/templates/default/ceph.conf.erb
+++ b/cookbooks/bcpc/templates/default/ceph.conf.erb
@@ -65,7 +65,7 @@
 [client.radosgw.gateway]
     host = <%=node['hostname']%>
     keyring = /var/lib/ceph/radosgw/$cluster-$id/keyring
-    log file = /var/log/ceph/radosgw.log
+    log file = /var/log/radosgw/radosgw.log
     rgw data = /var/lib/ceph/radosgw/$cluser-$id
     rgw enable ops log = false
     rgw enable usage log = false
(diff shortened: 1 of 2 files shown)


kelvk:
Tested with fresh (0.94.3) and existing build by running `logrotate -v -f /etc/logrotate.d/radosgw` couple times. Log rotates as expected and new log is written each time. Zabbix check also looks good.

PR #796, opened 2015-08-26
Cache binary build products outside bootstrap VM for faster bootstraps

erhudy:
This PR creates a directory on the local host and copies the build product results from running `bootstrap/shared/shared_build_bins.sh` to them at the end of the script so that they can reused in between local builds. In practice, this seems to cut 2-3 minutes off build times.

### DIFF
--- a/bootstrap/shared/shared_build_bins.sh
+++ b/bootstrap/shared/shared_build_bins.sh
@@ -35,7 +35,7 @@ pushd $BUILD_DEST
 # (will not exist if this is the first time run with the new script)
 if [ -d $BUILD_CACHE_DIR ]; then
   echo "Copying cached build products..."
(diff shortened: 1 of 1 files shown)


mihalis68:
Looks good to me

kamidzi:
I would recommend using slightly safer options to both `pwd` and `rsync`, something like `rysnc -axSH` and `pwd -P`. Also as an aside, @kelvk saw some interesting behaviour with file copies via vboxfs earlier. Patches copied in that manner were introducing extra spaces, I believe... @kelvk?

erhudy:
I've never seen anything like that with vboxsf. I will look at amending the script changes to incorporate your suggestions.

erhudy:
Rebased off master and updated pwd/rsync calls.

PR #801, opened 2015-08-27
Fix float association race condition in multi-head cluster

kelvk:
`check nova` runs every 10 minutes on head nodes. On multi-head clusters, float IP association sometimes fail if another test instance has grabbed the IP. Instead of failing the entire check, we should let it wait and retry.

### DIFF
--- a/cookbooks/bcpc/files/default/checks/nova
+++ b/cookbooks/bcpc/files/default/checks/nova
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-
+import syslog
 import socket
 import time
 import novaclient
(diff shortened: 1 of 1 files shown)


erhudy:
Will check this out in the lab. Will anything be logged anywhere to confirm the new behavior is working as expected?

kelvk:
It doesn't. We could add logging when `BadRequest` exception is raised. I can push an update if you like.

erhudy:
I wouldn't mind it, since it makes verification of the expected behavior straightforward.

kelvk:
Updated

erhudy:
Applied down in the lab, will let it bake.

erhudy:
Forgot about this, but nothing flamed out and so I shall merge it.

PR #807, opened 2015-09-02
Finish up rally setup

kamidzi:
This doesn't address the InsecurePlatformWarning, but with this small change, bcpc::rally-setup should work as intended; i.e, you should be able to do the following on a bootstrap node:

```
$ sudo chef-client -obcpc::rally-setup
$ rally use existing
$ rally deployment check
<...etc...>
```


erhudy:
In my opinion, we shouldn't be distributing other cookbooks as part of our repo. I don't like that we have chef-client included now, and I removed it as part of the cookbook refactoring, though I think it was distributed previously because we patched it (maybe we no longer need to include it now that we're using Chef 12 everywhere). Including other cookbooks could open us up to redistribution and licensing issues; in this case it happens to also be Apache-licensed, but it's probably a habit better avoided.

To have the local build process fetch the cookbook in question, you can add another line to `bootstrap/shared/shared_prereqs.sh` below line 51 (the Chef Supermarket URL format is pretty consistent). `shared_configure_chef.sh` will auto-expand it along with everything else in that directory.

kamidzi:
Duly noted. I disagree.

mihalis68:
@kamidzi could you explain a bit more please? I would think our default stance is to NOT copy external code into our project, even when the license does not forbid that. I would welcome comment on that from @jerenkrantz and @kpfleming

kpfleming:
It's definitely preferable to avoid copying dependencies into our own projects. If there's a concern about their availability at build/deploy time, that can be addressed by putting a copy of the dependency into a hosting location that we control, but not inside a project.

kamidzi:
@mihalis68, my concerns are as @kpfleming has suggested about the availability of the cookbooks at build-time as well as being able to easily "chef" in this change on an existing build, so to speak. Regardless, I can add it to `shared_prereqs.sh` as suggested by @erhudy.

mihalis68:
Yes I understand - valid concerns. Changing it, as you say, to make it a pre-requisite sounds good to me.

jerenkrantz:
Just to add to @kpfleming's comments, we can include and distribute any open-source code that we desire as long as we comply with the license terms.  However, unless we are modifying or patching it, it doesn't make sense to put it in our repository - e.g., we just refer to the external cookbook and have Chef download it at build/run-time.  If it is source code, it would be best to just include the patch in our repository rather than a full copy.  Previously, we had a patch to chef-client cookbook as it was broken on Chef 11; but, I'm not aware why we now include the full cookbook.

I would not be aware of any reason why we couldn't depend upon the Chef Supermarket at deploy/run-time - if that is really a reliability issue, we should bang on Chef rather than blindly copying cookbooks where we do not make any modifications.  CC @acaiafa @johnbellone

erhudy:
I should have specified in my original comment that I was raising the concern because of conversations I had had with @kpfleming previously about code redistribution.

johnbellone:
You should not depend on the Chef Supermarket at runtime unless you plan on vendoring all of the cookbooks. This is possible on a workstation with Internet access by running `berks vendor` which puts everything into a tarball.

erhudy:
The current local build procedure retrieves and caches specific versions of the cookbooks from the Chef Supermarket locally, per https://github.com/bloomberg/chef-bcpc/blob/master/bootstrap/shared/shared_prereqs.sh#L49-L56.

kpfleming:
@erhudy: That discussion (about distributing open source software) was primarily focused on binary distribution, which isn't an issue here (thankfully).

erhudy:
After building from this, the following steps got Rally into a working form:
```
vagrant@bcpc-bootstrap:~$ sudo chef-client -obcpc::rally-setup
[...snip...]
vagrant@bcpc-bootstrap:~$ export OS_TENANT_NAME=AdminTenant
vagrant@bcpc-bootstrap:~$ export OS_USERNAME=admin
vagrant@bcpc-bootstrap:~$ export OS_PASSWORD=xxxxxxx
vagrant@bcpc-bootstrap:~$ export OS_AUTH_URL=https://openstack.bcpc.example.com:5000/v2.0/
vagrant@bcpc-bootstrap:~$ rally deployment create --fromenv --name=existing
+--------------------------------------+----------------------------+----------+------------------+--------+
| uuid                                 | created_at                 | name     | status           | active |
+--------------------------------------+----------------------------+----------+------------------+--------+
| 366ba985-f28f-437a-b347-f62944f13013 | 2015-09-17 14:29:39.019629 | existing | deploy->finished |        |
+--------------------------------------+----------------------------+----------+------------------+--------+
Using deployment: 366ba985-f28f-437a-b347-f62944f13013
~/.rally/openrc was updated

HINTS:
* To get your cloud resources, run:
	rally show [flavors|images|keypairs|networks|secgroups]

* To use standard OpenStack clients, set up your env by running:
	source ~/.rally/openrc
  OpenStack clients are now configured, e.g run:
	glance image-list
vagrant@bcpc-bootstrap:~$ rally deployment check
2015-09-17 14:29:50.685 17866 WARNING oslo_config.cfg [-] Option "https_cacert" from group "DEFAULT" is deprecated for removal.  Its value may be silently ignored in the future.
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
keystone endpoints are valid and following services are available:
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
+-----------------+----------+-----------+
| services        | type     | status    |
+-----------------+----------+-----------+
| Compute Service | compute  | Available |
| EC2 Service     | ec2      | Available |
| Image Service   | image    | Available |
| Volume Service  | volume   | Available |
| cinderv2        | volumev2 | Available |
| keystone        | identity | Available |
+-----------------+----------+-----------+
```

erhudy:
Should **bcpc::rally-setup** be added to the bootstrap server run list?

kamidzi:
There's a sequencing issue IIRC, that's why it was previously omitted. The cluster has to up for rally-setup to work, but the bootstrap node comes up before the rest of the cluster. See: https://github.com/bloomberg/chef-bcpc/blob/feature/rally_setup/cookbooks/bcpc/recipes/rally-setup.rb#L21

As for having to execute
```
rally deployment create --fromenv --name=existing
```

that shouldn't be the case. See https://github.com/bloomberg/chef-bcpc/blob/feature/rally_setup/cookbooks/bcpc/recipes/rally-setup.rb#L75

erhudy:
I see. `rally deployment use existing` didn't work for me for whatever reason, but I will rebuild and retest again.

erhudy:
Rested again and got the expected result this time, not sure what went wrong the previous time:
```
vagrant@bcpc-bootstrap:~$ sudo chef-client -obcpc::rally-setup
[...snip...]
vagrant@bcpc-bootstrap:~$ rally deployment use existing
Using deployment: 1a12cf25-8e91-46d2-91b1-01b1b877e209
~/.rally/openrc was updated

HINTS:
* To get your cloud resources, run:
	rally show [flavors|images|keypairs|networks|secgroups]

* To use standard OpenStack clients, set up your env by running:
	source ~/.rally/openrc
  OpenStack clients are now configured, e.g run:
	glance image-list
vagrant@bcpc-bootstrap:~$ rally deployment check
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
keystone endpoints are valid and following services are available:
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
+-----------------+----------+-----------+
| services        | type     | status    |
+-----------------+----------+-----------+
| Compute Service | compute  | Available |
| EC2 Service     | ec2      | Available |
| Image Service   | image    | Available |
| Volume Service  | volume   | Available |
| cinderv2        | volumev2 | Available |
| keystone        | identity | Available |
+-----------------+----------+-----------+
```

Issue #809, opened 2015-09-04
dns_fill.py fails when processing orphaned instances


kamidzi:
In the case where an instance is created, float attached, then the corresponding project is deleted, `dns_fill.py` will error out without processing all the necessary records. The error appears as below:

```
Traceback (most recent call last):
  File "/usr/local/bin/dns_fill.py", line 129, in <module>
    args.func(args)
  File "/usr/local/bin/dns_fill.py", line 107, in c_run
    nova_rows = dnsp.generate_records_from_vms()
  File "/usr/local/bin/dns_fill.py", line 61, in generate_records_from_vms
    tenant = self.keystone.tenants.get(tid)
  File "/usr/lib/python2.7/dist-packages/keystoneclient/v2_0/tenants.py", line 83, in get
    return self._get("/tenants/%s" % tenant_id, "tenant")
  File "/usr/lib/python2.7/dist-packages/keystoneclient/base.py", line 136, in _get
    resp, body = self.client.get(url, **kwargs)
  File "/usr/lib/python2.7/dist-packages/keystoneclient/adapter.py", line 164, in get
    return self.request(url, 'GET', **kwargs)
  File "/usr/lib/python2.7/dist-packages/keystoneclient/adapter.py", line 200, in request
    resp = super(LegacyJsonAdapter, self).request(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/keystoneclient/adapter.py", line 89, in request
    return self.session.request(url, method, **kwargs)
  File "/usr/lib/python2.7/dist-packages/keystoneclient/utils.py", line 318, in inner
    return func(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/keystoneclient/session.py", line 389, in request
    raise exceptions.from_response(resp, method, url)
keystoneclient.openstack.common.apiclient.exceptions.NotFound: Could not find project: 771dc271db654194bb0b7e220f07a2e8 (HTTP 404) (Request-ID: req-5a4a08e1-daee-4016-b34a-0245ab26d071)
```

kamidzi:
btw, original issue reported by @kelvk on another system.

PR #810, opened 2015-09-04
Add check for floats attached to instances in non-existent tenants. F…

kamidzi:
…ixes #809

To reproduce:

1. Create a new project
2. Launch an instance in that project
3. Attach a float to said instance
4. Run `dns_fill.py`

### DIFF
--- a/cookbooks/bcpc/files/default/dns_fill.py
+++ b/cookbooks/bcpc/files/default/dns_fill.py
@@ -10,8 +10,9 @@ then creates a CNAME record to point to the public-X.X.X.X username
 
 import novaclient
 import keystoneclient
-from novaclient.v1_1 import client
+from novaclient.v2 import client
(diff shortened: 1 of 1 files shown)


kamidzi:
Btw, this is really a symptom of a--still unresolved--larger issues that appear to be have been under discussion for quite some time: 

http://lists.openstack.org/pipermail/openstack-dev/2015-February/055801.html
https://blueprints.launchpad.net/neutron/+spec/tenant-delete

erhudy:
Change looks reasonable, will set up a test tomorrow.

erhudy:
Tested that deletion now works properly even when the tenant associated with an instance has been deleted (and it logs the error to syslog). Also verified that adding instances still works. Additionally tested what would happen if an instance were created and a float were assigned, then the tenant is deleted - it logs to syslog and does not create an entry in `pdns.records`, which is fine.

Issue #812, opened 2015-09-08
Increase RabbitMQ file descriptor limit


kelvk:
1024 is generally too low. Either increase the default or at least parameterize this.

PR #814, opened 2015-09-08
Parameterize and increase RabbitMQ open files limit

kelvk:
Addresses #812. Note that the Zabbix trigger change will result in creation of a new trigger. The old trigger should be removed (via the UI).


erhudy:
Will test.

erhudy:
Worked as expected on 3-node build. Was not able to test the monitoring change because of Hilarious Mirror Hijinks, but it looks very routine (and the caveat about needing to manually delete the old trigger was already warned about).

Issue #808, opened 2015-09-04
Make toggle for swap on/off


kelvk:
`vm.swappiness=0` reduces tendency to swap, but swapping can still occur. It may be desirable to just completely disable swap.

kamidzi:
+1 for disabling swap.

erhudy:
Merged #815.

PR #815, opened 2015-09-09
Toggle for enabling/disabling swap

kelvk:
Addresses #808.


erhudy:
Testing.

erhudy:
While this works on local builds, I do not think it will work on hardware installs. The swap lines in `/etc/fstab` take the following format there:
```
# swap was on /dev/sda2 during installation
UUID=8f9a1217-ffc9-4690-a291-207f96556faa none            swap    sw              0       0
```

kelvk:
Pushed an updated commit.

erhudy:
Retesting.

erhudy:
Tested locally and on hardware by cheffing in, disabling swap via attribute and recheffing, then enabling swap via attribute and recheffing, in order to verify that it reactivates appropriately and comments/uncomments the correct line in `/etc/fstab`.

PR #805, opened 2015-09-01
Update Cinder default volumes quota from 10 to -1

dkaplan34:
Update Cinder default volumes quota from 10 to -1

PR #817, opened 2015-09-10
Fix quota-class-update command

kamidzi:
This isn't really essential, but it may be nice if we do use the command as per comments in #805. This integrates patch from https://bugs.launchpad.net/python-cinderclient/+bug/1423884

### DIFF
--- a/cookbooks/bcpc/files/default/fix-quota-class-update.patch
+++ b/cookbooks/bcpc/files/default/fix-quota-class-update.patch
@@ -1,0 +1,78 @@
+From 9db3430ccd58370bf5c699f80c98f16e97df1877 Mon Sep 17 00:00:00 2001
+From: liyingjun <yingjun.li@kylin-cloud.com>
+Date: Mon, 2 Mar 2015 16:01:34 +0800
+Subject: [PATCH] Fixes quota-class-update commands
+
+The quota-class-update commands for v1 and v2 are both broken,
+Fix v1 by returning result for update method in QuotaClassSetManager,
+fix v2 by changing arg 'class-name' to 'class_name' for consistent
+with v1.
+
+Change-Id: Ic7188f485b914dc55048cf056a0685618134e87a
+Closes-bug: #1423884
+---
+ cinderclient/v1/quota_classes.py | 6 ++++--
+ cinderclient/v2/quota_classes.py | 6 ++++--
+ cinderclient/v2/shell.py         | 4 ++--
+ 3 files changed, 10 insertions(+), 6 deletions(-)
+
+diff --git a/cinderclient/v1/quota_classes.py b/cinderclient/v1/quota_classes.py
+index 9e81e2c..37441b3 100644
+--- a/cinderclient/v1/quota_classes.py
++++ b/cinderclient/v1/quota_classes.py
+@@ -26,7 +26,7 @@ class QuotaClassSet(base.Resource):
+         return self.class_name
+ 
+     def update(self, *args, **kwargs):
+-        self.manager.update(self.class_name, *args, **kwargs)
++        return self.manager.update(self.class_name, *args, **kwargs)
+ 
+ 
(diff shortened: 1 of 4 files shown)


erhudy:
I will test for the bugfix, though I would like to hold off on merging until the patch provider is merged and this can be adapted to use it.

cookbooks/bcpc/files/default/fix-quota-class-update.patch
```diff
@@ -0,0 +1,78 @@
...
+ 3 files changed, 10 insertions(+), 6 deletions(-)
+
+diff --git a/cinderclient/v1/quota_classes.py b/cinderclient/v1/quota_classes.py
+index 9e81e2c..37441b3 100644
+--- a/cinderclient/v1/quota_classes.py
++++ b/cinderclient/v1/quota_classes.py
+@@ -26,7 +26,7 @@ class QuotaClassSet(base.Resource):
+         return self.class_name
+ 
+xx     def update(self, *args, **kwargs):
```
erhudy: Patch is failing to apply because of this line:
```
  * bash[patch-fix-quota-class-update] action runcinderclient/v1/quota_classes.py: OK
cinderclient/v2/shell.py: OK
cinderclient/v2/quota_classes.py: OK
patch: **** malformed patch at line 26: xx     def update(self, *args, **kwargs):
```

erhudy: Also the resource doesn't actually fail, Chef continues processing, so it took a few trips to figure out what was going wrong.

kamidzi: looks like I committed something I was testing wrt to bad patches. LML

kamidzi:
It should just be moved to patch provider anyways.

erhudy:
Reapplied after moving to the patch resource and it seems happy now - no anger when running `cinder quota-class-update`:
```
# cinder quota-class-update --volumes 10 defaults
+---------------+-------+
|    Property   | Value |
+---------------+-------+
|   gigabytes   |  1000 |
| gigabytes_HDD |   -1  |
| gigabytes_SSD |   -1  |
|   snapshots   |   10  |
| snapshots_HDD |   -1  |
| snapshots_SSD |   -1  |
|    volumes    |   10  |
|  volumes_HDD  |   -1  |
|  volumes_SSD  |   -1  |
+---------------+-------+
```

erhudy:
Do you want to collapse the commits down to 1, or merge as is?

kamidzi:
I think it makes sense to squash in this case

cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -76,6 +76,25 @@
     notifies :restart, "service[cinder-scheduler]", :immediately
 end
 
+# Deal with quota update commands
```
erhudy: After crush, convergence fails:
```
  * cookbook_file[/var/chef/cache/fix-quota-class-update.patch.SHASUMS] action create
    
    Error executing action `create` on resource 'cookbook_file[/var/chef/cache/fix-quota-class-update.patch.SHASUMS]'
    
    Chef::Exceptions::FileNotFound
    ------------------------------
    Cookbook 'bcpc' (5.6.1) does not contain a file at any of these locations:
      files/ubuntu-14.04/fix-quota-class-update.patch.SHASUMS
      files/ubuntu/fix-quota-class-update.patch.SHASUMS
      files/default/fix-quota-class-update.patch.SHASUMS
      files/fix-quota-class-update.patch.SHASUMS
```
This block can be removed entirely. I'm going to remove it locally and continue testing.

kamidzi:
Yeah. bad squash.

erhudy:
Things seemed fine after I removed the problematic block.

kamidzi:
Updated

erhudy:
Recheffed on a fresh build and everything worked as expected, tested out without a problem.

PR #820, opened 2015-09-11
Update Zabbix template for Linux

kelvk:
This introduces three changes:
1. Modified swap free trigger expression to account for [disabled swap](https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/attributes/default.rb#L78).
2. Increase threshold for max running processes
3. Raise alert when number of `nova-compute` processes is out of range 1-2. There is typically only one `nova-compute` in the process table, but the check occasionally raises an alert when a second process is found (presumably a child process, at least according to strace).

Rest of changes are due to Zabbix version upgrade and can be safely ignored.

Old triggers will still need to be removed manually after cheffing in.

### DIFF
--- a/cookbooks/bcpc/files/default/zabbix_bcpc_templates.xml
+++ b/cookbooks/bcpc/files/default/zabbix_bcpc_templates.xml
@@ -4879,7 +4879,7 @@
             <dependencies/>
         </trigger>
         <trigger>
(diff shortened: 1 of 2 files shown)


kelvk:
Commit updated.

erhudy:
Testing.

erhudy:
Cheffed in undramatically. Deleted duplicate triggers manually.

PR #822, opened 2015-09-14
Fixes #816 - boot from volume fails if volume does not match flavor size

### LINKED ISSUE #816: Boot from vol fails if volume size is greater than flavor size
Boot from volume instances fail when the volume size is greater than flavor size

erhudy:
This backports a patch from stable/kilo (merged after 2015.1.1 was released) to nova-api to fix the indicated issue.

I tested via a build off master (with 2015.1.1), and cherry-picking onto 5.6.1 and recheffing onto an existing 5.6.1 lab cluster. Both scenarios worked without incident.

**Testing information**
The symptom of the issue is that if you attempt to boot an instance from a root volume that does not match the size of the root disk for the selected flavor, Nova will fail instance creation with the error `Error: Flavor's disk is too small for requested image. (HTTP 400)`. Applying this patch will fix it.

**To test**
* Create a volume from an image for use as a root volume, intentionally specifying a size that no flavor will have (e.g., 53GB).
* Select a flavor like m1.small that specifies a smaller root volume (20GB).

### DIFF
--- a/cookbooks/bcpc/providers/patch.rb
+++ b/cookbooks/bcpc/providers/patch.rb
@@ -21,18 +21,18 @@ def whyrun_supported?
   true
 end
 
-# expects the name of the patch file and the cookbook (can be nil)
-# returns the path to the staged patch file
-def stage_patch(patch_file, file_cookbook)
-  path = ::File.join Chef::Config[:file_cache_path], patch_file
+# expects the name of the file to stage and the cookbook (can be nil)
+# returns the path to the staged file
+def stage_file(file_to_stage, file_cookbook)
+  path = ::File.join(Chef::Config[:file_cache_path], file_to_stage)
   cookbook = file_cookbook || cookbook_name
 
   Chef::Log.debug("creating #{path}")
   @file_patch_path = Chef::Resource::CookbookFile.new(path, run_context)
-  @file_patch_path.source patch_file
-  @file_patch_path.owner 'root'
-  @file_patch_path.mode 00644
-  @file_patch_path.cookbook cookbook
+  @file_patch_path.source(patch_file)
+  @file_patch_path.owner('root')
+  @file_patch_path.mode(00644)
+  @file_patch_path.cookbook(cookbook)
   @file_patch_path.run_action(:create)
 
   path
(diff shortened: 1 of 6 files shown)


erhudy:
One note: the actual cherry-pick was larger than this, but the remainder of the patchset was changes to the Nova unit tests.

cookbooks/bcpc/recipes/nova-work.rb
```diff
@@ -282,3 +282,29 @@
...
+  mode 00644
+end
+
+bash "patch-backport-for-nova-volume-boot-size" do
+  user "root"
+  code <<-EOH
+    cd /usr/lib/python2.7/dist-packages/
+    cp nova/compute/api.py nova/compute/api.py.prepatch
+    patch -p1 < /tmp/nova-volume-boot-size.patch
+    rv=$?
```
kamidzi: I don't believe this or any of the subsequent lines will be executed if the patch command on the line previous fails.

kamidzi: Also, granted the patches were tested, but what some hunks of the patch fail? Then the file will be in an undefined state...

erhudy: This was a copy and paste from another patch in the same recipe. What about extracting this patch stuff to a resource provider? It's formulaic.

erhudy: I cannot think of a scenario in which patch hunks would fail, given the presence of a guard to only apply the patch if the checksum of the affected file matches exactly.

kamidzi: The patch provider thing is something I had considered before, so I agree that's probably the way to go... As for the hunks failing, it would have to be something like the patch process dying (catching the right signal?) or something with the patchfile itself. *Extremely* Unlikely, but not unheard of  - I wouldn't call the point a showstopper.

erhudy: I am working on extracting the patch logic into a provider and will update this PR when it's ready.

erhudy:
Uploaded a new changeset with a custom patch LWRP. This provider is intended to replace all the other old patches. Scenarios I have tested it with:
* normal proper patching
* bad before/after checksums
* good before checksum but patch fails for some reason

cookbooks/bcpc/providers/patch.rb
```diff
@@ -0,0 +1,100 @@
...
+  @file_patch_path.mode 00644
+  @file_patch_path.cookbook cookbook
+  @file_patch_path.run_action(:create)
+
+  path
+end
+
+action :apply do
+  # stage all files
+  before_checksum_path = stage_patch new_resource.shasums_before_apply, new_resource.file_cookbook
```
kamidzi: I found this to be a bit confusing: calling `stage_patch` with a checksums file. I understand what it's doing, but its a bit confusing given the name of the method and its documentation

erhudy:
Pushed a changeset to rename `stage_patch` to `stage_file`.

kamidzi:
Functionality-wise, I like it. I would like to address some of the minor issues I have with the logging, but I won't hold this up.

PR #823, opened 2015-09-14
Collect Elasticsearch metrics

kelvk:
When deployed, metrics will show up in Graphite under the monitoring servers (bcpc-vm[456]).


erhudy:
Testing alongside #824.

erhudy:
Confirmed that it cheffed in without a problem and there are now `servers.[server-name].elasticsearch.*` metrics for each monitoring server.

PR #824, opened 2015-09-14
python-tz is required for Graphite webapp

### LINKED ISSUE #821: Absolute time range for graphite metrics fail on error 500

kelvk:
Fixes #821. Graphite web should not return a broken image when viewing the absolute time range of any metric.


erhudy:
Looks pretty trivial, building and will merge.

Issue #806, opened 2015-09-02
Absence of configurable default quotas for RGW


kelvk:
RGW users created with Keystone credentials have unlimited quotas. There is a [proposal](http://tracker.ceph.com/issues/12714) upstream to parameterize max-buckets, though other tunable quotas appear absent. In any case, workarounds may need to be considered until our requirements can be fulfilled natively.

kelvk:
Another upstream issue raised - http://tracker.ceph.com/issues/12919.

PR #825, opened 2015-09-15
Manage RGW user quotas

kelvk:
Per #806, quota limits are not enforced by default for users created in RadosGW via Keystone integration. This PR enables and sets either a default or predefined `max_size` [quota](http://ceph.com/docs/master/radosgw/admin/#set-user-quota) for each RGW user. 

Quota enforcement can be tested by adding `tester` to `default['bcpc']['rgw_quota']['user']` with `max_size` of `1`. When applied, `check rgw` will fail on `QuotaExceeded`.


erhudy:
Testing.

erhudy:
It wasn't working at first until I realized the recipe wasn't being included anywhere. Can you update the appropriate role to add **bcpc::rgw-quota**?

erhudy:
After adding the recipe and re-running, got the expected results with tester before changing the quota (check worked) and after dropping the quota to 1:
```
+------+-------+-------+---------------------------------------------------------------------------------+
| name | state |  time |                                       msg                                       |
+------+-------+-------+---------------------------------------------------------------------------------+
| rgw  | ERROR | 0.302 |                          S3ResponseError: 403 Forbidden                         |
|      |       |       | <?xml version="1.0" encoding="UTF-8"?><Error><Code>QuotaExceeded</Code></Error> |
+------+-------+-------+---------------------------------------------------------------------------------+
```

kelvk:
My bad. `BCPC-Headnode` role updated.

PR #827, opened 2015-09-17
Add haproxy BASIC auth for kibana backend

kamidzi:
Simply way to secure kibana via haproxy using BASIC HTTP autentication. It may be the case that
1) This is desired as an optional feature
2) Other haproxied urls will require authentication
The general approach can be extended

To test:
First, both `monitoring-admin-user` and `monitoring-admin-password` will need to be queried from the `configs` data bag. Then,
```
$ curl -I -k https://10.0.100.6/kibana/
HTTP/1.0 401 Unauthorized
Cache-Control: no-cache
Connection: close
Content-Type: text/html
WWW-Authenticate: Basic realm="monitoring.openstack.bcpc.example.com"

$ curl -u monitoring_admin:cgJgggUKago7R68SjHJC -I -k https://10.0.100.6/kibana/
HTTP/1.1 200 OK
X-App-Name: kibana
Accept-Ranges: bytes
Cache-Control: public, max-age=0
Last-Modified: Thu, 30 Jul 2015 18:47:51 GMT
ETag: W/"6f9-14ee04abbd8"
Content-Type: text/html; charset=UTF-8
Content-Length: 1785
Vary: Accept-Encoding
Date: Thu, 17 Sep 2015 16:48:09 GMT
```

### DIFF
--- a/cookbooks/bcpc/templates/default/haproxy-monitoring.cfg.erb
+++ b/cookbooks/bcpc/templates/default/haproxy-monitoring.cfg.erb
@@ -84,7 +84,7 @@ backend kibana-backend
   option httpchk HEAD /elasticsearch/ HTTP/1.1\r\nHost: <%=node['bcpc']['monitoring']['vip']%>
   http-check expect status 200
   acl AuthOkay_kibana http_auth(admins)
-  http-request auth realm <%= "monitoring.openstack.#{node['bcpc']['domain_name']}" %> if !AuthOkay_kibana
+  http-request auth realm <%= node['bcpc']['kibana']['fqdn'] %> if !AuthOkay_kibana
   reqrep ^([^\ :]*)\ /kibana/(.*) \1\ /\2
 <% @servers.each do |server| -%>
   <%= "server #{server['hostname']} #{server['bcpc']['management']['ip']}:5601 check inter 5s rise 1 fall 1" %>
--- a/cookbooks/bcpc/templates/default/haproxy-monitoring.cfg.erb
+++ b/cookbooks/bcpc/templates/default/haproxy-monitoring.cfg.erb
@@ -31,6 +31,9 @@ defaults
   timeout client 30m
   timeout server 30m
 
+userlist admins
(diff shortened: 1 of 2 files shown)


cookbooks/bcpc/templates/default/haproxy-monitoring.cfg.erb
```diff
@@ -80,6 +83,8 @@ backend kibana-backend
   balance source
   option httpchk HEAD /elasticsearch/ HTTP/1.1\r\nHost: <%=node['bcpc']['monitoring']['vip']%>
   http-check expect status 200
+  acl AuthOkay_kibana http_auth(admins)
+  http-request auth realm <%= "monitoring.openstack.#{node['bcpc']['domain_name']}" %> if !AuthOkay_kibana
```
erhudy: I recommend that the auth realm be changed here to use `node['bcpc']['kibana']['fqdn']`, which is what the landing page uses to generate the Kibana domain name.

kamidzi:
updated realm name.

cookbooks/bcpc/templates/default/haproxy-monitoring.cfg.erb
```diff
@@ -80,6 +83,8 @@ backend kibana-backend
   balance source
   option httpchk HEAD /elasticsearch/ HTTP/1.1\r\nHost: <%=node['bcpc']['monitoring']['vip']%>
   http-check expect status 200
+  acl AuthOkay_kibana http_auth(admins)
+  http-request auth realm <%= "#{node['bcpc']['kibana']['fqdn']}" %> if !AuthOkay_kibana
```
erhudy: Minor nit; string interpolation can be removed here now.

kamidzi: removed

erhudy:
Rested one last time and it works as expected. There seems to be a bit of a behavior issue under Chrome where after login the Kibana page doesn't load until you manually refresh; this doesn't seem to happen in Safari or Firefox.

PR #836, opened 2015-09-24
Add Diamond openstack collector

semenovroman:
Adds Diamond collector that sends general openstack stats to graphite. 
To test: build cluster with 'MONITORING_NODES=1', open graphite, you should see metrics under 'BC2' (might have to wait up to 15 minutes for metrics to appear - default collection interval)

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -1350,6 +1350,13 @@ default['bcpc']['diamond']['collectors']['rabbitmq']['queues_ignored'] = '.*'
 default['bcpc']['diamond']['collectors']['rabbitmq']['vhosts'] = nil
 # Ceph Collector parameters
 default['bcpc']['diamond']['collectors']['CephCollector']['metrics_whitelist'] = "ceph.mon.#{node['hostname']}.cluster.*"
+# BC2 Collecter parameters
+default['bcpc']['diamond']['collectors']['BC2'] = {
+  "interval" => "900",
+  "path" => "openstack",
+  "hostname" => "#{node['bcpc']['region_name']}",
+  "db_host" => "#{node['bcpc']['management']['vip']}",
+}
 
 
 ###########################################
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -1350,8 +1350,8 @@ default['bcpc']['diamond']['collectors']['rabbitmq']['queues_ignored'] = '.*'
 default['bcpc']['diamond']['collectors']['rabbitmq']['vhosts'] = nil
 # Ceph Collector parameters
 default['bcpc']['diamond']['collectors']['CephCollector']['metrics_whitelist'] = "ceph.mon.#{node['hostname']}.cluster.*"
-# BC2 Collecter parameters
-default['bcpc']['diamond']['collectors']['BC2'] = {
+# Openstack Collector parameters
+default['bcpc']['diamond']['collectors']['cloud'] = {
   "interval" => "900",
   "path" => "openstack",
   "hostname" => "#{node['bcpc']['region_name']}",
--- a/cookbooks/bcpc/files/default/diamond-collector-BC2-openstack.py
+++ b/cookbooks/bcpc/files/default/diamond-collector-BC2-openstack.py
@@ -1,0 +1,76 @@
+import diamond.collector
+import MySQLdb as mysql
+
+class BC2Collector(diamond.collector.Collector):
+
+    def get_default_config_help(self):
+        config_help = super(BC2Collector, self).get_default_config_help()
+        config_help.update({
+            'bc2_collector': 'Send BC2 openstack stats',
+        })
+        return config_help
+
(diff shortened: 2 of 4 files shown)


kamidzi:
I ran into the following:
```
[2015-09-28 21:53:36,793] [MainThread] Failed to import module: BC2. Traceback (most recent call last):
  File "/usr/lib/pymodules/python2.7/diamond/server.py", line 190, in load_collectors
    mod = __import__(modname, globals(), locals(), ['*'])
  File "/usr/share/diamond/collectors/BC2/BC2.py", line 2, in <module>
    import mysql.connector
ImportError: No module named connector
```

I'll investigate further

erhudy:
For our environment I think the desired import is `from MySQLdb import connector` (or you can set the connection object up by importing the entire MySQLdb module and instantiating with `conn = MySQLdb.connect(...)`, assuming the intent is to use MySQLdb as installed by the **python-mysqldb** module.

kelvk:
Minor comment: Can we maintain a generic name for the collector, such as "Openstack"?

kamidzi:
The issue seems to be with the way Diamond is loading plugins. It seems that at the point where BC2 is to be loaded as a module, the system include path includes other paths that interfere with the loading of the mysql module in `/usr/lib/python2.7/dist-packages`. Interestingly enough, adding that standard location to the front of the `sys.path` list doesn't appear to make a difference, or even restricting it to just that.

cookbooks/bcpc/templates/default/diamond.conf.erb
```diff
@@ -210,6 +210,16 @@ enabled = True
...
 
+[[BC2Collector]]
+enabled = True
+interval = 900
+path = openstack
+path_prefix = BC2
+hostname = <%=node['bcpc']['region_name']%>
+db_user = <%=get_config('mysql-root-user')%>
+db_password = <%=get_config('mysql-root-password')%>
+db_host = <%=node['bcpc']['management']['ip']%>
```
kamidzi: I think what you want here is `node['bcpc']['management']['vip']`, unless you want edit the access privs as well. See the following:
```
chef >  [ node['bcpc']['management']['ip'], node['bcpc']['monitoring']['vip'], node['bcpc']['management']['vip'] ].each do |ip|
chef > puts %x{ mysql -u root -p#{get_config('mysql-root-password')} -h #{ip} -e 'show databases' }
chef ?> end
Warning: Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'root'@'bcpc-vm4.hypervisor-bcpc.example.com' (using password: YES)

Warning: Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'root'@'10.0.100.6' (using password: YES)

Warning: Using a password on the command line interface can be insecure.
Database
information_schema
cinder
glance
heat
horizon
keystone
mysql
nova
pdns
performance_schema
test
 => ["10.0.100.14", "10.0.100.6", "10.0.100.5"]```

**Comment by kamidzi** [comment]
Also, I would move the collector config to a fragment, as per https://github.com/bloomberg/chef-bcpc/pull/830/files#diff-4ea3df38690c16fea3415fcbfacc9895R106

**Comment by kamidzi** [comment]
Regardless, I was able to get it working using @erhudy's suggestion with the following patch:
```
From cabce95c0975448eb56395823da496206898853e Mon Sep 17 00:00:00 2001
Date: Thu, 8 Oct 2015 10:19:09 -0400
Subject: [PATCH] Use MySQLdb instead of mysql-connector-python

This will avoid module name conflicts introduced by injection of
additional directories to sys.path by Diamond server plugin loading
mechanism.
---
 cookbooks/bcpc/files/default/diamond-collector-BC2-openstack.py | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/cookbooks/bcpc/files/default/diamond-collector-BC2-openstack.py b/cookbooks/bcpc/files/default/diamond-collector-BC2-opensta
index 165dbb7..55cd320 100644
--- a/cookbooks/bcpc/files/default/diamond-collector-BC2-openstack.py
+++ b/cookbooks/bcpc/files/default/diamond-collector-BC2-openstack.py
@@ -1,5 +1,5 @@
 import diamond.collector
-import mysql.connector
+import MySQLdb as mysql

 class BC2Collector(diamond.collector.Collector):

@@ -27,7 +27,7 @@ class BC2Collector(diamond.collector.Collector):


     def collect(self):
-        cnx = mysql.connector.connect(user=self.config['db_user'], password=self.config['db_password'], host=self.config['db_host'])
+        cnx = mysql.connect(user=self.config['db_user'], passwd=self.config['db_password'], host=self.config['db_host'])
         cursor = cnx.cursor()

         # Glance total images size
--
1.8.5.2 (Apple Git-48)
```

I do agree with @kelvk, that the structure in the UI may need some tweaking.

semenovroman:
I am open to suggestions regarding the module name.

kamidzi:
@kelvk - this is more your area: suggestions on the name?

kelvk:
I propose using a generic name for now. "Cloud" comes to mind.

PR #837, opened 2015-09-24
Add iptables safety valve that allows appending rules before the end

erhudy:
While ugly, the use case for this is to allow the addition of arbitrary iptables rules before the end of the iptables INPUT chain. (The specific use case here is to allow certain external hosts access to MySQL to obtain data for statistics purposes without adding 3306 to the list of ports open to the world.)


kelvk:
Testing.

erhudy:
Killing this off.

PR #839, opened 2015-09-24
Bump rabbitmq-server version to 3.5.5-3

kamidzi:
This isn't thoroughly tested yet, but it *is* necessary to complete build


kamidzi:
I don't see any glaring problems based on successful execution of the baseline rally tests documented before the upgrade.

erhudy:
Built with RabbitMQ 3.5.5 locally and everything is working as expected.

Issue #840, opened 2015-09-25
haproxy.cfg for monitoring re-generated on each Chef run


kelvk:
The configuration file is changed on each Chef run, resulting in restart of haproxy. We may need to persist the password string (https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/templates/default/haproxy-monitoring.cfg.erb#L35) in the data bag.

kamidzi:
Good catch! Looks like it's because the salt is being regenerated. The hash should be constant given a constant salt.

PR #841, opened 2015-09-25
open 8443/TCP for a web server for external integrations

erhudy:
This replaces #837 and opens 8443/TCP to allow running arbitrary servers within a cluster that provide endpoints for external integrations.


erhudy:
Will augment with basic auth for the HAProxy endpoint.

erhudy:
(Obviously the branch name is now a lie.)

erhudy:
Affected by #840, so the fix from that will be applied here as well.

PR #842, opened 2015-09-28
List buckets with prefix and log on failures in rgw check

kelvk:
`check rgw` occasionally fails, so adding some logging for debugging purposes. We will also prefix `bucket.list()` with key name since rest of bucket contents are not of interest to the check.

### DIFF
--- a/cookbooks/bcpc/files/default/checks/rgw
+++ b/cookbooks/bcpc/files/default/checks/rgw
@@ -6,6 +6,7 @@ import boto.s3.connection
 import unittest
 import uuid
 import time
+import syslog
 
 from numpy.random.mtrand import RandomState
 import binascii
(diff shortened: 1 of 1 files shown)


kamidzi:
Maybe, we could avoid all that looping altogether... something like the following?
```
--- a/cookbooks/bcpc/files/default/checks/rgw
+++ b/cookbooks/bcpc/files/default/checks/rgw
@@ -34,7 +34,8 @@ class TestRadosGateway(object):
             syslog.syslog(errmsg)
             return False, errmsg

-        if self.bucket_name not in [ b.name for b in self.conn.get_all_buckets()] :
+        bucket = self.conn.lookup(self.bucket_name, validate=False)
+        if bucket is None:
             errmsg = 'Unable to find bucket ' + self.bucket_name
             syslog.syslog(errmsg)
             return False, errmsg
@@ -59,7 +60,8 @@ class TestRadosGateway(object):
             syslog.syslog(errmsg)
             return False, errmsg

-        if key_name not in [key.key for key in bucket.list(prefix=key_name)]:
+        res = bucket.get_key(key_name)
+        if res is None:
             errmsg = 'Key ' + key_name + ' not returned in list bucket contents'
             syslog.syslog(errmsg)
             return False, errmsg
```

kelvk:
There are permission grants for listing operations to work, so it seems worthy to exercise these functionality.

kamidzi:
Speaking to @kelvk, it seems that in production, the loops are not all that time-consuming in addition to the prior comments.

PR #844, opened 2015-09-29
Convert DOS line endings to UNIX when reading in CA certificates

erhudy:
When a cert with DOS line endings is written out, it gets a blank line inserted after every real line and `update-ca-certificates` chokes. This fixes that.


kamidzi:
I've added a small change to also remove the trailing blank line.

PR #845, opened 2015-09-29
Replaces HAProxy config template with concat_fragment setup

erhudy:
This will allow building the HAProxy configuration out of multiple template/file fragments without needing to do things like hack up the main template with safety valve insertion points.

Running this on an existing or new build is a no-op; everything should continue to work the same. If you want to test adding things to the config from other recipes, try something like this:

```
#
# Cookbook Name:: bcpc
# Recipe:: haproxy-test
#
concat_fragment "haproxy-test-config" do
  order  "002"
  target "/etc/haproxy/haproxy.cfg"
  source "haproxy-test"
  source_provider Chef::Resource::CookbookFile
  notifies :create, "concat[/etc/haproxy/haproxy.cfg]", :immediately
end
```

Then add `files/default/haproxy-test` with some content, add `recipe[bcpc::haproxy-test]` to the run list for a head node, and rechef. The additional content will be appended to the end of the HAProxy configuration and HAProxy will be restarted. Further Chef runs should not converge any of the resources, because nothing's changing.


kelvk:
Testing

kelvk:
Played with some fragments (both files and templates) and fragment ordering and everything works as intended. Cheffed in cleanly and as expected, it is a no-op. This also builds fine from scratch. This presumably will go into the next release, but I'll hold merge for a bit should @kamidzi wishes to review as well.

erhudy:
The main gotcha with concat is, like the Puppet concat module I based the cookbook off, you can get into some weird situations with orphaned fragments if you don't mark them as `:delete` so that the cookbook can clean them up. Otherwise, you will wind up with these orphans hanging around the staging directory and being unexpectedly concatenated into the final file.

PR #846, opened 2015-09-29
adding 8080/TCP and 8443/TCP to default open ports

erhudy:
This replaces the more extensive PR in #841 with something that just opens 8080/8443 as hooks for additional services to run on.


erhudy:
Better idea! Moving the list of ports to an attribute.

PR #848, opened 2015-09-29
Move firewall TCP ports out of template into attribute

erhudy:
Third time's the charm, right? This moves the list of open TCP ports in iptables from the template to a list attribute, so that it can be adjusted as needed in the environment/roles.


kelvk:
Testing.

kelvk:
Works as expected.

PR #854, opened 2015-09-30
Create region in keystone.region table if it is not there already.

### LINKED ISSUE #853: Transitioning from template to SQL backend for Keystone service catalog can fail due to missing region
When testing the upgrade path from 5.6.1 to 5.7.0, I found that the Keystone endpoint creation step consistently failed with 500 errors from Keystone. Reviewing the logs showed a constraint failure:
```
2015-09-30 14:38:39.529 9971 ERROR keystone.common.wsgi [-] (IntegrityError) (1452, 'Cannot add or update a child row: a foreign key constraint fails (`keystone`.`endpoint`, CONSTRAINT `fk_endpoint_region_id` FOREIGN KEY (`region_id`) REFERENCES `region` (`id`))') 'INSERT INTO endpoint (id, legacy_endpoint_id, interface, region_id, service_id, url, enabled, extra) VALUES (%s, %s, %s, %s, %s, %s, %s, %s)' ('f6c25e3e7192411d83d43d78caf9763f', 'ba82bcfca86b4fa799289d6ba8331d0c', 'admin', 'XXXX', '2c03585ef00e4fb38ac841cbbf42dd8a', 'https://openstack.testing.domain.com:9292/v2', 1, '{}')
```

I investigated the database and found that the **keystone.region** table was empty. On a local build that had been built to use the SQL backend from the start, this table was populated with a single row as follows:
```
mysql> select * from region;
+------+-------------+------------------+-------+
| id   | description | parent_region_id | extra |
+------+-------------+------------------+-------+
| TLVP |             | NULL             | {}    |
+------+-------------+------------------+-------+
```
Checking another running cluster that had been bootstrapped from the template backend showed this table to be empty.

Inserting the row into the database manually allows things to proceed:
```
INSERT INTO keystone.region (id, description, parent_region_id, extra) VALUES ('XXXX', '', NULL, '{}');
```

semenovroman:
SQL is idempotent, can be safely run multiple times. Fixes #853

### DIFF
--- a/cookbooks/bcpc/recipes/keystone.rb
+++ b/cookbooks/bcpc/recipes/keystone.rb
@@ -170,6 +170,15 @@ bash "keystone-database-sync" do
     notifies :restart, "service[apache2]", :immediately
 end
 
+ruby_block "keystone-region-creation" do
+    block do
+        %x[ export MYSQL_PWD=#{get_config('mysql-root-password')};
+            mysql -uroot -e "INSERT INTO keystone.region (id, extra) VALUES(\'#{node['bcpc']['region_name']}\', '{}');"
+        ]
+    end
+    not_if { system "MYSQL_PWD=#{get_config('mysql-root-password')} mysql -uroot -e 'SELECT id FROM keystone.region WHERE id = \"#{node['bcpc']['region_name']}\"' | grep \"#{node['bcpc']['region_name']}\" >/dev/null" }
+end
+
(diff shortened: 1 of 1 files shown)


erhudy:
Can you rework this with not_if/only_if so that Chef will not converge the resource every time?

semenovroman:
@erhudy I force-pushed an updated version that ditches mysql "insert iff region does not exist" in favor of chefs `not_if`

erhudy:
Tested and it is appropriately idempotent and does not blow things up.

PR #857, opened 2015-10-05
Zabbix integration with Pagerduty

kelvk:
This PR provides a Zabbix script-based notification resource and configures Pagerduty integration (if enabled). Use of Pagerduty integration requires a service key. See https://www.pagerduty.com/docs/guides/zabbix-integration-guide/ for setup instructions.

The following Zabbix elements are setup when Pagerduty is enabled:
1. Pagerduty script - https://10.0.100.6/zabbix/media_types.php
2. Association of Pagerduty script media with Admin user - https://10.0.100.6/zabbix/users.php
3. Pagerduty trigger action - https://10.0.100.6/zabbix/actionconf.php

Note that neither update nor deletion of above Zabbix elements are supported by `bcpc_zbxnotify`.

A test fault can be simulated by executing `stop nova-compute` on any head/work node. Wait for Pagerduty's notification. When satisfied, execute `start nova-compute` and the issue should auto-resolve in Pagerduty.


cookbooks/bcpc/attributes/default.rb
```diff
@@ -201,6 +201,10 @@
 default['bcpc']['ntp_servers'] = ["pool.ntp.org"]
 default['bcpc']['dns_servers'] = ["8.8.8.8", "8.8.4.4"]
 
+# Proxy server for recipes to use
+# Example: http://proxy-hostname:port
+default['bcpc']['proxy_server'] = nil
```
kamidzi: Throwback? It's thursday! https://github.com/bloomberg/chef-bcpc/blob/master/docs/legacy_bootstrap_doc.md ;)

kelvk: I intended to deprecate the old attribute in favor of the new, as use of proxy extends beyond bootstrapping.

cookbooks/bcpc/providers/zbxnotify.rb
```diff
@@ -0,0 +1,143 @@
...
+    mt = zbx_api(auth, 'mediatype.create', params)
+    mediatypeid = mt['mediatypeids'][0]
+    Chef::Log.info("Mediatype #{@new_resource.name} created: #{mediatypeid}")
+  else
+    mediatypeid = mt[0]['mediatypeid']
+    Chef::Log.info("Mediatype #{@new_resource.name} already exists: #{mediatypeid}")
+  end
+
+  # Obtain admin UID
+  users = zbx_api(auth, 'user.get', {})
```
kamidzi: What about just getting admin user directly instead of looping? It seems like it is also--for better or worse--case-insensitive:

```
chef > zbx_api(token,'user.get',{filter: {alias: 'Admin'}})
 => [{"userid"=>"1", "alias"=>"Admin", "name"=>"Zabbix", "surname"=>"Administrator", "url"=>"", "autologin"=>"1", "autologout"=>"0", "lang"=>"en_GB", "refresh"=>"30", "type"=>"3", "theme"=>"default", "attempt_failed"=>"0", "attempt_ip"=>"", "attempt_clock"=>"0", "rows_per_page"=>"50"}]
chef > zbx_api(token,'user.get',{filter: {alias: 'admin'}})
 => [{"userid"=>"1", "alias"=>"Admin", "name"=>"Zabbix", "surname"=>"Administrator", "url"=>"", "autologin"=>"1", "autologout"=>"0", "lang"=>"en_GB", "refresh"=>"30", "type"=>"3", "theme"=>"default", "attempt_failed"=>"0", "attempt_ip"=>"", "attempt_clock"=>"0", "rows_per_page"=>"50"}]
```

cookbooks/bcpc/recipes/zabbix-pagerduty.rb
```diff
@@ -0,0 +1,66 @@
...
+#
+
+if node['bcpc']['monitoring']['pagerduty']['enabled'] then
+
+    # Pagerduty integration script
+    cookbook_file '/usr/lib/zabbix/alertscripts/pagerduty.py' do
+        source 'bins/pagerduty-zabbix-proxy.py'
+        owner 'root'
+        mode 00755
+        not_if 'test -f /usr/lib/zabbix/alertscripts/pagerduty.py'
```
kamidzi: This test may cause problems if the file already exists but has wrong contents. `cookbook_file` should handle it properly regardless (and ensure the contents)

cookbooks/bcpc/providers/zbxnotify.rb
kelvk: Definitely. Rewritten. The case insensitivity is not an issue.

cookbooks/bcpc/recipes/zabbix-pagerduty.rb
kelvk: The conditional exists because of subsequent `bcpc_patch` action, to avoid repetitive patching. I've made the test more specific - it checks for the the patched file's shasum.

kelvk:
Pushed an updated commit in response to above comments. Note that this also updates `shared_build_bins` to actually copy the Pagerduty script into the cookbook.

cookbooks/bcpc/attributes/default.rb
kamidzi: This might be bit clearer with something like proxy_server_url, since we aren't really in the habit of documenting cookbook attributes in the metadata. I do recognize and appreciate the comment above, but I'm sure if the variable is seen outside of this file, it may cause some confusion

kamidzi:
Some minor final nitpicks:

* The patchfile contains some extraneous spaces on would-be empty lines
* Some inconsistency with the way raise() is being called - parenthesized args vs not
* The naming of `proxy_server` variable

Again, these are minor. If time is tight either way, I'll merge tomorrow.

kelvk:
* Removed extraneous whitespaces
* Fixed inconsistency with a raise()
* Renamed `proxy_server` to `proxy_server_url`

PR #858, opened 2015-10-07
Parameterize nova default log levels

kelvk:
Parameterizes `default_log_levels` (http://docs.openstack.org/kilo/config-reference/content/list-of-compute-config-options.html) to provide more granular control over logging requirements. For example, `"oslo_concurrency.lockutils=DEBUG, oslo_concurrency.processutils=DEBUG"` can be set to log at DEBUG for oslo_concurrency module without enabling debug [globally](https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/attributes/default.rb#L551).

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -549,6 +549,8 @@ default['bcpc']['nova']['live_migration_patch'] = false
 default['bcpc']['nova']['verbose'] = false
 # Nova debug toggle
 default['bcpc']['nova']['debug'] = false
+# Nova default log levels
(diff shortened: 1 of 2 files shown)


kamidzi:
Maybe update the commented defaults to match http://docs.openstack.org/kilo/config-reference/content/list-of-compute-config-options.html ? Otherwise LGTM

kelvk:
Updated commit to omit `default_log_levels` entry by default. I thought of facilitating a no-op change, but in retrospect it is unnecessary since bouncing nova daemons is mostly non-disruptive.

PR #862, opened 2015-10-08
Fixes #859 - improves VIP change script behavior

### LINKED ISSUE #859: keepalived vip_* scripts don't always clean up completely
When a VIP migration occurs, the `vip_*` scripts in `/usr/local/bin` are supposed to gracefully handle the lifecycles of certain services that need to be started/stopped/restarted during these migrations. In practice it seems that they are susceptible to failure during network flap conditions, and so we probably need to augment them (`vip_change` in particular) to force synchronization at certain points to ensure that we're in the desired state instead of just relying on Upstart and all the service init scripts to do the right thing.

erhudy:
- merged `vip_won`/`vip_lost`/`vip_change` into one script
- avoid launching multiple instances of `vip_lost`/`vip_won` in the event of
  network interface flapping
- avoid attempting to repeatedly restart `td-agent`
- clean up any orphaned `td-agent` processes caused by the PID file being
  overwritten (this should no longer occur with the above changes, but
  will attempt to clean up just in case)
- call `/usr/local/bin/hup_openstack` instead of separately enumerating
  OpenStack services


erhudy:
Testing information
---

You can simulate VIP flapping on a local build with a single head node by repeatedly changing the state of eth1, e.g. `for i in $(seq 1 10); do ifdown eth1 && sleep 1 && ifup eth1; done`. With the old scripts, you should see numerous instances of `vip_won`/`vip_lost`. After replacing them with this PR, you should see only one instance of `vip_lost` and one of `vip_won`, and `vip_won` will disappear before `vip_lost` (because `vip_lost` is handling the restart of `td-agent`).

Additionally, if you want to test terminating untracked `td-agent` processes, you can spawn them by editing `/var/run/td-agent/td-agent.pid` and then poking the init script to start another set of processes.

mihalis68:
I have only read this, not tested it, but LGTM

erhudy:
@kelvk is scheduled to test it at some point, since it mostly affects the behavior of monitoring-related stuff.

kelvk:
I don't think we actually need to restart td-agent when the VIP jumps. I will review and test in any case.

kelvk:
td-agent, diamond and zabbix-agent can be omitted from `vip_change` script. LGTM otherwise.

erhudy:
All that work on detecting multiple td-agent processes for naught! I will amend accordingly.

kelvk:
Failover and process restarts work as expected.

PR #864, opened 2015-10-13
Add PowerDNS master/slave capability

kelvk:
Initial draft to add master/slave capability to PowerDNS. This allows hosting of slave zones on an independent cluster.

This also corrects the NS records, as the value should not be a numerical IP address.

Before:
<pre>
;; ANSWER SECTION:
bcpc.example.com.       300     IN      NS      10.0.100.5.
</pre>

After:
<pre>
;; ANSWER SECTION:
bcpc.example.com.       300     IN      NS      openstack.bcpc.example.com.
</pre>

Marking this as do-not-merge until I can write up a test approach.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -207,6 +207,21 @@ default['bcpc']['proxy_server_url'] = nil
 
 ###########################################
 #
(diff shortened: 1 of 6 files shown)


kelvk:
Closing for now.

PR #865, opened 2015-10-13
Add reserved tcp ports

kamidzi:
Prevents potential confict with expanded ephemeral port range:

zabbix      10050,10051
memcached   11211
rabbitmq    15672,55672
libvirtd    16509,16514
erlang      25672
keystone    35357,35358

Also allows configuration of ports additional to the hard-coded defaults above. Initially, I toyed with idea of calling `uniq` on the array/summarizing the ranges, but the kernel is generous enough to do so. Also, no doing so will make duplicates in the attributes file more obvious.

If everything works, this should be the following output by default:
```
$ cat /proc/sys/net/ipv4/ip_local_reserved_ports
10050-10051,11211,15672,16509,16514,25672,35357-35358,55672
```

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -1275,6 +1275,16 @@ default['bcpc']['mysql-head']['max_connections'] = 0
 
 ###########################################
 #
+# BCPC system (systcl) settings
+#
+###########################################
+#
+# Use this to *add* more reserved ports; i.e. modifiy value of
+# net.ipv4.ip_local_reserved_ports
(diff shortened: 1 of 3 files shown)


erhudy:
Without this patch, no reserved ports as expected:
```
$ cat /proc/sys/net/ipv4/ip_local_reserved_ports 

```

erhudy:
Tested adding and removing additional ports via `['bcpc']['system']['additional_reserved_ports']` successfully.

erhudy:
Cheffed onto an existing cluster undramatically, testing fresh build.

erhudy:
Clean build also worked fine.

PR #866, opened 2015-10-15
Increase default MySQL connections to ease multi-headnode deployments

erhudy:
This is a small change to boost the number of MySQL connections (again) to ease multi-head node deployments, by setting the minimum number of connections equivalent to a 3-head node cluster. This allows us to avoid the annoyance of having to rechef head nodes in some sort of pyramidal order in order  to avoid random convergence failures and 500 responses from APIs due to connections running out.


mihalis68:
Looks good to me

PR #869, opened 2015-10-20
upgrading Ceph to 0.94.4

erhudy:


erhudy:
Merging so that master will not blow up on this.

semenovroman:
tested on a fresh build - works as expected

PR #871, opened 2015-10-20
Various fixes to smooth 2015.1.0 -> 2015.1.1 upgrade

erhudy:
This accommodates a couple of failure cases I encountered when doing an in-place upgrade of OpenStack 2015.1.0 to 2015.1.1:

- removed compromised Ceph signing key from existing systems
- provided dpkg options for Glance/Heat upgrades that automatically
  answer the questions about whether to keep or replace config files
  (these options will keep ours, which are managed by Chef)
- added a terrible hack to distinguish installing Horizon vs upgrading
  it and ignoring failures when upgrading because static asset
  compression bombs out (known upstream issue)

### DIFF
--- a/cookbooks/bcpc/recipes/ceph-common.rb
+++ b/cookbooks/bcpc/recipes/ceph-common.rb
@@ -24,6 +24,15 @@ apt_repository "ceph" do
   distribution node['lsb']['codename']
   components ["main"]
   key "ceph-release.key"
+  notifies :run, "execute[apt-get update]", :immediately
+end
+
+# delete the compromised Ceph signing key from existing systems
+bash "remove-old-ceph-key" do
+  code <<-EOH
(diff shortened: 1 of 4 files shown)


erhudy:
Found additional issues around the Horizon package upgrade that will need resolved.

erhudy:
Cause of the issues in upgrading Horizon is because we move the Apache config file from conf-available to sites-available. Will move it back and fix things up here.

erhudy:
Amended to fix Horizon upgrade issues for realsies this time, safe to review again.

erhudy:
Finding some other issues (packages that should be upgraded being kept back when they should be upgraded, e.g. QEMU).

erhudy:
Too many issues with this right now, and package version pinning has been a terrible mess, so I am going to withdraw this PR and rework it.

PR #872, opened 2015-10-20
Remove use_bootstrap_v2 toggle that is no longer needed

erhudy:
This toggle originally gated more things, but they've all been removed, and we can safely use pip 7.0.3 for Rally.


kamidzi:
Also removed the attribute from the environment.

PR #875, opened 2015-10-22
Removing explicit package versions, adding repository pins

erhudy:
Explicit package version pinning was never successful and only made it more difficult to build master, especially with badly behaved upstream repositories that purged older versions of packages when new ones were released. In practice, we control the versions of software that are installed on clusters via local mirrors.

**Two important notes**: 

1. deploying this to an existing cluster will uncork an upgrade to Erlang 18.x that was previously being thwarted. If deploying to a multi-head node cluster, please read `docs/upgrading_rabbitmq_or_erlang.md` (included in this pull request).
2. Repository pins have been added to encourage apt to prefer packages from those repositories using **origin**. An unfortunate side effect of maintaining a local mirror is that every locally mirrored repository will have the same origin, negating the repository pin. (It is not possible to specify anything more than a simple hostname for the origin, which can be verified by editing the origin statement and then checking `apt-cache policy` output for an affected package.) This does not seem to be a problem at present, but could for example cause issues if the same version of Ceph were provided by both the Ceph mirror and Ubuntu Cloud Archive and apt for some reason decided to newly prefer the one from UCA over Ceph. I am open to suggestions on how to work around this.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -36,19 +36,6 @@ default['bcpc']['s3_ssl_intermediate_certificate'] = nil
 
 ###########################################
 #
-# Package versions
-#
-###########################################
-default['bcpc']['elasticsearch']['version'] = '1.5.1'
-default['bcpc']['ceph']['version'] = '0.94.5-1trusty'
-default['bcpc']['ceph']['version_number'] = '0.94.5'
-default['bcpc']['erlang']['version'] = '1:17.5.3'
-default['bcpc']['haproxy']['version'] = '1.5.15-1ppa1~trusty'
-default['bcpc']['kibana']['version'] = '4.0.2'
-default['bcpc']['rabbitmq']['version'] = '3.6.0-1'
-
-###########################################
-#
 #  Flags to enable/disable BCPC cluster features
 #
 ###########################################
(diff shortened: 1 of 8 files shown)


erhudy:
One possible workaround is specifying `default_release` explicitly on certain packages like Ceph that are provided from multiple locations. It may be possible to specify the release here that corresponds to the release name on the Ceph repo rather than the UCA repo. Will test and find out.

erhudy:
Found another interesting unintended consequence of pinning when using a local mirror: trusty-backports is now promoted to the same priority as other repositories due to the origin pin from other repos all matching the local mirror address, so this means all sorts of other interesting packages are sneaking in.

erhudy:
Force pushed a change on this branch that removes the repository pins. For Ceph packages, `default_release` is set to `trusty` to direct it to the Ceph upstream repo instead of UCA (where the release options are `trusty-proposed` or `trusty-updates`).

kamidzi:
I can confirm this works as a fresh build with no local package mirror.

erhudy:
With #886 being merged now, I have rebased this branch to fix the merge conflict in `ceph-common.rb`. On our current production clusters, deploying this would result in an Erlang upgrade, but that should be it.

erhudy:
After testing, deploying this does not actually result in an Erlang upgrade as long as the RabbitMQ version does not change. Since I removed the explicit Erlang package resources and made it a transitive dependency of RabbitMQ, Erlang will only get upgraded if RabbitMQ is upgraded.

kamidzi:
My only concern with this is that we lose the ability to canonically identify which versions of whatever packages whose versions have been specified were intended for any particular release.. though its probably sufficient to simply record it in the release notes.

kamidzi:
cheffed onto 5.10.0 cleanly (expectedly).

erhudy:
Yeah, it's not ideal. Even with the pinned packages, there were a lot of versions we didn't pin, so it was very selective with regards to what it would try to enforce a particular version for. We can probably work around some of it by doing something like you said, like publishing a list of software versions that a particular release has been tested with.

PR #876, opened 2015-10-23
Adding an early check to verify local package mirror is available

erhudy:
This is a minor fix in the Vagrantfile that will save me a lot of headaches, because I forget to restart Nginx after a computer reboot all the time.

### DIFF
--- a/bootstrap/vagrant_scripts/Vagrantfile
+++ b/bootstrap/vagrant_scripts/Vagrantfile
@@ -229,7 +229,7 @@ Vagrant.configure("2") do |config|
       unless ENV['BOOTSTRAP_APT_MIRROR'].nil? or ENV['BOOTSTRAP_APT_MIRROR'].empty?
         m.vm.provision "test-local-package-mirror", type: "shell" do |s|
           s.privileged = false
-          s.inline = "curl -s -S #{ENV['BOOTSTRAP_APT_MIRROR']} 1>/dev/null"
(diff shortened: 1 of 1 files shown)


bootstrap/vagrant_scripts/Vagrantfile
```diff
@@ -35,11 +35,11 @@ config_defaults = File.join(config_path, 'bootstrap_config.sh.defaults')
 config_overrides = File.join(config_path, 'bootstrap_config.sh.overrides')
 
 File.open(config_defaults).each do |line|
-    extract_and_export_envvar(line)
+  extract_and_export_envvar(line)
 end
 if File.exist?(config_overrides)
   File.open(config_overrides).each do |line|
-      extract_and_export_envvar(line)
+    extract_and_export_envvar(line)
```
erhudy: this and above are unrelated indentation fixes that were annoying me

kamidzi:
Added a timeout equivalent to half the normal 10000 millis used elsewhere. Otherwise it can take quite some time to fail...

Exporting a bogus value for BOOTSTRAP_APT_MIRROR:

```
$ vagrant up
...
==> vm3: curl: (28) Connection timed out after 5000 milliseconds
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

erhudy:
Looks good to me.

Issue #877, opened 2015-10-23
Patch issues with 2015.1.2


erhudy:
2015.1.2 was released upstream. Opening this issue to track patches that do not apply cleanly.

erhudy:
`bcpc_patch "nova-volume-boot-size"` does not apply cleanly. Not needed for 2015.1.2, will gate it to only apply on 2015.1.0 or 2015.1.1.

PR #878, opened 2015-10-23
Gate nova-volume-boot-size patch to only apply to 2015.1.0/2015.1.1

erhudy:
This is the only patch that does not apply cleanly onto 2015.1.2, because it is already included in 2015.1.2. Added gate to prevent it from applying to versions other than 2015.1.0 and 2015.1.1.

### DIFF
--- a/cookbooks/bcpc/recipes/nova-work.rb
+++ b/cookbooks/bcpc/recipes/nova-work.rb
@@ -310,10 +310,12 @@ end
 
 # this patch patches Nova to work correctly if you attempt to boot an instance from
 # a root volume larger than the root volume specified by the flavor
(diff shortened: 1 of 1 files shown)


erhudy:
Fixes #877.

kamidzi:
Otherwise the following bit of discouragement occurs:
```
    Error executing action `apply` on resource 'bcpc_patch[nova-volume-boot-size]'

    RuntimeError
    ------------
          errors raised by shasum both before and after:
          BEFORE: shasum: WARNING: 1 computed checksum did NOT match

          AFTER: shasum: WARNING: 1 computed checksum did NOT match
```

PR #879, opened 2015-10-23
Fixes #513 - makes scrub load threshold configurable

### LINKED ISSUE #513: Deepscrubbing threshold is too low.
As we have combined compute and ceph, the load threshold for scrubbing is basically useless for us. Its set currently at 0.5, which on almost any BCPC node never happens. 

http://ceph.com/docs/master/rados/configuration/osd-config-ref/#scrubbing

erhudy:
This fixes a longstanding issue where scrubs on OSDs on many machines would never happen, because the load threshold is set at a very low 0.5. This makes the threshold configurable.

### DIFF
--- a/cookbooks/bcpc/attributes/ceph.rb
+++ b/cookbooks/bcpc/attributes/ceph.rb
@@ -12,6 +12,7 @@ default['bcpc']['ceph']['pgp_auto_adjust'] = false
 # Need to review...
 default['bcpc']['ceph']['pgs_per_node'] = 1024
 default['bcpc']['ceph']['max_pgs_per_osd'] = 300
+default['bcpc']['ceph']['osd_scrub_load_threshold'] = 0.5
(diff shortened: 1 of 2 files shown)


erhudy:
Changed this PR to only make the scrub load threshold parameterizable without changing the default.

kelvk:
Do we want notifications to restart Ceph daemons?

kelvk:
... or inject the new setting into them live.

erhudy:
That's always been a difficult question because of Ceph's reliance on a single config file for everything and the possibility of Bad Things happening if Ceph daemons get rolled en masse. @kamidzi has brought it up elsewhere (e.g. #982) and it would be nice to have an actual solution to the problem. I was kind of hoping to be able to solve this problem with the concat providers but I don't think I was able to get notifications to work correctly.

kelvk:
I've added a comment to #982 . This is good to merge otherwise.

PR #885, opened 2015-10-30
Remove Apache frontend for RADOSgw

erhudy:
Since switching to Civetweb for RADOSgw, the Apache frontend has gone unused. This PR removes the Apache configuration entirely and makes a few other related changes to improve using RADOSgw, as part of an ongoing quest to simplify things slightly by removing unused features and code paths.

To test RADOSgw locally:

* Install **s3cmd** on your host.
* Set up `/etc/hosts` with two entries, one pointing `s3.bcpc.example.com` to 192.168.100.5, and one pointing `bucketname.s3.bcpc.example.com` also to 192.168.100.5, where **bucketname** is the name of the test bucket you will be creating:
```
192.168.100.5 s3.bcpc.example.com
192.168.100.5 erhudy.s3.bcpc.example.com
```
* Set up an s3cmd configuration file (default location is `$HOME/.s3cfg`). To get an access key, run `keystone ec2-credentials-list`, or `keystone ec2-credentials-create` if no credentials are listed here.
```
[default]
access_key = xxx
check_ssl_certificate = False
check_ssl_hostname = False
host_base = s3.bcpc.example.com
host_bucket = %(bucket)s.s3.bcpc.example.com
secret_key = yyy
use_https = True
```
* Verify `s3cmd ls` and `s3cmd la` do not return a 403 or other error (blank is fine).
* Create your testbucket with `s3cmd mb s3://bucketname`, using the name of the bucket you put into your hosts file.
* Put files in with `s3cmd put blah s3://bucketname`, and take them out with `s3cmd get s3://bucketname/blah`.


cookbooks/bcpc/attributes/default.rb
```diff
@@ -98,10 +93,6 @@
 #  Ceph settings for the cluster
 #
 ###########################################
-# Trusty is not available at this time for ceph-extras
-default['bcpc']['ceph']['extras']['dist'] = "precise"
```
erhudy: Not directly related to this change but not a repo we've needed in forever, so removed it.

cookbooks/bcpc/templates/default/ceph.conf.erb
```diff
@@ -84,7 +79,7 @@
 <% if config_defined('keystone-admin-token') %>
     rgw keystone url = openstack.<%=node['bcpc']['cluster_domain']%>:35358
     rgw keystone admin token = <%=get_config('keystone-admin-token')%>
-    rgw keystone accepted roles = admin Member _member_
+    rgw keystone accepted roles = Admin admin Member _member_
```
erhudy: Added capitalized "Admin" so that testing with the default admin user works correctly on local builds.

kamidzi:
I couldn't get the s3cmd command to work as documented above - but we can liaise about that offline. Though, is this intended to make retroactive changes as well; i.e. remove the currently running radosgw site? I believe it will still be active.

erhudy:
No, RADOSgw should continue to function and is supported by us in production.

erhudy:
Since we load-balance and terminate SSL for RADOSgw on HAProxy, the Apache site could be removed.

erhudy:
Unmentioned above: RADOSgw needs to be manually restarted on each node in order to pick up the configuration change that adds the **Admin** role to the list of allowed roles. Testing as the admin user that is created by default on local builds will fail with a 403 otherwise. Testing with a regular user should not be affected.

PR #886, opened 2015-11-02
Fixes to allow upgrading OpenStack in place to 2015.1.2

erhudy:
This can stand alone but can and should also be tested in conjunction with #875. I tested this both standalone and combined on local builds, and merged it with #875 when deploying on lab hardware to upgrade from the current production cluster configuration to the most recent Kilo patch release.

I wouldn't recommend merging it until we're ready to cut another major release.

### DIFF
--- a/cookbooks/bcpc/recipes/ceph-common.rb
+++ b/cookbooks/bcpc/recipes/ceph-common.rb
@@ -24,6 +24,14 @@ apt_repository "ceph" do
   distribution node['lsb']['codename']
   components ["main"]
   key "ceph-release.key"
+  notifies :run, "execute[apt-get update]", :immediately
+end
+
+# delete the compromised Ceph signing key (17ED316D)
+# if the new Ceph release key is installed (460F3994)
+bash "remove-old-ceph-key" do
+  code "apt-key del 17ED316D"
+  only_if "apt-key list | grep -q 460F3994 && apt-key list | grep -q 17ED316D"
 end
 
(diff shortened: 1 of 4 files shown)


cookbooks/bcpc/recipes/ceph-common.rb
```diff
@@ -43,15 +52,13 @@
 	EOH
 end
 
-# Installing CephFS but not activating it
-%w{librados2 librbd1 libcephfs1 python-ceph ceph ceph-common ceph-fs-common ceph-mds}.each do |pkg|
-    package pkg do
-        action :install
-        version node['bcpc']['ceph']['version']
-    end
+%w{librados2 librbd1 libcephfs1 python-ceph ceph ceph-common ceph-fs-common ceph-mds ceph-fuse}.each do |pkg|
```
erhudy: Added `ceph-fuse` here because it seems to be present on systems anyway and gets left behind at the old version if not explicitly called out.

erhudy:
Note: if merging with #875, there will be a merge conflict because both of them modify `ceph-common.rb` and it gets confused about the addition of the `ceph-fuse` package.

erhudy:
PR has been modified so that the old Ceph signing key will only be removed if the new one is installed. With this change, I believe this PR can be considered for merge immediately, because it does not force any breaking changes on existing production clusters.

erhudy:
This doesn't need to be tested in conjunction with #875 - that can be tested separately. This can stand on its own.

PR #887, opened 2015-11-03
Toggle to disable Heat

erhudy:
This PR adds a toggle to enable/disable Heat and turns Heat off by default, because we do not use it. With **bcpc.enabled.heat** set to **true**, this is a no-op. With it set to **false**, it will either not create anything for Heat, or will stop and disable Heat services if already installed (will not remove any existing Heat configuration or databases).


mihalis68:
Looks good to me. Using the ?w=1 tail on the URL for the diffs shows this is quite small when ignoring whitespace-only lines. Easier to grok that way.

PR #889, opened 2015-11-03
Make Glance Registry workers parameterized

erhudy:
Glance Registry escaped parameterization of its worker count, and so it starts a ginormous pile of workers on hardware nodes, each one of which scarfs a database connection. This PR updates it to derive its worker count from **bcpc.glance.workers** like Glance API.


kamidzi:
I don't believe glance registry is used? Is it?

erhudy:
Didn't analyze whether it's actively in use, just patching a few problem areas that caused excessive use of database connections.

PR #890, opened 2015-11-03
Do not automatically remove Ceph OSDs/mons

erhudy:
@mihalis68 and I had talked about this previously and agreed that this was undesirable behavior. A Chef run should not do anything destructive, particularly regarding storage systems. This PR removes the auto-reap behavior for both OSDs and monitors to avoid the possibility of something bad happening if Chef gets a bit weird.


mihalis68:
Looks good to me.

PR #891, opened 2015-11-03
Do not automatically remove RabbitMQ cluster nodes

erhudy:
Similar to #890, this removes the behavior that could cause a head node's RabbitMQ instance to be reaped automatically if something went wrong with node search.


mihalis68:
Looks good to me.

PR #892, opened 2015-11-03
Do not auto-reap Nova services when removed from Chef

erhudy:
I'm slightly more ambivalent about this one than #890 and #891, because removing this would mean we would need to do manual cleanup of Nova services if we undeploy hypervisors from a cluster (a more likely scenario than permanently removing head nodes). However, I do not like that it manually reaches into the database and deletes rows; the same can be accomplished via `nova service-delete` using IDs given by `nova service-list`. At most, this resource should be changed to do that instead of fiddling with the database.

### DIFF
--- a/cookbooks/bcpc/recipes/nova-head.rb
+++ b/cookbooks/bcpc/recipes/nova-head.rb
@@ -51,20 +51,5 @@ end
     end
 end
 
-ruby_block "reap-dead-servers-from-nova" do
-    block do
-        all_hosts = search_nodes("recipe", "nova-work").collect { |x| x['hostname'] }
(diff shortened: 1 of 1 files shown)


mihalis68:
I agree with this change either way. I'm ok with removing the reaping completely.

kamidzi:
I feel the service removal should just be a part of the procedure for undeploying/decommissioning a hypervisor.

erhudy:
Does that mean you feel it should be addressed in this resource, or should we put together an Ansible playbook instead to handle decommissioning?

kamidzi:
eegh... You know my feelings on ansible... That being said, I think that makes more sense as decom is more of an orchestration thing than configuration.

PR #893, opened 2015-11-04
Additonal bootstrap packages

kamidzi:
Can be useful on bootstrap node. Also changed annoying mode on getty.rb


erhudy:
Works as expected.

PR #898, opened 2015-11-06
Insert PTR records for hypervisor float/storage IPs

erhudy:
This adds PTR records for hypervisor storage/float IPs.


kelvk:
This looks to be missing the `100.0.10.in-addr.arpa` domain and corresponding NS/SOA records.

erhudy:
What are you seeing in `pdns.domains`? My local build turned up the results I expected.

kelvk:
<pre>
+----+--------------------------+--------+------------+--------+-----------------+---------+
| id | name                     | master | last_check | type   | notified_serial | account |
+----+--------------------------+--------+------------+--------+-----------------+---------+
|  1 | bcpc.example.com         | NULL   |       NULL | NATIVE |            NULL | NULL    |
|  2 | 100.168.192.in-addr.arpa | NULL   |       NULL | NATIVE |            NULL | NULL    |
|  3 | 127.1.in-addr.arpa       | NULL   |       NULL | NATIVE |            NULL | NULL    |
</pre>

erhudy:
Ah, I see. That does have the reverse domain for the floating IPs, and maybe the storage one just doesn't get created.

erhudy:
I haven't had time to fix this yet, but will circle back around on it this week and add the remaining missing PTR records.

erhudy:
Think it's fixed now and associates the PTR records with the correct domain. Need to verify that it chefs correctly onto an existing cluster.

PR #899, opened 2015-11-06
Limit loading of CloudCollector to head nodes

### LINKED ISSUE #896: Diamond CloudCollector should be disabled on monitoring cluster
On a monitoring cluster:

`[2015-11-05T16:46:03+00:00] ERROR: template[/etc/diamond/collectors/CloudCollector.conf] (bcpc::diamond line 133) had an error: RuntimeError: No config found for get_config(mysql-root-user)!!!`

kelvk:
Fixes #896

### DIFF
--- a/cookbooks/bcpc/recipes/diamond.rb
+++ b/cookbooks/bcpc/recipes/diamond.rb
@@ -142,6 +142,7 @@ if node['bcpc']['enabled']['metrics'] then
           )}
         })
         notifies :restart, "service[diamond]", :delayed
+        only_if "test -d /var/lib/mysql/nova"
     end
 
     service "diamond" do


kamidzi:
What about just testing for the role, like:
```
chef > puts 'do diamond' if node.roles.include?('BCPC-Headnode')
 => nil
chef > node.fqdn
 => "bcpc-vm4.hypervisor-bcpc.example.com"
```

kelvk:
I am trying to avoid making trips to the Chef server (I think that `node.roles` does) hence a local file test. I can apply your suggestion if this is the desired pattern.

kamidzi:
I believe L145 will still ensure that diamond starts up?

```
root@bcpc-vm4:~# sed -n 145p /var/chef/cache/cookbooks/bcpc/recipes/diamond.rb
        only_if "test -d /var/lib/mysql/nova"
root@bcpc-vm4:~# status diamond && service diamond stop && chef-client -obcpc::diamond &>/dev/null && status diamond
diamond start/running, process 26214
diamond stop/waiting
diamond start/running, process 29349
```

kelvk:
Yes, so does L148.

kamidzi:
Right, correction: I meant the block beginning on L148. I believe that would also need to be wrapped with a guard.

kelvk:
Do you mean Diamond should be disabled and stopped if `node['bcpc']['enabled']['metrics']` is false? If sio, I can see the usefulness of that during decommissioning, though it's probably better fit for separate issue.

kamidzi:
Well--I didn't even think of that--it would be a nice feature. I just mean in my testing, diamond still attempts to start on nodes that are `BCPC-Monitoring` and node `BCPC-Headnode` because of the service definition `start` action

kamidzi:
So, I misunderstood the purpose of this and conflated the service with the config. Looks good.

PR #900, opened 2015-11-06
Update nova-network Zabbix trigger

kelvk:
`nova-network` periodically forks a child process and that causes false alarms. When cheffed into an existing cluster, the old trigger needs to be removed manually via https://10.0.100.5/zabbix/triggers.php.


kamidzi:
As an aside, I was only able to disable the obsoleted triggers, not delete. Encountered:
```
Cannot delete templated trigger "Service Nova Network down on {HOST.NAME}:{BCPC-Headnode:proc.num[python,nova,,nova-network].last(0)}<>1".
```

kelvk:
You'd need to delete it from the parent template (BCPC-Worknode).

PR #911, opened 2015-11-17
Don't try to switch users on synchronize task

kamidzi:


erhudy:
On testing, it looks like setting `become: no` affects using **sudo** on both the local side that initiates the rsync, and on the capabilities of the rsync command on the other side, because I get permission failures when it tries to modify files owned by root on the remote host. There are some other commands that currently run as root that generate root-owned files that cause this to break. I will push another changeset on this PR to try to deal with those situations so rsync can run as non-root correctly.

erhudy:
View with `?w=1` to see the non-whitespace changes (fixed incorrect indentation in one file that was annoying me).

kamidzi:
I don't think it really makes sense to not be root and specify --perms to rsync. I've made a change to make the sudo rsync in preference to maybe triggering the condition I saw before.

erhudy:
I thought about that, but it's probably necessary to maintain the difference between 0644 and 0755 files so that scripts will execute as expected.

erhudy:
I'm fine with whatever, I didn't happen to encounter this particular problem.

erhudy:
After considering this further, I'm slightly hesitant to merge a kind of hacky playbook change in order to work around `PATH` being set differently after Ansible executes `sudo` (since that's presumably what's happening here). I would prefer to avoid circumventing Ansible's built-in privilege elevation mechanism.

kamidzi:
I'm not really sure it circumvents anything. Rather it makes it explicit that the rsync is going to be run under sudo. Can you reproduce this path thing?

erhudy:
The way it is right now is already explicit about rsync running with sudo, as the task inherits `become: yes` from the playbook. I never experienced the problem reported in the original issue.

erhudy:
Ansible 2.0 was just released and causes the synchronize tasks to break for a reason similar to what this PR is trying to solve. I am picking at it to find the Ansible-preferred way of handling the situation.

PR #915, opened 2015-11-18
Ignore DHCP clients we don't know on bootstrap node

erhudy:
This is a courtesy change to make us politer in a shared lab situation. `deny unknown-clients;` will prevent our DHCP server from handing out leases to clients it doesn't recognize.


kelvk:
The cobbler dhcp template was correctly changed but `cobbler sync` has to be invoked manually for the DHCP daemon to pick up the change. We should probably make this automatic.

erhudy:
Added a resource to run `cobbler sync` after changes to the DHCP template (and adjusted a few existing things to use the new resource).

PR #920, opened 2015-11-19
Don't apply flavors recipes on work nodes

erhudy:
This is a very small change to remove applying the **flavors** recipe on work nodes; head nodes already handle flavors, and this recipe can take a while to execute, so it will speed up Chef runs on work nodes.


mihalis68:
Looks good to me

PR #921, opened 2015-11-19
Don't hardcode the PNL mirror in the bootstrap sources.list template

erhudy:
derp

### DIFF
--- a/bootstrap/ansible_scripts/bootstrap_deployment/tasks-configure-package-serving.yml
+++ b/bootstrap/ansible_scripts/bootstrap_deployment/tasks-configure-package-serving.yml
@@ -10,6 +10,13 @@
     tags:
       - synchronize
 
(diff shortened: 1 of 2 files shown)


kamidzi:
What about using the indirect paths prefixed by `/bcpc/mirror-root`.  Just in case one is forced to migrate mirrors?

erhudy:
I'm okay with that, but it would require re-ordering some other tasks so that the symlinks are set up before the first `apt-get update`. This was the minimal-effort fix.

erhudy:
Force pushed over the old changeset with a new one that moves the symlink mapping up above the template creation, and replaces the paths in the template with ones derived from `{{ bootstrap_mirror_root_dir }}`.

kamidzi:
Tested local apt sources change with `ansible-playbook --tags configure_apt -v -i  <inventory-file> bootstrap_deployment/converge-bootstrap.yml`

PR #926, opened 2015-11-23
Add support for software RAID for ephemeral nodes

erhudy:
This improves ephemeral storage performance and removes dependency on hardware RAID striping by implementing an md stripe device to sit underneath the LVM volume group. It is disabled by default and will not attempt to begin managing things on existing ephemeral nodes.

NOTE: this is tagged **Do not merge** because monitoring has not yet been implemented. @kelvk is working out how to cleanly establish separate monitoring for ephemeral work nodes versus OSD work nodes. Once that is completed and merged, this PR will be updated with monitoring hooked into Zabbix. The feature works and can be tested.


roles/BCPC-EphemeralWorknode.json
```diff
@@ -1,20 +1,21 @@
...
-      "role[BCPC-Ephemeral]"
-    ],
-    "description": "A functional compute node in a BCPC cluster",
-    "chef_type": "role",
-    "override_attributes": {
-      "bcpc" : {
-          "aggregate_membership" : ["ephemeral_compute"],
-	  "nova" : {
-	      "ephemeral" : true,
-	      "ephemeral_disks": ["/dev/sdb", "/dev/sdc"]
```
erhudy: This was removed because it defaults to `/dev/md/md0` for all installations now.

erhudy:
Pending review and merge of #929. Once that's in, I will add monitoring to this PR.

Note for myself so I don't forget: the monitoring will need to distinguish systems using mdadm from systems still using hardware RAID, so there will need to be a little extra testing to not enable md device monitoring for matching systems (I think it will be sufficient to check whether **bcpc.software_raid.devices** is empty).

erhudy:
Monitoring has been added and this is ready for testing.

**NOTE**: if you are testing this on an existing local cluster, please update your node run lists to include the `BCPC-Hardware-Virtual` role, otherwise attempting to run Chef on the ephemeral work node will blow up when it tries to assemble the RAID device.

erhudy:
If you would like to simulate a drive failure, you can delete one of the block devices out from under the RAID 0 via `echo 1 > /sys/class/block/sd_/device/delete`. Doing so will cause I/O errors when attempting to manipulate logical volumes.

kelvk:
Works as expected when cheffed into an existing cluster. Will test with fresh build.

Issue #764, opened 2015-08-05
Update Zabbix templates for Ephemeral role


kelvk:
Testing in #759 uncovered (expectedly) absence of RGW check items on bcpc-vm3. Updating Zabbix templates may be in order.

PR #929, opened 2015-11-25
Zabbix templates rework and severity changes

kelvk:
Functionally, this creates a template for ephemeral work nodes to address #764. The `BCPC-Worknode` template is significantly reworked, so the below steps must be adhered to when deploying onto an existing cluster.

1. Remove `BCPC-Worknode` template (unlinking all elements)
2. Rechef cluster
3. Remove all `BCPC-Worknode` hosts
4. Wait for about [10 minutes](https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/attributes/default.rb#L1503) for the removed nodes to be rediscovered and associated with expected host groups / templates. You may wish to reduce discoery interval for `10.0.100.0/24` to 1 minute via https://10.0.100.6/zabbix/discoveryconf.php.
5. Wait about 10 minutes for https://10.0.100.6/zabbix/tr_status.php to show all _OK_ (except a Ceph Health Warn)

I have also bundled a change to reduce severity for some headnode triggers.

### DIFF
--- a/cookbooks/bcpc/files/default/zabbix_bcpc_templates.xml
+++ b/cookbooks/bcpc/files/default/zabbix_bcpc_templates.xml
@@ -1,13 +1,33 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <zabbix_export>
     <version>2.0</version>
-    <date>2015-08-07T15:44:19Z</date>
+    <date>2015-11-25T18:43:43Z</date>
     <groups>
         <group>
             <name>Templates</name>
         </group>
     </groups>
     <templates>
+        <template>
+            <template>BCPC-EphemeralWorknode</template>
+            <name>BCPC-EphemeralWorknode</name>
+            <description/>
+            <groups>
+                <group>
+                    <name>Templates</name>
+                </group>
+            </groups>
+            <applications/>
+            <items/>
+            <discovery_rules/>
+            <macros/>
+            <templates>
+                <template>
+                    <name>Template BCPC Compute</name>
+                </template>
+            </templates>
+            <screens/>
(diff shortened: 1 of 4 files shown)


kelvk:
It did occur to me that removal of the hosts will result in loss of historical data, but I see this as a minor annoyance because Zabbix triggers are based on very short-term conditions. This is somewhat implied with a short retention of [7 days](https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/attributes/default.rb#L1506).

erhudy:
Will test on an existing build.

erhudy:
Tested in the following way per above directions:
1. From **Configuration > Templates**, selected **BCPC-Worknode** and used action **Delete selected with linked elements.**
2. Recheffed entire cluster.
3. Deleted templateless work nodes from Zabbix (**bcpc-vm2** and **bcpc-vm3**).
4. Waited for re-discovery of the two deleted nodes.

bcpc-vm3 was rediscovered correctly, but bcpc-vm2 was not.

erhudy:
To be clear, by saying bcpc-vm2 was not rediscovered correctly, it never showed up at all in the hosts list after its deletion.

kelvk:
I can reproduce this. When removing `BCPC-Worknode` template in step 1, remove `BCPC-Worknode auto registration` action in https://10.0.100.6/zabbix/actionconf.php?eventsource=2 as well.

erhudy:
Re-testing with additional step for dealing with auto-registration.

erhudy:
After re-testing with the additional step, everything works as expected.

Issue #928, opened 2015-11-24
Unable to create VMs when a head node is unreachable


kelvk:
I have tested reproduction of this on virtual and physical nodes and seems fairly consistent.

1. Down the management interface on a non-VIP holding head node
<pre>ifdown em1</pre>
2. Test instance booting on another head node
<pre>check nova</pre> or <pre>nova --debug boot --flavor m1.tiny --image 'Cirros 0.3.4 x86_64 test1</pre>

The instance will hang in scheduling/building state.

erhudy:
I was able to reproduce. This appears to be at least partially due to https://github.com/rabbitmq/rabbitmq-server/issues/224 (found by @kelvk, I'm just linking it), and partially possibly due to a very long heartbeat interval (580 seconds prior to 3.5.5, we are running 3.5.4 everywhere).

erhudy:
As part of reproducing the issue, I examined stuck processes on the node that was leading on a particular queue (**conductor** in this case) and found that there are indeed wedged Erlang processes. The RabbitMQ server reports connections associated with these processes as being an **Unknown** connection, rather than the expected notation of of source and destination nodes/ports.

erhudy:
Upgrading to 3.5.6 helped with the stuck processes issue in RabbitMQ. Modifying sysctl variables and setting `net.ipv4.tcp_retries2` to 8 instead of the default of 15 helped with evicting dead clients from RabbitMQ much faster, allowing queues like **conductor** to return to normal. Unfortunately, Galera decided enough was enough and blew up in my face, bringing testing to an abrupt end, but we can probably get significant improvement in this regard by knocking TCP retries down and upgrading RabbitMQ.

kelvk:
Fixed

PR #930, opened 2015-11-30
Improve RabbitMQ resilience in the face of failures

erhudy:
This PR changes a few things about RabbitMQ's configuration, and one tunable in the kernel TCP/IP stack, in an attempt to ameliorate failure conditions described in #928. #928 describes a failure situation in which an unresponsive client can stop message processing within the cluster on account of not acknowledging messages on a queue (**conductor** being the most prominent example, though there are likely others). The explicit **heartbeat** setting in this PR and the downward adjustment of **net.ipv4.tcp_retries2** act to more quickly evict failed clients.

Since **net.ipv4.tcp_retries2** is a systemwide setting, please test thoroughly to evaluate for any unintended impacts elsewhere. In practice, since the change of this setting reduces TCP connection termination from ~924 seconds to ~100 seconds (still a very long time), I don't anticipate any actual impact from the change, but you never know.

In addition, this enables **pause_minority** partition handling in order to reduce operational burden after a cluster partition. It is possible that both sides of the cluster may still briefly evolve separately in the event of a network partition, but in testing **pause_minority** avoided Big Red Warnings when healing a cluster after a network partition.

**NOTE**: Versions of RabbitMQ prior to 3.5.5 have a separate bug that can result in failed clients never being evicted, per https://github.com/rabbitmq/rabbitmq-server/issues/224. The presence of this bug has been verified in 3.5.4, which is running on our current production clusters, and it can be determined whether there are stuck processes by running `rabbitmqctl eval 'rabbit_diagnostics:maybe_stuck().'`. Stuck processes will never evict and require the broker to be restarted in order to purge them; attempting to terminate the connection or underlying Erlang process will not succeed.

### DIFF
--- a/cookbooks/bcpc/templates/default/rabbitmq.config.erb
+++ b/cookbooks/bcpc/templates/default/rabbitmq.config.erb
@@ -7,5 +7,5 @@
 [
   {kernel,   [{inet_dist_use_interface, {<%=node['bcpc']['management']['ip'].gsub('.',',')%>}}]},
   {rabbitmq_management, [{listener, [{port, 55672}, {ip, "<%=node['bcpc']['management']['ip']%>"}]}, {redirect_old_port, false}]},
-  {rabbit, [{loopback_users, []}, {heartbeat, 60}, {cluster_partition_handling, pause_minority}, {tcp_listeners, [{"<%=node['bcpc']['management']['ip']%>", 5672}]}]}
+  {rabbit, [{loopback_users, []}, {heartbeat, <%=node['bcpc']['rabbitmq']['heartbeat']%>}, {cluster_partition_handling, pause_minority}, {tcp_listeners, [{"<%=node['bcpc']['management']['ip']%>", 5672}]}]}
 ].
--- a/cookbooks/bcpc/templates/default/rabbitmq.config.erb
+++ b/cookbooks/bcpc/templates/default/rabbitmq.config.erb
@@ -6,7 +6,6 @@
 
 [
   {kernel,   [{inet_dist_use_interface, {<%=node['bcpc']['management']['ip'].gsub('.',',')%>}}]},
-  {rabbitmq, [{tcp_listeners, [{"<%=node['bcpc']['management']['ip']%>", 5672}]}]},
   {rabbitmq_management, [{listener, [{port, 55672}, {ip, "<%=node['bcpc']['management']['ip']%>"}]}, {redirect_old_port, false}]},
-  {rabbit, [{loopback_users, []}]}
+  {rabbit, [{loopback_users, []}, {heartbeat, 60}, {cluster_partition_handling, pause_minority}, {tcp_listeners, [{"<%=node['bcpc']['management']['ip']%>", 5672}]}]}
 ].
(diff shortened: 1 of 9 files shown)


cookbooks/bcpc/templates/default/rabbitmq.config.erb
```diff
@@ -6,7 +6,6 @@
 
 [
   {kernel,   [{inet_dist_use_interface, {<%=node['bcpc']['management']['ip'].gsub('.',',')%>}}]},
-  {rabbitmq, [{tcp_listeners, [{"<%=node['bcpc']['management']['ip']%>", 5672}]}]},
```
erhudy: The `tcp_listeners` setting here was consolidated into the `rabbitmq` line below and should be a no-op from Rabbit's perspective (it was moving the setting to under `rabbit` anyway, per the output of `rabbitmqctl environment`).

kelvk:
When I tested 3.5.4 upgrade to 3.5.6 last week, I did not observe any issues. I will rechef this against an existing cluster that already has 3.5.6.

kelvk:
VM builds can still remain stuck as described in #928, notably if the configured primary AMQP broker in `rabbit_hosts` is dead. We should set `heartbeat_timeout_threshold` (http://docs.openstack.org/kilo/config-reference/content/configuring-rpc.html), or perhaps a system-wide equivalent, so Openstack can detect dead brokers quickly. Since applying `heartbeat_timeout_threshold=60`, VM builds have been consistently successful, albeit longer than usual (approximately 40-60 seconds instead of ~15 seconds) due to retries.

kelvk:
Pushed commit to enforce heartbeats on Openstack daemons. I ran the following to test:

1. `VBoxManage controlvm bcpc-vm2 poweroff`
2. `check nova` on remaining head nodes. They should complete successfully, despite taking longer than usual.

erhudy:
Will check it out.

erhudy:
After powering off a head node and running `check nova` on the first head node, it never succeeded. I gave it about 3 minutes, but **conductor** didn't have any consumers at all, so I gave up. I retried again and it succeeded after 97 seconds. I then tried on the other head node and it succeeded after about 30 seconds there. There is definitely still going to be a gap during which we can lose messages during a head node failure; it looks like it takes the cluster a couple of minutes after a failure to really stabilize.

I wonder if combining these changes with your changes to prefer the local broker would further improve resiliency?

erhudy:
After restarting all services on remaining live head nodes, experienced a check failure that cleared up on the next run. It still appears that things are not as good as they could be, so I'm going to re-test with a shuffled node list that prefers the local node per @kelvk's earlier experiments.
```
+------+-------+--------+--------------------------------------------------------------------------------------------------------------+
| name | state |  time  |                                                     msg                                                      |
+------+-------+--------+--------------------------------------------------------------------------------------------------------------+
| nova | ERROR | 17.290 | <class 'novaclient.exceptions.NotFound'> : Instance c0e30a3f-dcfa-44d5-bfb0-9ee1e920a1dc could not be found. |
+------+-------+--------+--------------------------------------------------------------------------------------------------------------+
```

erhudy:
Added a changeset that will rewrite the RabbitMQ server lists to prefer the local host first.

erhudy:
The general summary of this PR based on the testing @kelvk and I have done is that it these changes are not silver bullets that will make a failed head node unnoticeable from a user perspective, but will hopefully improve certain aspects of the failure modes in areas we can control.

Issue #908, opened 2015-11-13
routemon no longer needed


mihalis68:
The routemon service in chef-bcpc tries to spot and fix default routes going away from the various bcpc routing tables. This was necessary when we were not correctly marking such routes as static, however due to https://github.com/bloomberg/chef-bcpc/pull/718 that is now fixed. Remove the routemon stuff entirely on new cluster builds.

kamidzi:
Closed via 24496b5

Issue #894, opened 2015-11-04
Disable keystone for real


kelvk:
Disable the separate keystone process for real as accidental `start keystone` can result in numerous failed respawns.

kamidzi:
I suggest we just gzip/move away `/etc/init/keystone.conf`

PR #931, opened 2015-11-30
Purge deprecated keystone upstart job, routemon from checks

kamidzi:
Small fix; responds to #908, and changes in #894


erhudy:
Tested and works as expected.

PR #932, opened 2015-12-01
cluster inventory arbitrary grouping

kamidzi:
I found myself wanting to group nodes by rack when generating inventory files from the cluster.yml. This is a first pass to allow inventory grouping criteria expressible as a regex.

### DIFF
--- a/bootstrap/ansible_scripts/scripts/cluster_yaml_to_inventory.py
+++ b/bootstrap/ansible_scripts/scripts/cluster_yaml_to_inventory.py
@@ -9,6 +9,7 @@ import jinja2
 import os
 import sys
 import yaml
+import re
 
 
 INVENTORY_TEMPLATE = """[localhost]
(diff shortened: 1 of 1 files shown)


bootstrap/ansible_scripts/scripts/cluster_yaml_to_inventory.py
```diff
@@ -133,21 +221,41 @@ def render_inventory(path, ssh_user):
...
-        description='Takes cluster.yaml and emits an Ansible inventory.')
+        description='Takes cluster.yaml and emits an Ansible inventory.',
+        formatter_class=argparse.RawTextHelpFormatter)
     parser.add_argument('path', help='path to cluster.yaml')
+    parser.add_argument('-G', '--grouping',
+        help='''set the grouping options
+
+If appears multiple times, last occurring value takes precedence
+The format is as follows:
+<label>:prefix=<some_prefix>,host_pattern=<some_regex>,grouping_key=<str>
```
erhudy: Can you adjust this line or the example line so that the order matches? This confused me for a few minutes until I realized **prefix** and **grouping_key** were transposed between this line and the example.

erhudy:
Seems to work. I wasn't able to trigger the nice failure path if `grouping_key` wasn't found in `host_pattern`. It blew up on me with a problem in string formatting:
```
➜  scripts git:(c2ef130) ./cluster_yaml_to_inventory.py -G test:grouping_key=rack,host_pattern='fake-node-r(?P<notrack>\d+)n\d+',prefix='rack-' ~/Desktop/cluster.yml
  File "./cluster_yaml_to_inventory.py", line 277, in <module>
    main()
  File "./cluster_yaml_to_inventory.py", line 273, in main
    print(render_inventory(args.path, args.ssh_user, args.__dict__))
  File "./cluster_yaml_to_inventory.py", line 173, in render_inventory
    g_opts = parse_grouping_opts(opts['grouping'])
  File "./cluster_yaml_to_inventory.py", line 168, in parse_grouping_opts
    'given by host_pattern' % opts[set_label]['grouping_key'])
TypeError: not all arguments converted during string formatting
```
I was able to fix the problem and will push a changeset.

erhudy:
I shed a tear for the loss of PEP-8 compliance, but have no good ideas for how to reformat the command-line example of how to use `--grouping`.

erhudy:
Do you want to squash before merge?

kamidzi:
K. squashed.

PR #933, opened 2015-12-01
Extract configfile loading to common functions

kamidzi:
This should allow running `shared_prereqs.sh` independently when behind a proxy.

### DIFF
--- a/bootstrap/shared/shared_functions.sh
+++ b/bootstrap/shared/shared_functions.sh
@@ -28,3 +28,28 @@ check_for_envvars() {
   done
   if [[ $FAILED_ENVVAR_CHECK != 0 ]]; then exit 1; fi
 }
+
+load_configs(){
+  # Source the bootstrap configuration defaults and overrides.
+  # If the overrides file is at the previous expected location of bootstrap.sh,
+  # move it and notify the user.
+  if [[ -f "$REPO_ROOT/bootstrap/config/bootstrap_config.sh" ]]; then
+    if [[ ! -f "$REPO_ROOT/bootstrap/config/bootstrap_config.sh.overrides" ]]; then
+      echo "Performing one-time move of bootstrap_config.sh to bootstrap_config.sh.overrides..."
+      mv $REPO_ROOT/bootstrap/config/bootstrap_config.sh $REPO_ROOT/bootstrap/config/bootstrap_config.sh.overrides
+    else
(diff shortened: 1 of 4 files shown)


erhudy:
The changes seem to now expect that `BOOTSTRAP_HTTP_PROXY` and `BOOTSTRAP_HTTPS_PROXY` will always be set. When running on a system not behind proxy servers which do not have these variables configured:
```
BCPC Vagrant BootstrapV2 0.2
Environment variable BOOTSTRAP_HTTP_PROXY must be set!
Environment variable BOOTSTRAP_HTTPS_PROXY must be set!
Environment variable BOOTSTRAP_HTTPS_PROXY must be set!
```

kamidzi:
Moved the guard, should fix that issue. Though is someone were to want to add another step to the bootstrap process before the prereq fetching that requires proxy configuration, she'd have to be...aware.  Force-pushed the change.

erhudy:
Is the goal that `shared_prereqs.sh` should be executable as a standalone script without being called from `BOOT_GO.sh`? If I try to run `shared_prereqs.sh` by itself (e.g. via `bootstrap/shared/shared_prereqs.sh`), it fails thusly:
```
➜  chef-bcpc git:(f609d15) bootstrap/shared/shared_prereqs.sh 
bootstrap/shared/shared_prereqs.sh: line 11: check_for_envvars: command not found
```

erhudy:
To be clear, I don't think that's a regression; there's always been ordering issues with these scripts and making them standalone, hence wrapping it all in `BOOT_GO.sh` to handle those issues. Since `shared_prereqs.sh` is exceptional among these scripts in that it has a valid reason to be executed standalone, maybe we should intentionally duplicate any functions into it as needed so that it can run by itself?

kamidzi:
Yeah, that may be a more reasonable approach. I'll re-work this in another PR

kamidzi:
Actually, come to think of it - My original intention was to simply make the invocation work as documented in the anisble build guides, for example https://github.com/bloomberg/chef-bcpc/blob/master/docs/ansible_local_build_guide.md

kamidzi:
How about we just merge just as a temporary measure, since this behavior is already documented?

erhudy:
Fine with me, it's not breaking anything that isn't already broken. I will verify.

erhudy:
Trashed `~/.bcpc-cache` and re-ran everything via `BOOT_GO.sh` on both a system with direct Internet access and a system using a proxy server. Everything seems to work as expected.

PR #935, opened 2015-12-04
Don't assume the local server's in the list of servers passed in

### LINKED ISSUE #934: Initial head node convergence fails after #930 was merged
Just hit this error when building from master:
```
Chef::Mixin::Template::TemplateError (no implicit conversion from nil to integer) on line #135:

133: # Rabbit message queue settings
134: #rabbit_host=<%=node['bcpc']['management']['vip']%>
135: rabbit_hosts=<%=get_shuffled_servers(@servers, prefer_local=true).map{|x| x['bcpc']['management']['ip'] + ":5672"}.join(',')%>
```

erhudy:
Fixes #934. My logic was faulty and assumed that the local server would always be in the list passed in; if it's not, `Array#index` returns `nil`, which then gets inserted at the head of the list and causes Ruby to detonate when it tries to concatenate `nil` and `:5672` in the template.


kelvk:
Tested successfully on a clean build with 3 heads and 1 work.

PR #938, opened 2015-12-08
Fixes #937: disables mod_python

### LINKED ISSUE #937: Keystone WSGI workers may crash
From time to time Keystone WSGI workers may terminate with the error message `Fatal Python error: PyEval_AcquireThread: NULL new thread state`. This appears to be caused by a bad interaction between mod_wsgi and mod_python. mod_python is not needed and can be safely disabled via `a2dismod python && service apache2 restart`.

erhudy:


kamidzi:
I don't think php5 is unused. @kelvk? Building with monitoring:
```
 * The apache2 configtest failed.
    STDERR: Output of config test was:
 
   AH00526: Syntax error on line 17 of /etc/apache2/conf-enabled/zabbix.conf:
    Invalid command 'php_value', perhaps misspelled or defined by a module not included in the server configuration
    Action 'configtest' failed.
```

kamidzi:
Additionally, just patching the zabbix template seems insufficient since apache2 runs before zabbix recipe. I had attempted the following half-measure:
```
From ed834d39ba701aa16f0d9323af4519f1c5252287 Mon Sep 17 00:00:00 2001
Date: Tue, 8 Dec 2015 12:57:07 -0500
Subject: [PATCH] Add IfModule guard for php_value settings

---
 cookbooks/bcpc/templates/default/apache-zabbix-global.conf.erb | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/cookbooks/bcpc/templates/default/apache-zabbix-global.conf.erb b/cookbooks/bcpc/templates/default/apache-zabbix-global.con
index 51d3458..57f8eb2 100644
--- a/cookbooks/bcpc/templates/default/apache-zabbix-global.conf.erb
+++ b/cookbooks/bcpc/templates/default/apache-zabbix-global.conf.erb
@@ -14,9 +14,11 @@
     Order allow,deny
     Allow from all

+    <IfModule mod_php5.c>
     <% @php_settings.each do |key, value| %>
     php_value <%= key %> <%= value %>
     <% end %>
+    </IfModule>
 </Directory>

 <Directory "/usr/share/zabbix/conf">
--
```

kelvk:
php is needed on monitoring, but should not be on head nodes.

erhudy:
Yeah, I forgot about Zabbix requiring PHP. :(

erhudy:
Force smashed the branch with a replacement changeset that only deals with mod_python.

PR #946, opened 2015-12-11
nova and cinder quota management

kelvk:
Manage quotas via Chef. The use of a generated quota configuration file is to avoid trips to the database to query/update if nothing has changed, since `set-os-quota.py` only runs when notified. The default nova cores limit for `AdminTenant` is deliberately set to unlimited as `check nova` on multi-head clusters may hit the default cores limit of `4`.

Testing:

1. Set the limits via Chef environment, i.e.
<pre>
    "bcpc": {
      "quota": {
        "nova": {
          "AdminTenant": {
            "security_groups": 99,
            "floating_ips": 99
          },
          "AnotherTenant": {
            "cores": 99
          }
        },
        "cinder": {
          "AdminTenant": {
            "gigabytes": 99
          }
        }
      },
</pre>
2. Rechef. Quota update will run for each tenant that has its limits changed.


cookbooks/bcpc/files/default/set-os-quota.py
```diff
@@ -0,0 +1,87 @@
...
+    self.db = MySQLdb.connect(read_default_file=db_conf)
+    self.config = config
+
+  """
+  Retrieve project/tenant UUID
+  """
+  def _get_tenant_id(self, project):
+    uuid_re = re.compile('[0-9a-f]{32}\Z', re.I)
+    c = self.db.cursor()
+    c.execute("SELECT id FROM keystone.project WHERE name = %s", project)
```
erhudy: I have found that MySQLdb sometimes does weird things when supplying single arguments; certain versions will allow this syntax, other ones freaked out and required it to be in a tuple, e.g. `(project,)`. The tuple syntax worked everywhere, so I updated all of my code to use tuples even in the case of single arguments.

cookbooks/bcpc/files/default/set-os-quota.py
```diff
@@ -0,0 +1,87 @@
...
+    else:
+      return None
+
+  """
+  Retrieve list of non-default quotas from database for the project
+  """
+  def _get_current_quota(self, component, tenant_id):
+    self.db.select_db(component)
+    c = self.db.cursor()
+    c.execute("SELECT resource, hard_limit FROM quotas WHERE project_id = %s AND deleted = 0", tenant_id)
```
erhudy: Same comment as above about tuples.

cookbooks/bcpc/files/default/set-os-quota.py
```diff
@@ -0,0 +1,87 @@
...
+    self.config = config
+
+  """
+  Retrieve project/tenant UUID
+  """
+  def _get_tenant_id(self, project):
+    uuid_re = re.compile('[0-9a-f]{32}\Z', re.I)
+    c = self.db.cursor()
+    c.execute("SELECT id FROM keystone.project WHERE name = %s", project)
+    tenant_id = c.fetchone()[0]
```
erhudy: You may want to test `tenant_id is not None` before indexing into it, in the event the project cannot be found.

erhudy:
Seems to work in testing. Created a tenant, assigned Cinder and Nova quotas to it, verified that they were set appropriately. Reduced quotas and recheffed, and they were reduced as expected.

erhudy:
I would like to suggest adding a filter somewhere (probably in `_construct_quota_cmd`) that will explode if you try to pass in an invalid/unrecognized item to change the quota for. Something like:
```
for (resource, limit) in quota:
    if component == 'nova':
        if resource not in ['instances', 'cores', 'ram', 'floating_ips', ...]:
            raise Exception('kaboom')
    elif component == 'cinder':
        if resource not in ['gigabytes', 'volumes', 'snapshots']:
            raise Exception('kaboom')
    else:
        raise Exception("unrecognized component %s" % component)
```

erhudy:
One other thought: what should we expect to happen if we attempt to reduce a quota beyond what is already in use? `nova quota-update` has a `--force` flag for this kind of situation. Do we want to force the update to happen? It doesn't look like `cinder quota-update` has the same flag.

kelvk:
If we are doing this via Chef, we should also support such reductions. I will add `--force` for nova quota updates. Cinder by [default](https://bugs.launchpad.net/neutron/+bug/1304234) forces such reductions, though normalized command line options would be nice.

kelvk:
Pushed new update to include suggested changes and logging.

erhudy:
Retesting.

erhudy:
Got failure as expected when specifying an invalid resource:
```
    STDERR: Traceback (most recent call last):
      File "/usr/local/bin/set-os-quota.py", line 118, in <module>
        quota = OSQuota(config).run()
      File "/usr/local/bin/set-os-quota.py", line 106, in run
        quota_cmd = self._construct_quota_cmd(component, configured_quota, tenant_id)
      File "/usr/local/bin/set-os-quota.py", line 82, in _construct_quota_cmd
        raise Exception("No such cinder resource: %s", resource)
    Exception: ('No such cinder resource: %s', 'gigablorts')
```

cookbooks/bcpc/files/default/set-os-quota.py
```diff
@@ -0,0 +1,118 @@
...
+          current_quota = self._get_current_quota(component, tenant_id)
+          configured_quota = self._parse_configured_quota(config[component][project])
+          """ If defined quota does not match in-effect quota, apply update """
+          if cmp(current_quota, configured_quota) != 0:
+            quota_cmd = self._construct_quota_cmd(component, configured_quota, tenant_id)
+            process = subprocess.Popen(quota_cmd)
+            process.communicate()
+            if process.returncode != 0:
+              syslog.syslog(syslog.LOG_ERR, "Unable to update quota for %s: %s" %(project, str(quota_cmd)))
+              raise Exception('Unable to update quota: %s', quota_cmd)
```
erhudy: This should be `raise Exception('Unable to update quota: %s' % quota_cmd)`.

cookbooks/bcpc/files/default/set-os-quota.py
```diff
@@ -0,0 +1,118 @@
...
+                            'server_groups', 'server_group_members']:
+          syslog.syslog(syslog.LOG_ERR, "No such nova resource: %s" %resource)
+          raise Exception("No such nova resource: %s", resource)
+      elif component == 'cinder':
+        if resource not in ['volumes', 'snapshots', 'gigabytes']:
+          syslog.syslog(syslog.LOG_ERR, "No such cinder resource: %s" %resource)
+          raise Exception("No such cinder resource: %s", resource)
+      else:
+        syslog.syslog(syslog.LOG_ERR, "Unrecognized component %s" %component)
+        raise Exception("Unrecognized component %s", component)
```
erhudy: This should be `raise Exception("Unrecognized component %s" % component)`.

cookbooks/bcpc/files/default/set-os-quota.py
```diff
@@ -0,0 +1,118 @@
...
+        if resource not in ['instances', 'cores', 'ram', 'floating_ips', 'fixed_ips', 'metadata_items',
+                            'injected_files', 'injected_file_content_bytes', 'injected_file_path_bytes',
+                            'key_pairs', 'security_groups', 'security_group_rules',
+                            'server_groups', 'server_group_members']:
+          syslog.syslog(syslog.LOG_ERR, "No such nova resource: %s" %resource)
+          raise Exception("No such nova resource: %s", resource)
+      elif component == 'cinder':
+        if resource not in ['volumes', 'snapshots', 'gigabytes']:
+          syslog.syslog(syslog.LOG_ERR, "No such cinder resource: %s" %resource)
+          raise Exception("No such cinder resource: %s", resource)
```
erhudy: This should be `raise Exception("No such cinder resource: %s" % resource)`.

cookbooks/bcpc/files/default/set-os-quota.py
```diff
@@ -0,0 +1,118 @@
...
+    quota_cmd = [component, 'quota-update']
+    """ Validate provided component and resources for quota-update """
+    for (resource, limit) in quota:
+      if component == 'nova':
+        if resource not in ['instances', 'cores', 'ram', 'floating_ips', 'fixed_ips', 'metadata_items',
+                            'injected_files', 'injected_file_content_bytes', 'injected_file_path_bytes',
+                            'key_pairs', 'security_groups', 'security_group_rules',
+                            'server_groups', 'server_group_members']:
+          syslog.syslog(syslog.LOG_ERR, "No such nova resource: %s" %resource)
+          raise Exception("No such nova resource: %s", resource)
```
erhudy: This should be `raise Exception("No such nova resource: %s" % resource)`.

erhudy:
Syslog updated as expected when recheffing with valid changes:
```
Dec 14 18:28:36 bcpc-vm1 set-os-quota.py: Updated cinder quota for quota_test: (('gigabytes', 1800),)
Dec 14 18:28:37 bcpc-vm1 set-os-quota.py: Updated nova quota for quota_test: (('cores', 16), ('ram', 2048))
```

kelvk:
Exceptions corrected

erhudy:
Tested cheffing with a non-existent project:
```
Dec 14 18:50:30 bcpc-vm1 set-os-quota.py: Non-existent project defined: tenant_doesnt_exist
```
Also tested reducing quotas for both Nova and Cinder below existing usage and verified that both took hold.

I think this is good to merge with the fixes to the exception message formatting. If possible, it would be good to reformat it in according with PEP-8, but if there isn't time to do so I will merge as-is.

kelvk:
Updated to align with PEP-8 guidelines. Thanks for your patience!

erhudy:
Retesting with updated script and will merge if everything looks good.

PR #951, opened 2015-12-22
Make sure to check for bootstrap-files/<chef_bcpc_version> ...

kamidzi:
even with prebuilts>. I claim this check is necessary always, though there could be something off with my procedure...

### DIFF
--- a/bootstrap/ansible_scripts/common_playbooks/validate_environment.yml
+++ b/bootstrap/ansible_scripts/common_playbooks/validate_environment.yml
@@ -52,12 +52,11 @@
     register: files_build_resources_dir_stat
     become: no
     run_once: true
-    when: not use_prebuilt_binaries
     tags:
       - always
 
   - fail: msg="Directory {{ controlnode_files_dir }}/{{ chef_bcpc_version }} does not exist"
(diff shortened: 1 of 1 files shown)


erhudy:
Yeah, it is necessary in order to source the Chef packages (maybe other things too, but Chef is the one I know of). I thought I had fixed the playbook accordingly, but apparently not.

erhudy:
This takes the same action as the `check for presence of local misc files directory for building binaries on bootstrap node (versioned)` resource below it, so I think the same could be achieved by removing the **when** clause from there. I'm not sure what happened to the other task that's supposed to fail the playbook depending on the value of the registered variable, but the task you added could be edited to
```
  - fail: msg="Directory {{ controlnode_files_dir }}/{{ chef_bcpc_version }} does not exist"
    when: not files_build_resources_dir_stat.stat.exists
    tags:
      - always
```
...and then everything works as expected.

kamidzi:
Yeah, I saw that and wasn't sure what it did, so I didn't change it. I'll modify accordingly

erhudy:
I bet I probably screwed up and "fixed" it initially by removing the task that would fail the playbook, but didn't remove the when clause from the register task as well. That would explain where the failure task went.

kamidzi:
Verified and force-pushed the change.

Issue #953, opened 2015-12-22
Upgrade chef-client to at least 12.5.1


kamidzi:
It's about time. Plus, this update http://www.rubydoc.info/gems/chef/12.5.1/Chef/Mixin/Template/TemplateContext is useful for #939

kamidzi:
Upgrading to anything 12.4.0+ (tested 12.4.0, 12.5.1, 12.6.0) seems to trigger the following:

```
* directory[/var/run/ceph/guests/] action create
    * cannot determine user id for 'libvirt-qemu', does the user exist on this system?
    ================================================================================
    Error executing action `create` on resource 'directory[/var/run/ceph/guests/]'
    ================================================================================

    Chef::Exceptions::UserIDNotFound
    --------------------------------
    cannot determine user id for 'libvirt-qemu', does the user exist on this system?

    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/bcpc/recipes/ceph-common.rb

     92: directory "/var/run/ceph/guests/" do
     93:   owner "libvirt-qemu"
     94:   group "libvirtd"
     95:   mode  "0755"
     96: end
     97:
```

kamidzi:
FWIW, seem likes the combo of chef-server 12.3.0 / chef-client 12.5.1 will take

kamidzi:
Additionally, chef-server 12.3.0 plays nicely with chef-client 12.3.0+, though it would probably make little sense to run a client < 12.5.1, as that is the version of the internal client shipping with chef-server 12.3.0

erhudy:
Resolved by #1008.

PR #961, opened 2015-12-28
Allow specification of chef versions

kamidzi:
Corollary to unfiled fix for #953

### DIFF
--- a/bootstrap/shared/shared_configure_chef.sh
+++ b/bootstrap/shared/shared_configure_chef.sh
@@ -23,13 +23,26 @@ KNIFE=/opt/opscode/embedded/bin/knife
 
 # install and configure Chef Server 12 and Chef 12 client on the bootstrap node
 # move nginx insecure to 4000/TCP is so that Cobbler can run on the regular 80/TCP
-do_on_node bootstrap "sudo dpkg -i \$(find $FILECACHE_MOUNT_POINT/ -name chef-server\*deb -not -name \*downloaded | tail -1) \
+if [[ -n "$CHEF_SERVER_DEB" ]]; then
+    debpath="$FILECACHE_MOUNT_POINT/$CHEF_SERVER_DEB"
+    CHEF_SERVER_INSTALL_CMD="sudo dpkg -i $debpath"
+else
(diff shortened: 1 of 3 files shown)


bootstrap/shared/shared_configure_chef.sh
```diff
@@ -70,7 +83,11 @@ do_on_node bootstrap "$KNIFE cookbook upload -a \
 # install and bootstrap Chef on cluster nodes
 i=1
 for vm in $vms $mon_vms; do
-  do_on_node $vm "sudo dpkg -i \$(find $FILECACHE_MOUNT_POINT/ -name chef_\*deb -not -name \*downloaded | tail -1)"
+  # Try to install a specific version, or just the latest
+  if [[ -z "$CHEF_CLIENT_DEB" ]]; then
+    echo "Installing latest chef-client found in $vm:$FILECACHE_MOUNT_POINT"
+  fi
+  do_on_node bootstrap "$CHEF_CLIENT_INSTALL_CMD"
```
erhudy: This needs to be `do_on_node $vm` rather than `do_on_node bootstrap`.

erhudy:
After changing the `do_on_node bootstrap` line marked above to `do_on_node $vm` (so that the Chef client is upgraded on each VM instead of the unintentional effect of upgrading the client repeatedly on the bootstrap node), this works as expected.

erhudy:
Re-checking and will merge on success.

erhudy:
One observation before merge: the if statement at https://github.com/bloomberg/chef-bcpc/pull/961/files#diff-d37421a6d1f4a8b5352df28c5d878870R87 never seems to be executed, even if you don't set `CHEF_CLIENT_DEB` anywhere. I think this behavior is still fine because you will get the default version of Chef that we want you to use if you are not explicitly specifying one, just seems slightly different than intended.

kamidzi:
I didn't run into that; the idea is that it is initialized at https://github.com/bloomberg/chef-bcpc/blob/f9a36f9e94ef3e99da2c55cd7bab19f65c551113/bootstrap/config/bootstrap_config.sh.defaults#L76. We may want to fix this if that's the case.

PR #962, opened 2015-12-29
Make swap size configurable for VMs

### LINKED ISSUE #851: Extract SWAP_SIZE for created vms
This really should have been included as part of #849, for completeness...
See https://github.com/bloomberg/chef-bcpc/blob/master/bootstrap/vagrant_scripts/Vagrantfile#L159

kamidzi:
- Closes #851.
 - Also adds config dumping util.

Also fixes a subtle bug with re-provisions. If the swapfile size has been decreased, the configured swap would not shrink correspondingly as that larger quantity would have already been `fallocate`'d; hence the call to `truncate`. A good testing command (for re-provisions) is:

```
 vagrant provision vm1 --provision-with add-swap-space && vagrant ssh vm1 -c 'ls -l --si /swap; cat /etc/fstab ; cat /proc/swaps'
```

It can also be seen that if only `truncate` is used, then growing the swap file will also not work as the file will be sparse and the provisioner will complain:

```
==> vm1: Running provisioner: add-swap-space (shell)...
    vm1: Running: inline script
==> vm1: Setting up swapspace version 1, size = 450556 KiB
==> vm1: no label, UUID=37d5b221-5486-40e1-a799-0a19468c9af3
==> vm1: swapon:
==> vm1: /swap: skipping - it appears to have holes.
```

### DIFF
--- a/bootstrap/config/bootstrap_config.sh.defaults
+++ b/bootstrap/config/bootstrap_config.sh.defaults
@@ -66,6 +66,9 @@ export BOOTSTRAP_APT_MIRROR=
 # Valid range is 0-3 (default 0).
 export MONITORING_NODES=0
 
+# Sets the size of the swapfile allocated (in MiB) to created VMs
+export VM_SWAP_SIZE=8192
+
 # These directories are where the repository and file cache will be mounted via
 # shared folders inside each VM.
 export REPO_MOUNT_POINT=/chef-bcpc-host
--- a/bootstrap/vagrant_scripts/Vagrantfile
+++ b/bootstrap/vagrant_scripts/Vagrantfile
@@ -152,17 +152,19 @@ EOH
 $add_swap_script = <<-EOH
   #!/bin/bash
   swap_path='/swap'
+  swap_size="#{ENV['VM_SWAP_SIZE'] ||= 8192}M"
   # This could be a re-provision
-  if grep -w "^${swap_path}" /proc/swaps ; then
-    swapoff "${swap_path}"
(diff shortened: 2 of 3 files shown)


erhudy:
The swap part works as expected, was able to run `vagrant provision` to grow and shrink the swap file.

Since `dump_config.sh` was added, I also checked it out and found that since it purges the environment before running, any settings with interpolated variables don't appear correctly in the output (e.g., the default setting for `BOOTSTRAP_CACHE_DIR` is `$HOME/.bcpc-cache`, which results in the output `BOOTSTRAP_CACHE_DIR=/.bcpc-cache`). Specifically whitelisting `$HOME` to be passed in and then excluding it from the output resulted in the expected output. Naturally, we can't know all the environment variables that might require interpolation, but allowing `$HOME` through does result in the right output for the default configuration:
```
env -i HOME="$HOME" /bin/bash -c "export -n PWD SHLVL  ; source $defaults && source $override && printenv | \
    sed '/^_=/d' | grep -v ^HOME | sort"
```
Maybe in a future update we could namespace all variables to begin with `BCPC_BOOTSTRAP_` and then it wouldn't be necessary to purge the environment or whitelist particular environment variables; instead we could simply run the command to import the defaults and overrides in the bash subshell and then grep for `^BCPC_BOOTSTRAP_`. What do you think?

kamidzi:
I've added $HOME for now, but yes we should really namespace the envvars in a future effort

erhudy:
Re-tested provisioning and `dump_config.sh`.

PR #963, opened 2016-01-06
Route VNC traffic through Haproxy to prevent SSL certificate issues

semenovroman:
VNC traffic goes through Haproxy that does SSL termination with proper cert (hence no ssl warning in horizon).

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -177,7 +177,7 @@ default['bcpc']['management']['interface-parent'] = nil
 # list of TCP ports that should be open on the management interface
 # (generally stuff served via HAProxy)
 default['bcpc']['management']['firewall_tcp_ports'] = [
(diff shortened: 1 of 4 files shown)


erhudy:
Should we move the listener port for HAProxy to something other than 6080/TCP, in the event that we need to look at the console of an instance that's on a head node? In production we don't allow instances to schedule on head nodes other than the check instance, but it's a scenario that comes up often in testing due to resource constraints on our local systems.

semenovroman:
There shouldn't be any issues with connecting to vm scheduled on headnode through VNC. Haproxy is listening on vip, vnc-proxy is listening on normal ip:
```
ubuntu@ll21-bcpc-r1n3:~$ sudo netstat -ltpn | grep 6080
tcp        0      0 192.168.194.46:6080     0.0.0.0:*               LISTEN      3507905/python  
tcp        0      0 192.168.194.206:6080    0.0.0.0:*               LISTEN      3503171/haproxy
```

erhudy:
Right, someday I'll actually remember what bind addresses are.

PR #966, opened 2016-01-12
updates to address Ansible 2.0 issues/deprecation warnings

erhudy:
(all should be backwards compatible with Ansible 1.9.4)

This should also address the issue originally raised in #911 without needing to override the rsync command to `sudo rsync`, by setting `become: no` and enforcing permissions on the remote end. Note that it is not intended that these playbooks be run as root.

### DIFF
--- a/bootstrap/ansible_scripts/bootstrap_deployment/tasks-configure-chef-server.yml
+++ b/bootstrap/ansible_scripts/bootstrap_deployment/tasks-configure-chef-server.yml
@@ -29,7 +29,7 @@
       - chef
 
   - name: Run Chef Server configuration
-    sudo: yes
+    become: yes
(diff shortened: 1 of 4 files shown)


erhudy:
Rebased just now to fix merge conflicts with master.

kelvk:
The bootstrap directories are created with intended user/group ownership, but `synchronize` subsequently overwrites them with the source's. Pushed commit to ignore source ownerships.

Bootstrap node convergence with 1.9.4 and 2.0.1.0 seems to work fine. @erhudy, any comments before I merge?

erhudy:
As long as it doesn't make rsync get stupid, that's fine. The ultimate goal is that the rsync'd files on the destination should all be owned by **operations** instead of existing with nonexistent UID/GIDs.

PR #968, opened 2016-01-14
Moving Chef playbook contents to task lists

erhudy:
...in order to allow for composing additional playbooks from these tasks (e.g., a single playbook to install Chef, set up the roles, and converge the node, or a playbook that could stand up an entire 3-headnode cluster without intervention)


erhudy:
When reviewing this, the places that indicate large amounts of content being added/removed were really places where I moved it to a `tasks-*` playbook and then included that playbook back in the original playbook.

bootstrap/ansible_scripts/software_deployment/tasks-chef-target.yml
```diff
@@ -5,5 +5,9 @@
   gather_facts: no
   serial: "{{ serial }}"
   tasks:
-    - name: Run chef-client on {{ target }}
+    - name: Run chef-client on target
       command: chef-client
+      register: chef_client_result
+      retries: 1
```
erhudy: This builds in a single retry of the Chef run to work around some transient issues we've seen when deploying real clusters.

PR #971, opened 2016-01-21
Add ability to use vip instead of domain name for VNC proxy

semenovroman:
This solves an issue when cluster domain name can not be resolved (for example local installations)


erhudy:
Tested and has the expected effect: when setting **bcpc.vnc.proxy_use_vip** to true, the constructed URL goes through the VIP instead of the hostname.

PR #972, opened 2016-01-22
Lower default RAM quota to 8Gb vs 50Gb

semenovroman:
It is impossible to launch a VM with more than 8Gb, given default 4 vcpu limit. 50Gb quota, with only 8Gb usable confuses tenants. This does not affect existing tenancies.


erhudy:
Do you think we should also adjust maximum default instances to 4, so that people don't go "it says I can make 10 instances but I only have resources for 4!"?

semenovroman:
imho there should be no limit on number of instances they can launch. If a tenant has quota to launch 100 tiny instances, let them do it (resource exhaustion can be prevented with quota on floats etc).

erhudy:
Default quota change has expected effect.

PR #973, opened 2016-01-25
Move openstack security policies to separate files

semenovroman:
This makes it easier to work with default attributes. I would also move couple other things, like ceph and monitoring, to separate files. Thoughts?


kamidzi:
All for it.

erhudy:
double plus thumbs up

mihalis68:
And my axe

erhudy:
Tested and compared to output of a master build, everything looks the same as expected. Please rebase against master when able and then I will merge.

semenovroman:
Looks like this PR did not survive branch deletion on my repo. I submitted a new PR.

PR #976, opened 2016-01-26
Parameterize Carbon's MAX_UPDATES_PER_SECOND

kelvk:
This is a no-op that can be safely cheffed in.


erhudy:
With #977 having been merged, can you move the attribute to `attributes/monitoring.rb`?

kelvk:
Done. Update pushed.

PR #977, opened 2016-01-26
Split default attributes file into multiple files

semenovroman:
This makes it easier to work with default attributes.


erhudy:
Testing.

PR #978, opened 2016-01-27
Update vagrant_create.sh

semenovroman:
Not escaped `$existing_nets_reg_ex` var results in:

==> bootstrap: VM not created. Moving on...
Starting local Vagrant cluster...
egrep: ^vboxnet1$: No such file or directory
egrep: ^vboxnet1$: No such file or directory
egrep: ^vboxnet1$: No such file or directory
egrep: ^vboxnet1$: No such file or directory

### DIFF
--- a/bootstrap/vagrant_scripts/vagrant_create.sh
+++ b/bootstrap/vagrant_scripts/vagrant_create.sh
@@ -28,7 +28,7 @@ function remove_DHCPservers {
 
     VBoxManage list dhcpservers | grep -E "^NetworkName:\s+HostInterfaceNetworking" | awk '{print $2}' |
     while read -r network_name; do
-      [[ -n $existing_nets_reg_ex ]] && ! egrep -q $existing_nets_reg_ex <<< $network_name && continue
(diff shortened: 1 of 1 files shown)


kamidzi:
I don't think that's the only problem with script... seems like the regex isn't constructed properly [as per "will produce..." comment] in any even as `vm_networks` is a multiline string.

erhudy:
This function was something I transplanted from the old `vbox_create.sh`, I think. Since we already engage in certain other acts of bad citizenship that assume we're the only thing running in VirtualBox on a particular host, maybe we should go all the way and simply vaporize all VirtualBox DHCP servers before setting up any instances in Vagrant?

erhudy:
What do you guys want to do with this PR?

kamidzi:
This is essentially cosmetic as I see it; I think the real issue should be fixed, so I've personally held off on merging it.

erhudy:
Merging this to get it out of the way (it deleted the DHCP server that I set up on vboxnet0 fine).

PR #981, opened 2016-01-28
Report ceph health details to Zabbix and pagerduty

semenovroman:
Send first 7 lines of 'ceph health detail' to Zabbix (and zabbix pagerduty plugin), instead of just 'HEALTH_{OK,WARN,ERROR}'. This helps identify exact issue quicker and retain a record of event in Zabbix.

testing instructions:
* build cluster with at least 1 monitoring VM (set 'export MONITORING_NODES=1' in bootstrap_config.sh.overrides).
* upset ceph (ceph osd set noout)
* observe ceph health trigger with details in Zabbix


kelvk:
The `ceph health` check works fine and is getting its output logged into Zabbix history as expected. However, the use of ``ITEM.VALUE`` Zabbix macro is breaking Pagerduty notification, as `/usr/lib/zabbix/alertscripts/pagerduty.py' appears to expect a single-line value.

semenovroman:
Due to a very specific format that PagerDuty and Slack integrate this PR changed to report first line of `ceph health` and propagate it to Slack in 'status' filed (such that it is visible in Slack message)

semenovroman:
After reviewing Zabbix-PagerDuty-Slack integration, @kelvk and I decided to store `ceph health` in Zabbix and only do basic notification in Slack. Otherwise, other Zabbix checks risk being broken due to the way Pagerduty and Slack integrate.

kelvk:
The reporting to Slack for `ceph health` is:
`Ceph Health Warn : PROBLEM HEALTH_WARN too many PGs per OSD (624 > max 300) for bcpc-vm1`

However, other triggers can look odd:
`Service Nova Conductor down on bcpc-vm1 : PROBLEM 0 for bcpc-vm1`
`/etc/passwd has been changed on bcpc-vm1 : PROBLEM 2557212734 for bcpc-vm1`

Elaborating on what @semenovroman and I discussed, the Pagerduty Zabbix [guide](https://www.pagerduty.com/docs/guides/zabbix-integration-guide/) seems to imply that the content of `trigger` and `resolve` notification actions should be as-is, so deviating from that could cause unexpected consequences for us down the road. Perhaps they would be receptive improving the integrations to cater for additional information, but it sounds like something to pursue separately.

semenovroman:
So, it looks like Slack webhook integration is pretty simple, I will look into doing Zabbix->Slack integration, bypassing PagerDuty oddities.

Issue #835, opened 2015-09-21
Monitoring for BCPC-Bootstrap role


kelvk:
While outage of `BCPC-Bootstrap` node does not impact cluster availability, it still is pretty important since it provides administration functionalities. Monitoring should cover below areas at minimum:
- usual host/OS components
- chef server
- backup jobs (#794)

PR #990, opened 2016-02-10
Rename Vagrant-built bootstrap node

kelvk:
All of our Zabbix templates and hostgroups have names equivalent to Chef roles, i.e. `BCPC-Headnode`. However, Zabbix apparently cannot associate templates with hosts that have the same name. I am proposing that we change the name of Vagrant-built bootstrap node to `bcpc-vm-bootstrap` in order to maintain the Zabbix template/hostgroup naming convention when we implement #835.

The Ansible bootstrap tools seem to use `ansible-bcpc-bootstrap` as the hostname, and the legacy ones are probably due to be deprecated. I can look into extending the PR to these tools if anyone feels it is necessary.

This PR is tested by completely rebuilding from scratch via `BOOT_GO.sh`.

### DIFF
--- a/cookbooks/bcpc/templates/default/hosts.erb
+++ b/cookbooks/bcpc/templates/default/hosts.erb
@@ -25,7 +25,7 @@ ff02::2 ip6-allrouters
     if @bootstrap_node
       bs_node_ip = ( @bootstrap_node['bcpc']['bootstrap'] && @bootstrap_node['bcpc']['bootstrap']['server'] )
       str = <<-EoF
-#{bs_node_ip}\t#{@bootstrap_node['hostname']}.#{@bootstrap_node['bcpc']['hypervisor_domain']}\t#{@bootstrap_node['fqdn']}
+#{bs_node_ip}\t#{@bootstrap_node['hostname']}.#{@bootstrap_node['bcpc']['hypervisor_domain']}\t#{@bootstrap_node['fqdn']} #{@bootstrap_node['hostname']}
       EoF
     end
 %>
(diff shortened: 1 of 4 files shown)


mihalis68:
I'm not sure I understand this, but we have to support arbitrary hostnames at least in the case of real clusters in the data center. I would be most strongly against starting to have recipes forcing hostnames.

kelvk:
There isn't any forcing of hostnames in the recipes and will not affect real clusters. This is basically to rename statically-named bootstrap node in the Vagrant bootstrapping scripts, as the current `bcpc-bootstrap` hostname conflicts with upcoming use of `BCPC-Bootstrap` as a template/hostgroup name in Zabbix.

cookbooks/bcpc/templates/default/hosts.erb
```diff
@@ -25,7 +25,7 @@ ff02::2 ip6-allrouters
     if @bootstrap_node
       bs_node_ip = ( @bootstrap_node['bcpc']['bootstrap'] && @bootstrap_node['bcpc']['bootstrap']['server'] )
       str = <<-EoF
-#{bs_node_ip}\t#{@bootstrap_node['hostname']}.#{@bootstrap_node['bcpc']['hypervisor_domain']}\t#{@bootstrap_node['fqdn']}
+#{bs_node_ip}\t#{@bootstrap_node['hostname']}.#{@bootstrap_node['bcpc']['hypervisor_domain']}\t#{@bootstrap_node['fqdn']} #{@bootstrap_node['hostname']}
```
kelvk: This provides a short hostname for the bootstrap node, like we do for rest of the nodes.

erhudy:
Built a full 6-VM set and everything seemed fine.

kamidzi:
Just a side-note, this change incompatible with 5.11.1 and requires a full rebuild. I should also note that the changes referenced in https://github.com/bloomberg/chef-bcpc/pull/990#issue-132839034, don't actually change the hostname, but rather the vagrant machine name. I think the intention here was to accomplish the reverse; i.e., change the hostname, but not the machine name.

PR #992, opened 2016-02-11
Add baseline bootstrap node monitoring

kelvk:
Addresses part of #835 (backup job to be tackled separately). If your bootstrap's hostname is not `bcpc-bootstrap`, this can be cheffed in cleanly.

Once deployed, the bootstrap node should be visible in Kibana, Graphite and Zabbix. The chef_server.status Zabbix item and associated trigger will take at least 10 minutes to show up.


erhudy:
If it's not too much trouble/too scope creepy, could you add some sort of basic TFTP server monitoring to this ticket? That's one area that I've found likes to silently flake out. Even just monitoring for the tftpd server process would be a good start.

kelvk:
Pushed update with tftpd monitoring included.

erhudy:
Tested successfully on fresh build. Going back and retesting by applying over existing build.

erhudy:
Worked fine over top of an existing build (remember to update BCPC-Bootstrap role and rechef bootstrap node as well).

PR #993, opened 2016-02-16
Disable selection of compute host in functional check

kelvk:
This should make the check more representative of actual usage. When applied, the `check nova` cron job will create VMs every 10 minutes on only `bcpc-vm2`, since it is the only host in `general_compute` aggregate. Verifiable by querying `nova.instances` tables, i.e. `select created_at, deleted_at, display_name, host from instances order by created_at desc;`

### DIFF
--- a/cookbooks/bcpc/files/default/checks/nova
+++ b/cookbooks/bcpc/files/default/checks/nova
@@ -96,14 +96,16 @@ class TestNovaCompute(object):
         #
         if 'host' in self.config:
             availability_zone = self.config["cluster"] + ':' + self.config["host"]
+            server = self.client.servers.create(self.instance_name,
+                                                images[0],
+                                                flavors[0],
+                                                availability_zone=availability_zone,
+                                               )
         else:
-            availability_zone = self.config["cluster"]
-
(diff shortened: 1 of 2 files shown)


erhudy:
Testing.

erhudy:
Converged successfully. Tested in default configuration and also with adding bcpc-vm1 to the general compute aggregate to verify the scheduler was being exercised (watched `virsh list` during check execution to verify that instances were being scheduled on both hosts).

On one launch (the last of 20 `check nova` executions in a bash for loop), I got the following error from the check even though the instance was deleted successfully and nova-compute on the hosting hypervisor did not record anything unusual:
```
+------+-------+--------+---------------------------------------------------------+
| name | state |  time  |                           msg                           |
+------+-------+--------+---------------------------------------------------------+
| nova | ERROR | 11.669 | <type 'exceptions.Exception'> : unable to delete server |
+------+-------+--------+---------------------------------------------------------+
```
I could not reproduce after 50 subsequent `check nova` invocations.

erhudy:
@kelvk suspected correctly that it occurred because my test check ran at the same time as the cron job. Was able to reproduce that scenario.

PR #994, opened 2016-02-16
Parameterize and change conntrack defaults

kelvk:
`nf_conntrack: table full, dropping packet` could be observed on busy hypervisors. While it is not exactly known what the breakdown of active sessions or root cause is, increasing the defaults seems like a common operator practice. `conntrack` package is also installed for viewing of connection tracking table.


kelvk:
Added another commit to collect basic conntrack metrics via Diamond.

erhudy:
One question I had was whether it was necessary to have the separate resource to set hashsize, but it does appear to be the case; setting a new value on `net.netfilter.nf_conntrack_buckets` has no effect. An unfortunate design.

erhudy:
Minor quibble; please update the `set-nf_conntrack-hashsize` resource to use `grep -q` so that this doesn't happen:
```
  * ruby_block[set-nf_conntrack-hashsize] action run32768
 (skipped due to not_if)
```

kelvk:
Pushed update with `grep -q` change.

PR #995, opened 2016-02-17
Rework Keystone endpoint/service creation

erhudy:
### Overview
This PR revises the Chef resources that converge Keystone endpoints and services in the following ways:
* copied-and-pasted Chef resources have been replaced with generalized resources in an `each` block
* the specification of all metadata for endpoints/services has been moved to a hash in the default attributes
* parsing the JSON output of the `openstack` command via `jq` has been replaced by Ruby's JSON parser

### Rationale
* Testing upgraded API versions has been very confusing because of the number of different places where you have to modify multiple Chef resources to successfully uplift an API version.
* Resources that can be templated should be.

### Testing
* This should be a no-op on an already built cluster and should produce the same results as the previous mechanism on a newly built cluster (the commands being executed are the same).
* The diff on `keystone.rb` is really confusing and I would recommend comparing the old and new versions side by side rather than trying to make sense of the crazy diff.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -235,52 +235,10 @@ default['bcpc']['horizon']['disable_panels'] = ['containers']
 
 ###########################################
 #
-#  Keystone Settings
+# Service catalog (API versions/endpoints)
 #
 ###########################################
-#
-# Defaut log file
-default['bcpc']['keystone']['log_file'] = '/var/log/keystone/keystone.log'
-# Eventlet server is deprecated in Kilo, so by default we
-# serve Keystone via Apache now.
-default['bcpc']['keystone']['eventlet_server'] = false
-# Turn caching via memcached on or off.
-default['bcpc']['keystone']['enable_caching'] = true
-# Enable debug logging (also caching debug logging).
-default['bcpc']['keystone']['debug'] = false
-# Enable verbose logging.
-default['bcpc']['keystone']['verbose'] = false
-# Set the timeout for how long we will wait for Keystone to become operational
-# before failing (configures timeout on the wait-for-keystone-to-be-operational
-# spinlock guard).
-default['bcpc']['keystone']['wait_for_keystone_timeout'] = 120
-# Set the number of Keystone WSGI processes and threads to use by default on the
-# public API (experimentally threads > 1 may cause problems with the service
-# catalog, for now we recommend scaling only in the processes dimension)
-default['bcpc']['keystone']['wsgi']['processes'] = 5
-default['bcpc']['keystone']['wsgi']['threads'] = 1
-# The driver section below allows either 'sql' or 'ldap' (or 'templated' for catalog)
-# Note that not all drivers may support SQL/LDAP, only tinker if you know what you're getting into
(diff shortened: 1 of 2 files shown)


erhudy:
Items that are intended to be handled down the road but are not addressed in this PR:
* if the API version for an endpoint changes, the old endpoint will be automatically removed
* pervasively replace hardcoded endpoint URLs throughout the repo with templated values so that an API version change will be reconfigured in all appropriate places

erhudy:
I am not happy with the depth at which things are embedded in the matrix structure and am happy to entertain suggestions for making it less unwieldy.

erhudy:
I renamed **bcpc.keystone.api_matrix** to **bcpc.catalog** to make it a little shorter - on a separate commit at present, but probably makes sense to crush.

kamidzi:
Ran into the following:
```
chef:recipe > resources("ruby_block[keystone-create-identity-endpoint]").run_action(:run)
<...>
Compiled Resource:
------------------
# Declared in /var/chef/cache/cookbooks/bcpc/recipes/keystone.rb:269:in `block in from_file'

ruby_block("keystone-create-identity-endpoint") do
  action "run"
  retries 0
  retry_delay 2
  block_name "keystone-create-identity-endpoint"
  declared_type :ruby_block
  cookbook_name "bcpc"
  recipe_name "keystone"
  block #<Proc:0x000000037c0f08@/var/chef/cache/cookbooks/bcpc/recipes/keystone.rb:270>
  only_if { #code block }
end

NoMethodError: ruby_block[keystone-create-identity-endpoint] (bcpc::keystone line 269) had an error: NoMethodError: undefined method `[]' for nil:NilClass
        from /var/chef/cache/cookbooks/bcpc/recipes/keystone.rb:285:in `block (3 levels) in from_file'
        from /opt/chef/embedded/apps/chef/lib/chef/resource/conditional.rb:106:in `call'
        from /opt/chef/embedded/apps/chef/lib/chef/resource/conditional.rb:106:in `evaluate_block'
        from /opt/chef/embedded/apps/chef/lib/chef/resource/conditional.rb:95:in `evaluate'
        from /opt/chef/embedded/apps/chef/lib/chef/resource/conditional.rb:86:in `continue?'
        from /opt/chef/embedded/apps/chef/lib/chef/resource.rb:1074:in `block in should_skip?'
        from /opt/chef/embedded/apps/chef/lib/chef/resource.rb:1073:in `each'
        from /opt/chef/embedded/apps/chef/lib/chef/resource.rb:1073:in `find'
        from /opt/chef/embedded/apps/chef/lib/chef/resource.rb:1073:in `should_skip?'
        from /opt/chef/embedded/apps/chef/lib/chef/resource.rb:561:in `run_action'
        from (irb#1):27
```
Doesn't look like `node['bcpc']['keystone']['catalog']` is defined anywhere? As seen below:
```
chef:recipe > node['bcpc']['keystone']['catalog']
 => nil
```

Is this intended to be `node['bcpc']['catalog']['identity']` ?

erhudy:
Yes, there's an error on 285 where I messed up the replace. I'll fix it tomorrow.

erhudy:
Force pushed the fix to the wrongly named attribute.

PR #998, opened 2016-02-22
Enable crashkernel

### LINKED ISSUE #991: Improve post-crash analysis capabilities
Because real machines sometimes kernel panic, it would be good to be able to collect debugging information on kernel panics instead of screenshotting what's visible of the crash trace off the console.

erhudy:
- installs and enables kdump-tools on already deployed systems (requires system cold reboot to activate)
- updates preseed template to install kdump-tools in post-install and use the desired kernel line specified in the environment (borrowed from https://github.com/bloomberg/chef-bach/pull/377)


.gitignore
```diff
@@ -26,3 +26,4 @@ environments/*
 vbox_override.sh
 .*.sw*
 *~
+.DS_Store
```
erhudy: Piggyback change (unrelated to the rest of the PR).

cookbooks/bcpc/templates/default/cobbler.bcpc_ubuntu_host.preseed.erb
```diff
@@ -96,12 +94,8 @@ d-i     pkgsel/include string openssh-server
 
 d-i     preseed/late_command string true && \
         wget "http://$http_server:$http_port/cblr/svc/op/nopxe/system/$system_name" -O /dev/null ;\
-        <% if @node['bcpc']['preseed'] && @node['bcpc']['preseed']['kernel'] %>
-        in-target apt-get install -y <%= @node['bcpc']['preseed']['kernel'] %> ;\
-        <% else %>
-        true ;\
-        <% end %>
-        wget "http://$http_server:$http_port/cblr/svc/op/nopxe/system/$system_name" -O /dev/null ;\
```
erhudy: I removed the above section because it seemed redundant if we are specifying the kernel version directly as the value of `base-installer/kernel/image`. (The `wget` line was removed because it was a duplicate.)

erhudy:
I checked out the Cobbler template change on hardware and verified that a system built using it does come with `kdump-tools` already install and crashkernel activated.

erhudy:
This can be tested as is, but I need to augment a few additional areas in the templates (e.g., removing `nousb` from the crash kernel's arguments so that you can actually type things in BusyBox).

erhudy:
Added a few more parameterizations of items and this is ready to go. 128MB of memory wasn't enough for the crash kernel to work on hardware, but 256MB worked fine.

*NOTE*: there is some issue with how our VirtualBox stuff is set up that causes the crash kernel itself to panic if you cause a kernel panic in the system runtime kernel (e.g., poking sysrq). If you want to test the crash dump part of the kernel, you'll need to use hardware or figure out what's making VirtualBox angry.

erhudy:
Running into a Brand New Set of out of memory issues on PXE bootstraps that should've been (and previously were) solved already. Back to arguing with computers again...

erhudy:
This needs a fix to ensure that linux-image-extra gets installed, which contains certain modules that our hardware needs to boot. If you allow it to track the latest kernel, it works fine, but if you specify an older kernel package, that package will get installed without the extras, and then the system will fail to boot.

erhudy:
Added a little logic in the template (forgive me) that will automatically install the appropriate linux-image-extra package if a particular kernel version is specified in **bcpc.preseed.kernel**.

erhudy:
Force pushed to fix merge conflict.

semenovroman:
After more experimentation the following was found:
* setting `crashkernel=384M` is sufficient for successful dump, `256M` fails to boot crash kernel
* `512M` and `768M` values also work
* >= `1024M` will fail to boot crashkernel at all (we can check if this will fail by a message in `dmesg` on a live system - `crashkernel reservation failed - No suitable area found.`)

Hence, crashkernel should be set >= 384 && <= 768, after setting and rebooting `dmesg` should be checked for a `crashkernel reservation failed` message, then actual kernel dump should be tested to make sure it does not run OOM during dump.

reference:
* https://github.com/torvalds/linux/blob/master/Documentation/kernel-parameters.txt#L781-L816
* http://h20195.www2.hp.com/V2/getpdf.aspx/4AA5-4775ENW.pdf (pages 13-16)
* https://lkml.org/lkml/2012/3/13/36
* https://github.com/torvalds/linux/blob/master/Documentation/kdump/kdump.txt

erhudy:
Rebased off master to resolve merge conflict.

erhudy:
Closing this PR for now so that it does not languish. It will be revisited someday.

PR #999, opened 2016-02-23
Allow specifying a particular kernel for GRUB to boot

erhudy:
If desired kernel is unavailable, it will use the latest kernel, so there is an automatic fallback built in if an incorrect or missing kernel version is specified.


cookbooks/bcpc/recipes/system.rb
```diff
@@ -43,18 +70,6 @@
...
-    code <<-EOH
-        for i in /sys/block/sd?; do
-            echo deadline > $i/queue/scheduler
-        done
-        echo GRUB_CMDLINE_LINUX_DEFAULT=\\\"\\$GRUB_CMDLINE_LINUX_DEFAULT elevator=deadline\\\" >> /etc/default/grub
-        update-grub
-    EOH
-    not_if "grep 'elevator=deadline' /etc/default/grub"
-end
-
```
erhudy: I removed this resource and implemented it in the template for `/etc/default/grub` instead, since that file is now managed by a **template** resource.

cookbooks/bcpc/templates/default/system.etc_default_grub.erb
```diff
@@ -0,0 +1,47 @@
...
+#GRUB_DISABLE_LINUX_UUID=true
+
+# Uncomment to disable generation of recovery mode menu entries
+#GRUB_DISABLE_RECOVERY="true"
+
+# Uncomment to get a beep at grub start
+#GRUB_INIT_TUNE="480 440 1"
+
+# add deadline elevator
+GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT elevator=deadline"
```
erhudy: Line corresponding to the removed resource from above.

cookbooks/bcpc/attributes/default.rb
```diff
@@ -14,6 +14,11 @@
 default['bcpc']['virt_type'] = "kvm"
 # Define the kernel to be installed. By default, track latest LTS kernel
 default['bcpc']['preseed']['kernel'] = "linux-image-generic-lts-trusty"
+# Define a specific kernel version to have GRUB default to (if non-nil)
+# - specify kernel like pattern "3.13.0-61-generic"
+# - a wrong pattern here will not cause the system to fail to boot, it will
+#   instead revert to the default boot order
+default['bcpc']['kernel_version'] = nil
```
erhudy: I am open to suggestions on how to relate this attribute and **bcpc.preseed.kernel**, though doing so may not be necessary (or desirable).

kamidzi:
Looks like using a bogus value, wrong pattern, etc, will cause the chef run to fail. Something like the following:

```
Recipe: bcpc::system
  * apt_package[linux-image-BOGUS] action install
    * No candidate version available for linux-image-BOGUS
    Error executing action `install` on resource 'apt_package[linux-image-BOGUS]'
```

erhudy:
Yes, I don't think that's avoidable. In my view the Chef run should fail on bad information so that the person can fix it instead of assuming everything worked when it actually fails silently.

kamidzi:
I agree that it should fail, though the comment about supplying an incorrect value not causing the machine boot to fails is a bit confusing as I misunderstood what the intended behavior is... Reading it, I had expected the change to work, but not cause problems rather than not re-configure at all. Just to be clear, the chef-failure is the intended behavior on bad inputs?

erhudy:
Chef should die, yes. The comment about a bad value not being able to render the system unbootable wasn't meant to be an implication that Chef would allow it through, more as a reassurance that you can't hard brick the system even if somehow it did make its way through Chef.

erhudy:
The origin of that comment came from before I added the package resources, which I added when I realized that not installing linux-image-extra and generating an initrd could cause physical systems to fail to boot. Do you think the comment should be updated?

kamidzi:
Not essential, but updating the comment would disambiguate.

kamidzi:
All looks good. I had one set of circumstances where a sequence of upgrades caused the system to fail to boot (in VirtualBox), but I could not reproduce. The following cases worked fine:

from -45 -> "bogus", reboot
from -45 -> -77, reboot, -77 -> -45, reboot, -45 -> bogus, reboot, "bogus" -> nil
from -45 -> -77, reboot, -77 -> -45, reboot, -45 -> "bogus", "bogus" -> nil, reboot

PR #1000, opened 2016-02-29
Enable Liberty support and make it the default version

erhudy:
For full details on enabling Liberty support, please see `docs/installing_liberty.md`.

This should cleanly chef into existing Kilo builds. Upgrading from Kilo to Liberty within this PR is available and can be triggered by changing **bcpc.openstack_release** to `liberty`. A few changes will be made to existing Kilo installations, mostly reordering segments of configuration files because of configuration items in `[DEFAULT]` in Kilo being moved to their own section in Liberty.

**Remember to upload all roles again before re-cheffing for an upgrade!**

Recommended focus areas for testing
---
* Check that the explicit API versions in `/root/api_versionsrc` don't break anything - where necessary I forced utilities that complained to use a particular API version, but I may have missed something
* (Liberty only) Check that all equivalent functionality from Kilo works as expected, particularly in Horizon
* (upgrading only) Check that the service catalog endpoints are rewritten correctly to use the new versions
* Test multiple head node upgrades (strongly recommended to stop services on all head nodes but the VIP holder, upgrade that one, upgrade the remaining ones in turn, then upgrade workers)


erhudy:
Auspiciously given PR #1000.

erhudy:
One callout for the service catalog changes: I changed the interpolation character from `$` to `%` on certain service URIs that require the tenant ID to be inserted into the URI. Kilo seems to be fine with either, while Liberty refused to work with `$`.

erhudy:
This can be tested but please withhold merge until #1016 (at which point I will rebase and update the `linux_net.py` patch, because a different set of checksums is needed for Liberty).

erhudy:
Rebased off master and added support for upgrading in-place (with a few other tweaks forthcoming).

erhudy:
This PR is now fully open for testing. All three paths of fresh Kilo, fresh Liberty and Kilo-to-Liberty should work and should all be tested and exercised.

Known issues
---
* Some of the checks get crabby under Liberty. `check nova` in particular reports scheduling issues that tend to go away after a HUP or two, or launching another instance with `nova boot`.
* The cleanup recipe currently only has a HUP in it and doesn't clean up other things left behind by Kilo at present.
* A number of newer packages in the UCA Liberty repository are not upgraded by the recipes (mostly QEMU and Python dependency packages). These can be upgraded at the user's convenience.
* With Liberty, CLI commands output **InsecurePlatformWarning** error messages from urllib3. Fixing this would require upgrading to Python 2.7.11.

erhudy:
Sorry for the magnitude of the pull request - I had intended to keep the Kilo-to-Liberty upgrade stuff in a separate commit, but smashed it all together during a rebase.

erhudy:
Added a commit that adds the **nova_api** database. I don't believe this is really used in Liberty (maybe if using cells), but is required for Mitaka. Liberty lays the groundwork for it and it doesn't hurt anything to add it now.

erhudy:
The issue with `check nova` not working after an initial build appears to be because the work nodes are not being enrolled in host aggregates properly on the initial run, but then they do get enrolled on a second Chef run, and then `check nova` works. Need to dig into that issue.

erhudy:
Problem is because the compute nodes are not populated for some reason on the first run:
```
2016-04-26 13:41:04.792 65413 DEBUG nova.api.openstack.wsgi [req-8ec40dd5-343d-4c2c-abe0-3b41d238728e 2f29a1754b9946a88d54290cb4278a61 3b078997009941028fd4d23359d2adaa - - -] Action: 'action', calling method: <bound method AggregateController._add_host of <nova.api.openstack.compute.aggregates.AggregateController object at 0x7f743f8d1910>>, body: {"add_host": {"host": "bcpc-vm1"}} _process_stack /usr/lib/python2.7/dist-packages/nova/api/openstack/wsgi.py:789
2016-04-26 13:41:04.812 65413 INFO nova.api.openstack.wsgi [req-8ec40dd5-343d-4c2c-abe0-3b41d238728e 2f29a1754b9946a88d54290cb4278a61 3b078997009941028fd4d23359d2adaa - - -] HTTP exception thrown: Compute host bcpc-vm1 could not be found.
```

erhudy:
Underlying issue appears to be a race condition on the permissions for `/var/run/libvirt/libvirt-sock`. The post-install script for nova-compute-libvirt adds the nova user to the libvirtd group, but nova-compute has already started by this point and fails to connect to libvirt. Restarting nova-compute resolves the issue, registers the compute node, and allows it to be added to host aggregates.

erhudy:
Pushed fix that ensures nova-compute is restarted after nova-compute-libvirt is installed (really after nova-compute-qemu or nova-compute-kvm are installed, both of which pull in nova-compute-libvirt as a dependency).

kamidzi:
Kilo-to-Liberty seems to work cleanly.

kamidzi:
Ran into some issues with stability of rabbimq-server and non-responsive v3 endpoints. Will need to rebuild to attempt to reproduce though..

erhudy:
On fresh build or upgrade?

kamidzi:
Upgrade. I have a fresh build going in parallel

erhudy:
Okay, I will be interested to hear the results.

cookbooks/bcpc/templates/default/adminrc.erb
```diff
@@ -4,12 +4,16 @@
...
-# export OS_CACERT=/etc/ssl/certs/ssl-bcpc.pem
-# both Glance API versions are enabled but prefer v2
-export OS_IMAGE_API_VERSION=2
+<% if get_api_version(:identity) == "3" %>
+export OS_PROJECT_DOMAIN_NAME=default
+export OS_USER_DOMAIN_NAME=default
+<% end %>
+
+# API versions in a separate file so that they can be used when bootstrapping Keystone services
+source /root/api_versionsrc
```
kamidzi: Something I ran into and to be careful about is when guards such as 
```
not_if " . /root/adminrc ..."
```
are written. It seems like if guard interpreter is not set explicitly to `:bash`, there can be silent failures with the `source` keyword. One can see this by running chef in debug mode.

erhudy: I'm just used to writing `source`, it should be `.` everywhere.

erhudy:
Changed `source` to `.` in `adminrc`. I didn't find any other instances where I did that, though I may have missed other instances (searching for "source " turns up a lot of stuff).

erhudy:
After discussion with @kamidzi, this PR has been updated to make Liberty the preferred default version.

erhudy:
Rebased and force pushed this branch to resolve merge conflicts against master.

erhudy:
Rebased again to resolve more merge conflicts.

kamidzi:
Upgrading libvirt-bin seems to be problematic. I get the following:
```
STDERR: Configuration file '/etc/libvirt/qemu.conf'
     ==> Modified (by you or by a script) since installation.
     ==> Package distributor has shipped an updated version.
...
```

erhudy:
We modify that file in the **libvirt-device-acls** resource, but maybe it's no longer necessary. I will take a look when time permits.

erhudy:
Looks like we probably still need to modify that file in order to allow QEMU access to `/dev/net/tun`. I will modify the recipes to turn that into a properly managed template and check for differences between Kilo and Liberty that need to be accounted for. I will also look at trying to reconfigure apt/dpkg across the board to use `-o Dpkg::Options::='--force-confdef' -o Dpkg::Options::='--force-confold'` instead of requiring this to be inserted in more and more places in the recipes, because of all the places where the issue of a managed config file conflicting with dpkg crops up.

erhudy:
Added a changeset that reconfigures apt/dpkg systemwide to not get their jimmies rustled by our managed config files. (`/etc/libvirt/qemu.conf` does not change between Kilo and Liberty, so I left it as is.)

erhudy:
Encountered this failure when attempting to upgrade a multiple headnode cluster:
```
2016-05-26 19:37:58.535522 2016-05-26 19:37:58.535 265529 DEBUG keystone.common.cache.core [req-6997a30b-6983-49f4-931f-11dbb168c252 - - - - -] CACHE_GET: Key: "'d5db38400f345c634aee9b815a0cd215fef0aaf7'" Value: "([{'expires': '2016-05-26T19:56:46.000000Z', 'id': 'f9dfd5ee30a5fcdb272d1bdb194d63f3', 'audit_id': '3cfpyOK6TASABqGScRdnxw'}], {'ct': 1464289707.503163, 'v': 1})" get /usr/lib/python2.7/dist-packages/keystone/common/cache/core.py:60
2016-05-26 19:37:58.539069 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi [req-6997a30b-6983-49f4-931f-11dbb168c252 - - - - -] 'str' object has no attribute 'strftime'
2016-05-26 19:37:58.539131 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi Traceback (most recent call last):
2016-05-26 19:37:58.539140 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi   File "/usr/lib/python2.7/dist-packages/keystone/common/wsgi.py", line 248, in __call__
2016-05-26 19:37:58.539145 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi     result = method(context, **params)
2016-05-26 19:37:58.539150 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi   File "/usr/lib/python2.7/dist-packages/keystone/common/controller.py", line 163, in inner
2016-05-26 19:37:58.539154 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi     return f(self, context, *args, **kwargs)
2016-05-26 19:37:58.539159 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi   File "/usr/lib/python2.7/dist-packages/keystone/auth/controllers.py", line 560, in revocation_list
2016-05-26 19:37:58.539164 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi     t['expires'] = utils.isotime(expires)
2016-05-26 19:37:58.539168 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi   File "/usr/lib/python2.7/dist-packages/keystone/common/utils.py", line 504, in isotime
2016-05-26 19:37:58.539172 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi     st = at.strftime(_ISO8601_TIME_FORMAT
2016-05-26 19:37:58.539177 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi AttributeError: 'str' object has no attribute 'strftime'
2016-05-26 19:37:58.539180 2016-05-26 19:37:58.536 265529 ERROR keystone.common.wsgi 
```

erhudy:
Turned out to seemingly be caused by stale *.pyc files with old Keystone code causing issues. Vaporized all of those from `/usr/lib/python2.7/dist-packages`, restarted Apache, and everything worked as expected again. Will look at adding a resource to trigger cleaning these files up on OpenStack package upgrades.

erhudy:
After finishing the upgrade on the head node successfully, it looks like we will need to update our policies to work correctly with newer endpoints. Will review what needs to be changed.

erhudy:
The problem appears to have been caused by data in memcached written in a format by Kilo that the Liberty KVS provider did not like - the code expected to be able to decode the timestamps in the cache to datetime.datetime objects, but instead was leaving them as strings, which caused explosions further down the line. Dumping the cache out on Keystone upgrade should solve that issue.

Added two more changesets - one to clean out PYC files on package upgrades, and one to flush memcached on Keystone upgrade.

Turning my attention to policy updates now.

erhudy:
More gory details on the above issue with tokens: it now appears that tokens being written using the Identity v2 API are not format-compatible with tokens written using the Identity v3 API. This issue is clearly not yet solved.

erhudy:
After some additional experimentation, I believe that there exists a reproducible test case that triggers the issue:

1. Verify that everything's working via the command line after sourcing `/root/adminrc` (try out commands like `nova list`, `glance image-list`, `openstack server list`, etc.)
2. Log into Horizon as the admin user.
3. Log out of Horizon.
4. Wait a few seconds.
5. Try the CLI commands again.

This reliably sets up the failure for me. Flushing all memcached instances restores functionality.

erhudy:
The evergreen solution for all of our other Keystone woes, Fernet tokens, appears to resolve this issue. The bug appears to be in the revocation logic for PKI tokens. Rather than expend time working on PKI tokens when we intend to abandon them anyway, I will just add a changeset on this PR to bootstrap Fernet tokens in the environment and wire Keystone up to use them. (It won't deal with key rotation and will just stash the Fernet keys in the data bag for now, like all our other terrible antipatterns.)

erhudy:
Added a changeset that switches to Fernet tokens for Liberty and reverts the parameterization of the token provider type - tokens are fixed to PKI on Kilo and Fernet otherwise. Fernet keys are persisted in the data bag. I made an effort to add some code to deal with rotations appropriately, but since Chef does not know that keys were rotated, it will just clobber values in the data bag with old keys and revert the rotation. Key rotation can be manually executed by rotating keys on one head node via `keystone-manage fernet_rotate`, recheffing to update the data bag, then deleting the contents of `/etc/keystone/fernet-keys` on other head nodes and recheffing (so that the recipe has no existing keys to read into the data bag and clobber the newer keys).

erhudy:
Rebased off master to incorporate all the changes that have gone into master in the last 2 weeks.

erhudy:
Current outstanding issues to be addressed that I am aware of:
* update Glance, Keystone, Heat policies for Liberty
* modify Horizon openrc template to pre-populate user/project domain environment variables

erhudy:
Interesting issue cropping up for me now (I merged #1054 into my test branch to verify the presence of the problem). I thought it was a recurrence of the ownership issue on the libvirt socket, but on my local builds now it looks like compute hosts are now being registered with their FQDN instead of their hostname, which is causing failures when attempting to add compute hosts to an aggregate:

```
root@bcpc-vm1:/var/log/cinder# nova service-list
/usr/lib/python2.7/dist-packages/urllib3/util/ssl_.py:97: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
+----+------------------+--------------------------------------+----------+---------+-------+----------------------------+-----------------+
| Id | Binary           | Host                                 | Zone     | Status  | State | Updated_at                 | Disabled Reason |
+----+------------------+--------------------------------------+----------+---------+-------+----------------------------+-----------------+
| 1  | nova-scheduler   | bcpc-vm1.hypervisor-bcpc.example.com | internal | enabled | up    | 2016-06-18T20:53:33.000000 | -               |
| 2  | nova-cert        | bcpc-vm1.hypervisor-bcpc.example.com | internal | enabled | up    | 2016-06-18T20:53:29.000000 | -               |
| 3  | nova-consoleauth | bcpc-vm1.hypervisor-bcpc.example.com | internal | enabled | up    | 2016-06-18T20:53:29.000000 | -               |
| 4  | nova-conductor   | bcpc-vm1.hypervisor-bcpc.example.com | internal | enabled | up    | 2016-06-18T20:53:34.000000 | -               |
| 6  | nova-compute     | bcpc-vm1.hypervisor-bcpc.example.com |          | enabled | up    | 2016-06-18T20:53:37.000000 | -               |
| 7  | nova-network     | bcpc-vm1.hypervisor-bcpc.example.com | internal | enabled | up    | 2016-06-18T20:53:32.000000 | -               |
| 8  | nova-compute     | bcpc-vm2.hypervisor-bcpc.example.com |          | enabled | up    | 2016-06-18T20:53:35.000000 | -               |
| 9  | nova-network     | bcpc-vm2.hypervisor-bcpc.example.com | internal | enabled | up    | 2016-06-18T20:53:26.000000 | -               |
| 10 | nova-compute     | bcpc-vm3.hypervisor-bcpc.example.com |          | enabled | up    | 2016-06-18T20:53:34.000000 | -               |
| 11 | nova-network     | bcpc-vm3.hypervisor-bcpc.example.com | internal | enabled | up    | 2016-06-18T20:53:28.000000 | -               |
+----+------------------+--------------------------------------+----------+---------+-------+----------------------------+-----------------+
```

```
  * bcpc_host_aggregate[general_compute] action member
    
    Error executing action `member` on resource 'bcpc_host_aggregate[general_compute]'
    
    RuntimeError
    ------------
    Failed to add host bcpc-vm2 to aggregate general_compute: Compute host bcpc-vm2 could not be found. (HTTP 404) (Request-ID: req-5ad20bef-392f-4080-9bb0-958f9847c8d7)
```

erhudy:
After further testing, this appears to be a behavior change between Vagrant 1.8.1 and 1.8.4. We have always provided the FQDN for VM hostnames in the Vagrantfile, so I'll look at a way to work around this problem.

erhudy:
Repaved this PR to resolve a merge conflict in `cookbooks/bcpc/providers/osflavor.rb`. Also reordered/consolidated some fragmented changesets, mostly grouping the policy changes into a single changeset per service, and consolidating Fernet token changes into one changeset.

erhudy:
Rebased once again to merge the regression changeset into the Fernet token changeset and resolve a merge conflict on the version number.

erhudy:
There's a place right now during Kilo->Liberty upgrade where it bombs out during flavor creation, then one of the daemons that gets restarted afterward gets back into a good state and a subsequent Chef run proceeds normally. It's probably nova-api that needs to be restarted earlier in the upgrade.

erhudy:
Flavor creation error is probably a timing issue. I am testing with a synchronization resource in the flavors recipe to verify that everything is ready and not returning 503s before attempting to manage flavors.

erhudy:
Fixes pushed for flavor creation error and also for 2 wrong return codes in the HAProxy configuration.

erhudy:
As a reminder, please remember to re-upload all roles before starting the upgrade, because some of them changed. I tend to forget to do that.

cookbooks/bcpc/attributes/catalog.rb
```diff
@@ -0,0 +1,98 @@
...
+    'description' => 'OpenStack Image Service',
+    'ports' => {
+      'admin' => 9292,
+      'internal' => 9292,
+      'public' => 9292
+    },
+    'uris' => {
+      'admin' => 'v2',
+      'internal' => 'v2',
+      'public' => 'v2'
```
erhudy: I suspect that these should actually be set to `''`, having done some more research on Glance. Glance treats its version numbers differently than other services and appends `v1` or `v2` depending on the version of the API being requested. Testing this on both Kilo and Liberty.

kamidzi:
Is the intent to completely deprecate Cinder APIv1 by manipulating the endpoint urls in the service catalog. I realize the API is deprecated, but if this is case, it should probably be noted in release notes.

erhudy:
Yes. The only reason the fake Cinder v1 endpoint remains around is for Horizon (see second-to-last bullet point on http://docs.openstack.org/releasenotes/horizon/liberty.html).

PR #1001, opened 2016-03-01
Make monitoring node searches more specific

kelvk:
No-op change. This paves the way to allow more granular control over what each monitoring node in a cluster should run.


kamidzi:
No-op confirmed.

PR #1003, opened 2016-03-02
Parameterize bcpc.preseed.kernel

erhudy:
- also install linux-image-extra if a particular kernel version is being
  tracked
- remove redundant wget from late_command
- move success signal to Cobbler to end of late_command

### DIFF
--- a/cookbooks/bcpc/templates/default/cobbler.bcpc_ubuntu_host.preseed.erb
+++ b/cookbooks/bcpc/templates/default/cobbler.bcpc_ubuntu_host.preseed.erb
@@ -50,7 +50,7 @@ d-i     clock-setup/ntp boolean true
 d-i     clock-setup/ntp-server  string <%= @node['ntp']['servers'].first %>
 # This will install the base 12.04 kernel image.  If you comment out the
 # following line, it will install the upgraded kernel LTS stack.
-d-i     base-installer/kernel/image     string linux-server
+d-i     base-installer/kernel/image     string <%= @node['bcpc']['preseed']['kernel'] %>
 d-i     passwd/root-login       boolean false
 d-i     passwd/make-user        boolean true
 d-i     passwd/user-fullname    string ubuntu
@@ -95,14 +95,11 @@ d-i     debian-installer/exit/poweroff  boolean false
 d-i     pkgsel/include string openssh-server
 
 d-i     preseed/late_command string true && \
-        wget "http://$http_server:$http_port/cblr/svc/op/nopxe/system/$system_name" -O /dev/null ;\
-        <% if @node['bcpc']['preseed'] && @node['bcpc']['preseed']['kernel'] %>
-        in-target apt-get install -y <%= @node['bcpc']['preseed']['kernel'] %> ;\
-        <% else %>
-        true ;\
+        <% unless node.default['bcpc']['preseed']['kernel'] == node['bcpc']['preseed']['kernel'] %>
+        in-target apt-get -y install <%= node['bcpc']['preseed']['kernel'].gsub('linux-image', 'linux-image-extra') %> ;\
         <% end %>
-        wget "http://$http_server:$http_port/cblr/svc/op/nopxe/system/$system_name" -O /dev/null ;\
         in-target env no_proxy=$http_server wget "http://$http_server:$http_port/cobbler/pub/scripts/get-ssh-keys" -O /root/get-ssh-keys ;\
         in-target chmod 0755 /root/get-ssh-keys ; in-target install -d -m0600 /root/.ssh ;\
         in-target env no_proxy=$http_server /root/get-ssh-keys root -o /root/.ssh/authorized_keys ;\
-        echo -e "\n<%=@bootstrap_node['bcpc']['bootstrap'] && @bootstrap_node['bcpc']['bootstrap']['server']%>\t<%=@bootstrap_node['fqdn']%>\n" >> /target/etc/hosts
+        echo -e "\n<%=@bootstrap_node['bcpc']['bootstrap'] && @bootstrap_node['bcpc']['bootstrap']['server']%>\t<%=@bootstrap_node['fqdn']%>\n" >> /target/etc/hosts ;\
+        wget "http://$http_server:$http_port/cblr/svc/op/nopxe/system/$system_name" -O /dev/null ;
--- a/cookbooks/bcpc/templates/default/cobbler.bcpc_ubuntu_host.preseed.erb
+++ b/cookbooks/bcpc/templates/default/cobbler.bcpc_ubuntu_host.preseed.erb
@@ -95,8 +95,9 @@ d-i     debian-installer/exit/poweroff  boolean false
 d-i     pkgsel/include string openssh-server
 
 d-i     preseed/late_command string true && \
-        <% unless node.default['bcpc']['preseed']['kernel'] == node['bcpc']['preseed']['kernel'] %>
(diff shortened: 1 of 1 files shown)


cookbooks/bcpc/templates/default/cobbler.bcpc_ubuntu_host.preseed.erb
```diff
@@ -95,14 +95,11 @@ d-i     debian-installer/exit/poweroff  boolean false
 d-i     pkgsel/include string openssh-server
 
 d-i     preseed/late_command string true && \
-        wget "http://$http_server:$http_port/cblr/svc/op/nopxe/system/$system_name" -O /dev/null ;\
-        <% if @node['bcpc']['preseed'] && @node['bcpc']['preseed']['kernel'] %>
-        in-target apt-get install -y <%= @node['bcpc']['preseed']['kernel'] %> ;\
-        <% else %>
-        true ;\
+        <% unless node.default['bcpc']['preseed']['kernel'] == node['bcpc']['preseed']['kernel'] %>
```
kamidzi: So is this really about the versions not being the same, or a consequence of the default value being a meta package? I understand that semantically, the result is the same. I'm ambivalent as to whether "linux-image-generic-lts-<whatever-distro>" should just be hardcoded here instead of sort of hardcoded in the default attributes file. Leaving it as it is in the PR now would minimize churn on this file, but how often do we change distros?

erhudy: It's a consequence of the default value being a meta package. If the default value is `linux-server` or `linux-image-generic-lts-trusty`, and you're okay with that, then you get whatever the latest kernel is along with all the extra modules, which can be verified by following the dependencies back to `linux-image-generic` and seeing that it pulls in both the latest versions of `linux-image` and `linux-image-extra`. If you say you want a particular version that isn't the latest one on the mirror (which is the use case for this PR), specifying `linux-image-3.13.0-XX-generic` won't give you the matching `linux-image-extra` package, and so the template abuses ERB slightly to add a little logic that will attempt to find and install the appropriately matching package. Otherwise, our hardware doesn't boot and everyone is sad.

I'm not sure what you're proposing to hardcode here. What does your solution look like?

(The corollary to this issue is why the hell do some kernel modules need to be broken out into an extras package in 2016? Disk space is cheap and this wouldn't be an issue if every module just came as part of `linux-image` as they all should.)

kamidzi: I was thinking something like `unless /^linux-image-generic-/.match node['bcpc']['preseed']['kernel']`. I say this only because I understood the conditional as "unless the default attribute value matches the current attribute value of ['bcpc']['preseed']['kernel'], install the extras." I view the scenario more as "install the extras unless the currently selected kernel is a metapackage [which will just do it]."

erhudy: That's reasonable, I'm fine with an approach that doesn't explicitly depend on the default setting of bcpc.preseed.kernel being a metapackage. I'll push something to change the behavior tomorrow.

kamidzi:
For `bcpc.preseed.kernel`, the intent was for the specification of metapackages only. For example, using an existing bootstrap server and bringing up some pxe clients with `spawn_local_vms.sh`, you can see that as of today, the stock ubuntu has two kernels series available, `3.13.0.79`, `3.13.0.24` from different repos using
```
apt-cache policy linux-{generic,image-{server,generic{,-lts-trusty}}}
```
(The above output will be different for vagrant builds; i.e, no records are listed as installed)

Configuring `bcpc.preseed.kernel` to `3.13.0.24-generic` and PXE booting a client yields the following result, from `/var/log/apt/history.log.` (These are all installations during OS install)

```
Start-Date: 2016-03-03  13:45:10
Commandline: apt-get -o APT::Status-Fd=4 -o APT::Keep-Fds::=5 -o APT::Keep-Fds::=6 -q -y --no-remove install linux-generic
Install: libnl-genl-3-200:amd64 (3.2.21-1, automatic), linux-headers-generic:amd64 (3.13.0.79.85, automatic), wireless-regdb:amd64 (2013.02.13-1ubuntu1, automatic), linux-image-extra-3.13.0-79-generic:amd64 (3.13.0-79.123, automatic), linux-image-3.13.0-79-generic:amd64 (3.13.0-79.123, automatic), libnl-3-200:amd64 (3.2.21-1, automatic), linux-headers-3.13.0-79-generic:amd64 (3.13.0-79.123, automatic), linux-headers-3.13.0-79:amd64 (3.13.0-79.123, automatic), crda:amd64 (1.1.2-1ubuntu2, automatic), linux-firmware:amd64 (1.127.20, automatic), linux-image-generic:amd64 (3.13.0.79.85, automatic), linux-generic:amd64 (3.13.0.79.85)
End-Date: 2016-03-03  13:45:35
...
Start-Date: 2016-03-03  13:48:21
Commandline: apt-get -y install 3.13.0-24-generic
Install: linux-signed-image-3.13.0-24-generic:amd64 (3.13.0-24.47), libunwind8:amd64 (1.1-2.2ubuntu3, automatic), linux-cloud-tools-3.13.0-24:amd64 (3.13.0-24.47, automatic), linux-tools-common:amd64 (3.13.0-79.123, automatic), linux-cloud-tools-common:amd64 (3.13.0-79.123, automatic), linux-cloud-tools-3.13.0-24-generic:amd64 (3.13.0-24.47), libdw1:amd64 (0.158-0ubuntu5.2, automatic), linux-headers-3.13.0-24:amd64 (3.13.0-24.47, automatic), linux-image-extra-3.13.0-24-generic:amd64 (3.13.0-24.47), linux-image-3.13.0-24-generic:amd64 (3.13.0-24.47), linux-tools-3.13.0-24-generic:amd64 (3.13.0-24.47), sbsigntool:amd64 (0.6-0ubuntu7, automatic), linux-headers-3.13.0-24-generic:amd64 (3.13.0-24.47), linux-tools-3.13.0-24:amd64 (3.13.0-24.47, automatic)
End-Date: 2016-03-03  13:48:46
```

Two things to note:
* specifying the kernel in that manner _x.y.z.0-w-generic_ will pull in all the extras and more
* the specified kernel is not the one that is activated for boot, see:
```
root@name:/var/log/apt# uname -r
3.13.0-79-generic
root@name:/var/log/apt# apt-cache policy linux-generic
linux-generic:
  Installed: 3.13.0.79.85
  Candidate: 3.13.0.79.85
  Version table:
 *** 3.13.0.79.85 0
        500 http://us.archive.ubuntu.com/ubuntu/ trusty-updates/main amd64 Packages
        500 http://security.ubuntu.com/ubuntu/ trusty-security/main amd64 Packages
        100 /var/lib/dpkg/status
     3.13.0.24.28 0
        500 http://us.archive.ubuntu.com/ubuntu/ trusty/main amd64 Packages
```
Inspecting the the list of packages and grub config, the older kernel is indeed installed and added to the boot config, simply not as the default option.

Continuing with version `3.13.0.24`, you'll notice that the output for the following are different:
```
apt-cache showpkg  linux-image-3.13.0-24-generic
apt-cache showpkg  linux-image-3.13.0-24
```

Wrt `bcpc.preseed.kernel`, specifying `linux-image-3.13.0-24-generic` does the right thing wrt to grub config and extra packages. Specifying `linux-image-3.13.0-24` _does not_ wrt to grub config.

--

I say all this merely as an observation and to hopefully prevent others from making the same config mistake I had.

PR #1006, opened 2016-03-09
Change from using cp -r to git clone

erhudy:
This fixes an issue reported by a user where the `cp -r` was somehow dereferencing a symlink in the Diamond repository used during the binary build stage, causing a subsequent `git checkout` command to fail. Switching to `git clone` reconstitutes a fresh copy of the repository from the raw Git objects, working around this potential issue.


erhudy:
(PR comes with some free whitespace changes courtesy of Atom)

erhudy:
If testing, remove `$HOME/.bcpc-cache/build_bins_cache` from your workstation so that the bootstrap VM has to rebuild all binaries instead of copying them forward from a previous build.

PR #1017, opened 2016-04-04
Persist ipset setup across reboots

kelvk:
Monitoring IP sets are not re-created when a monitoring node reboots since they are saved in `/tmp` and only created via chef runs. Insert a snippet like below into Chef environment, rechef and reboot.
<pre>
    "bcpc": {
      "monitoring": {
        "vip": "10.0.100.6",
        "external_clients": [
          "10.0.98.1-10.0.98.2",
          "10.0.99.1-10.0.99.2"
        ]
      },
      ...
    }
</pre>

Following the reboot,  the IP addresses should appear in `ipset list`...
<pre>
root@bcpc-vm4:~# ipset list
Name: monitoring-clients
Type: hash:ip
Revision: 2
Header: family inet hashsize 1024 maxelem 65536
Size in memory: 16568
References: 1
Members:
10.0.99.1
10.0.98.2
10.0.99.2
10.0.98.1
</pre>

... and the list referenced by `iptables`.
<pre>
root@bcpc-vm4:~# iptables -L INPUT-monitoring -n
Chain INPUT-monitoring (1 references)
target     prot opt source               destination         
ACCEPT     all  --  0.0.0.0/0            0.0.0.0/0            match-set monitoring-clients src
</pre>


erhudy:
Tested out fine on a fresh build. Redoing the build without this PR and then will verify that cheffing over top works as expected.

erhudy:
Cheffed over existing build successfully.

PR #1018, opened 2016-04-05
Provider for creating Zabbix client auto-registration actions

kelvk:
Replaces functionality in `zabbix_api_auto_discovery.erb` which creates Zabbix client auto-registration actions. This also now permits user-defined Zabbix metadata/templates/hostgroups that each action should associate registered clients with.

For testing purpose, delete the listed actions via https://10.0.100.6/zabbix/actionconf.php?&eventsource=2 and rechef monitoring host. The actions should reappear in the UI.

### DIFF
--- a/cookbooks/bcpc/libraries/zabbix.rb
+++ b/cookbooks/bcpc/libraries/zabbix.rb
@@ -41,3 +41,13 @@ def zbx_api(auth, method, params)
 
   zbx_res['result']
 end
+
+def zbx_auth
+  params = {
+    :user     => get_config('zabbix-admin-user'),
+    :password => get_config('zabbix-admin-password')
+  }
+  auth = zbx_api(nil, 'user.login', params)
+  raise 'Zabbix authentication failed' if auth.nil?
+  auth
+end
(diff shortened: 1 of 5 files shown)


erhudy:
Testing fresh build, then will test over existing build.

erhudy:
Worked successfully on a fresh build, re-testing over an existing build.

Nitpick about the provider behavior: it indicates **(up to date)** even when it is actually re-converging the resource. The part of the provider that is performing the convergence action should be wrapped in a `converge_by` block so that Chef will correctly indicate that the resource is being re-converged (e.g., https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/providers/patch.rb#L85).

Current behavior:
```
  * bcpc_zbx_autoreg[BCPC-Bootstrap] action create (up to date)
  * bcpc_zbx_autoreg[BCPC-EphemeralWorknode] action create (up to date)
  * bcpc_zbx_autoreg[BCPC-Monitoring] action create (up to date)
  * bcpc_zbx_autoreg[BCPC-Worknode] action create (up to date)
```

erhudy:
Also it appears that Headnode is missing from the list of auto-registration actions to be performed; or at least, it was there previously and now is not.

erhudy:
Seems to work fine when cheffed over an existing build; does not attempt to recreate auto-registration actions that were created by the previous mechanism.

Leaving this open while awaiting comment on the above two issues.

kelvk:
Added missing headnode action and `converge_by` block.

erhudy:
Re-testing with changes.

PR #1019, opened 2016-04-06
Update Zabbix templates

kelvk:
This change adds the following placeholder [templates](https://10.0.100.6/zabbix/templates.php) to facilitate more granular template association with hosts in the future.

```
Template App Ceph Mon
Template App HAProxy
Template App Keepalived
Template App Rados Gateway
Template Openstack API
Template Rados Gateway API
```


erhudy:
Adds the specified new templates and does not appear to change existing templates (at least based on item/trigger counts before and after).

PR #1024, opened 2016-04-14
Wait to give nova-network time to set up tenant networks

erhudy:
This is a hackaround for an issue experienced internally when rebooting hypervisors where we were attempting to restart instances before nova-network had recreated the tenant networks. This change will delay 30 seconds per instance to be restarted to give nova-network time to get around to recreating the tenant network for that instance (nova-network recreates all tenant networks on restart starting from the lowest-numbered network and going up, a process which takes about 20-30 seconds per network).

### DIFF
--- a/bootstrap/ansible_scripts/hardware_management/restart-target-nodes.yml
+++ b/bootstrap/ansible_scripts/hardware_management/restart-target-nodes.yml
@@ -97,7 +97,7 @@
       register: instance_stopped
       until: instance_stopped.rc == 0
       ignore_errors: true
-      retries: 10
+      retries: 20
       delay: 3
       with_items: running_instances
       when: item != ""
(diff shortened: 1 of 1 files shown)


erhudy:
Reading this PR again, I forgot to update it when I found out that the strategy for determining the number of instances doesn't actually work correctly when operating across multiple hosts. Ansible seems to aggregate the number of instances in total across all hosts even when operating on a single host, so when I tried to use this for real Ansible wanted to wait about an hour and a half.

kelvk:
I was testing this and found `running_instances|length` returned the character length of the value, so a single running instance would result in 1200 seconds delay. Will re-test.

erhudy:
Oh, is that what's happening? That's obviously not correct.

PR #1027, opened 2016-04-18
Improve ephemeral test behavior regarding autobackup behavior

erhudy:
This PR does 3 things:
* switches off autobackup behavior on `lvcreate` and `lvremove` (this does not stop autobackup completely, but reduces the noise a bit)
* adds a cron job to clean up older autobackup files from  `/etc/lvm/archive` nightly
* adjusts the failure regex to specifically look for the words "read failed" instead of just "failed" to work around an issue where LVM autobackup behavior would collide with an existing autobackup and put the word "failed" in the output, even though the LV creation/deletion was successful


kelvk:
Recheffed and check works as expected.

PR #1029, opened 2016-04-20
Normalize Diamond CPU collection

kelvk:
Instead of collecting stats for every core, just collect normalized totals.

When cheffed in, the CPU non-totals (`bcpc-vm4:/opt/graphite/storage/whisper/servers/*/cpu/cpu*/*`) will no longer be updated.


erhudy:
Testing in conjunction with #1030.

erhudy:
Works as intended when added to existing build - individual core whisper files are no longer updated, only the ones under `total`. Will verify that it works properly on a fresh build and then merge.

erhudy:
Verified on fresh build, merging.

PR #1030, opened 2016-04-20
Parameterize Graphite metric whitelist/blacklist

kelvk:
Facility to selectively accept/drop metrics based on provided whitelist/blacklist regular expressions. The following snippet can be inserted into the environment to test:
```
      "graphite": {
        "max_creates_per_min": "inf",
        "use_whitelist": {
          "enabled": "True",
          "blacklist": [
            "servers\\..*\\.mysql\\.Com.*"
          ]
        }
      },
```
All other metrics should continue to be updated except for `/opt/graphite/storage/whisper/servers/bcpc-vm1/mysql/Com*`.


erhudy:
Testing in conjunction with #1029.

erhudy:
Works as specified - using the test snippet caused the `Com_*` whisper files to stop being updated. Also tested with default attributes without an issue. Will re-test on fresh build and then merge.

erhudy:
Validated on fresh build with default attributes and also modified environment file. Merging.

Issue #1023, opened 2016-04-14
Upgrade Diamond


semenovroman:
Version of Diamond that we are currently using ( https://github.com/bloomberg/chef-bcpc/blob/7f4ae8e3281f8e59678b510f6c5cb483f8e8fb84/bootstrap/shared/shared_build_bins.sh#L29 ) is leaking memory (and can consume all free RAM) in some situations. 
Issue was reported to Diamond ( https://github.com/python-diamond/Diamond/issues/84 ) and fixed ( https://github.com/python-diamond/Diamond/pull/256 ).
We should upgrade to a version of Diamond that incorporates that fix.

erhudy:
Fixed in #1034.

PR #1034, opened 2016-04-26
Upgrade Diamond to 4.0.439

kelvk:
Addresses #1023. Collection for NTP and LoadAverage's normalized metric counters are also disabled.


erhudy:
This will have strange interplay with `shared_prereqs.sh` if the repo has already been cloned from GitHub, but before the commit with the new hash was added (it will blow up when it tries to check out a nonexistent hash).

How do you want to handle this situation? Maybe `shared_prereqs.sh` could be augmented to try a `git pull` if the `python-diamond` directory already exists?

erhudy:
Also does nobody provide packages of Diamond still? Do we need to keep on doing this stuff with Git repos?

kelvk:
I have not seen any .deb and am hesitant to use the PIP package because I found it lagging fixes when I had to upgrade to the current version. I actually had the cloned directory manually removed but does sound like the right thing to do to update `shared_prereqs.sh` - will review.

kelvk:
Updated commit to perform a `git pull` before checking out. `diamond.deb` needs to be removed for the new package to be built.

erhudy:
The updated changeset does a `git pull` during the binary build phase, which is not desired as it will try to contact the Internet during build. The `git pull` should be in `shared_prereqs.sh`, which is what mirrors things from Internet to the local machine prior to the binary build.

This happens at present when you have an existing `~/.bcpc-cache` and it gets to the binary build step:
```
Cloning into 'Diamond'...
done.
Already up-to-date.
fatal: reference is not a tree: 3409b118be6af7b3ae8868297a2edf8cc7c5614c
Connection to 127.0.0.1 closed.
```

kelvk:
Updated once more. Both `diamond.deb` and download flag `python-diamond_downloaded` need to be removed to pull repo updates and rebuild the package.

erhudy:
Testing by deleting `$HOME/.bcpc-cache/build_bins_cache` and `$HOME/.bcpc-cache/python-diamond_downloaded`. Left `$HOME/.bcpc-cache/python-diamond` in place to verify that it would update the directory in place.

erhudy:
I'd like to propose that rather than using the presence of the `_downloaded` file, which is a hack that doesn't work well with Git, instead let's test for the presence of the desired target version with `git -C ${BOOTSTRAP_CACHE_DIR}/python-diamond log --pretty=format:'%H' | grep -q ${VER_DIAMOND}`, which will return 0 if the version we want is in the local tree and non-zero otherwise.

The pseudocode flow can be:
if `${BOOTSTRAP_CACHE_DIR}/python-diamond` does not exist then clone
otherwise check for the desired commit with the above command
if desired commit not found then `git pull`

erhudy:
New version of Diamond seems to work okay and transmit data to Graphite. Verifying that the bug referenced in #1023 is actually fixed is beyond the scope of this review.

What do you think of the above suggestion for managing idempotence on the Diamond repo pulls?

kelvk:
Pushed commit with suggested change.

erhudy:
Testing locally with existing build and binary cache (after removing `build_bins_cache` only) and completely fresh build, including fresh binary cache.

erhudy:
Looks like elasticsearch-head repo was accidentally removed in the latest update.

kelvk:
Sorry. I have replaced it with `clone_repo` now.

erhudy:
After the fix, this built successfully from a completely fresh build and also over top of an existing build. Will merge tomorrow.

erhudy:
Merge notes: this will require some manual intervention from the user to remove `diamond.deb` from their bin directory/binary build cache. Easiest thing to do is to remove `$HOME/.bcpc-cache/build_bins_cache` and allow the scripts to rebuild everything.

For Ansible script builds, you will want to create a new directory under your bootstrap-files location with the updated binaries.

PR #1037, opened 2016-05-09
Fixes error introduced by #1036 that causes failure to converge

erhudy:
#1036 introduced a stupid bug that will cause a convergence failure on ephemeral disk nodes due to an invalid LVM configuration being generated. Without this bugfix in place, convergence will fail due to a filter in `lvm.conf` that resembles the following:
```
filter = [ ["\"a|^/dev/md/md0$|\", "] "a|^/dev/md127$|", "r|.*|" ]
```
This bugfix adds a `#join` method call to the result of the map so that the syntactically correct form will be output:
```
filter = [ "a|^/dev/md/md0$|",  "a|^/dev/md127$|", "r|.*|" ]
```

### DIFF
--- a/cookbooks/bcpc/recipes/lvm.rb
+++ b/cookbooks/bcpc/recipes/lvm.rb
@@ -29,7 +29,7 @@ if node['bcpc']['nova']['ephemeral']
     group  'root'
     mode   00644
     variables(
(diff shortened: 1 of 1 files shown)


erhudy:
Testing locally with existing build and binary cache (after removing `build_bins_cache` only) and completely fresh build, including fresh binary cache.

kamidzi:
Had just kicked off a fresh build - slipped it in.

PR #1039, opened 2016-05-09
Rename flavors

erhudy:
This re-scales 2xlarge flavors to progress in the same cadence as the other flavors (multiples of 1 vCPU/2GB memory).

Additionally:
- specifies both extra_specs for all flavors
- explicitly declares attributes for all m1 flavors
- indented the block 1 indent inward

### DIFF
--- a/cookbooks/bcpc/libraries/utils.rb
+++ b/cookbooks/bcpc/libraries/utils.rb
@@ -267,6 +267,16 @@ def generate_vrrp_vrid()
     results.first
 end
 
+# this takes the blisteringly maddening openstack CLI JSON output of the form
+# [{"Field": "x", "Value": "y"}, ...] and turns it into a regular hash
+def openstack_json_to_hash(input)
+  input.collect {
+    |v| { v['Field'] => v['Value'] }
+  }.reduce({}) {
+    |target_hash, v| target_hash.merge(v)
+  }
+end
+
 def join_aggregate_action
   node['bcpc']['in_maintenance'] ? :depart : :member
 end
--- a/cookbooks/bcpc/providers/osflavor.rb
+++ b/cookbooks/bcpc/providers/osflavor.rb
@@ -2,7 +2,7 @@
 # Cookbook Name:: bcpc
 # Provider:: osflavor
 #
-# Copyright 2013, Bloomberg Finance L.P.
+# Copyright 2016, Bloomberg Finance L.P.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


erhudy:
There appears to be an issue with the **bcpc_osflavor** provider and it is not correctly updating the number of vCPUs on existing builds.

erhudy:
Added a changeset that modifies the provider to allow it to converge flavors when the flavor attributes change (previously it only updated the flavor if extra_specs changed).

erhudy:
Re-scoped this PR to rename the ephemeral flavors to **nondurable1**.

erhudy:
Re-scoped even wider to also rename m1 to **generic1**.

cookbooks/bcpc/libraries/utils.rb
```diff
@@ -267,6 +267,16 @@ def generate_vrrp_vrid()
     results.first
 end
 
+# this takes the blisteringly maddening openstack CLI JSON output of the form
+# [{"Field": "x", "Value": "y"}, ...] and turns it into a regular hash
+def openstack_json_to_hash(input)
```
kamidzi: IIRC  there was mention of some instances of `-f json` not including `Field` keys. Is that the case? If so, maybe getting rid of nil key is worthwhile?

cookbooks/bcpc/providers/osflavor.rb
```diff
@@ -24,61 +24,93 @@ def whyrun_supported?
...
 
 def nova_cli
-  # Note the amazing lack of consistency between openstack CLI and nova CLI when it 
-  # comes to args e.g. "--os-user-name" vs "--os-username".  
-  args =  ["nova", 
-      "--os-project-name", node['bcpc']['admin_tenant'], 
+  # Note the amazing lack of consistency between openstack CLI and nova CLI when it
+  # comes to args e.g. "--os-user-name" vs "--os-username".
+  args =  ["nova",
+      "--os-project-name", node['bcpc']['admin_tenant'],
```
kamidzi: I believe this is transposed with python-openstackclient args?

erhudy: Yeah, that looks like an oversight. `openstack

cookbooks/bcpc/libraries/utils.rb
erhudy: What about adding a check that returns the input if a Hash is passed in instead of an Array? The purpose of the method is to unwind the Array into a Hash, so if something that's already correctly formatted (or we guess it's correctly formatted because it's a Hash already) is passed in, we could just reflect that back.

PR #1042, opened 2016-05-12
Re-work patches to handle python-nova 2015.1.2 - 2015.1.4

kamidzi:
This should handle the cases. I can easily confirm 2015.1.3, 2015.1.4, but I have no easy way to test 2015.1.2 without some prolonged setup.


erhudy:
I needed to adjust the version number for python-nova 2015.1.2 to `1:2015.1.2-0ubuntu2~cloud0` in order to get the guard to successfully trigger, otherwise Chef skipped over the resource.

erhudy:
Added one additional commit to stop dpkg from getting angry at our managed `keystone.conf` when upgrading from 2015.1.2 to 2015.1.4 packages. Tested fresh builds with 2015.1.2 mirror set, 2015.1.4 mirror set, and upgrading from 2015.1.2 to 2015.1.4. @kamidzi, do you want to crush this down or merge as-is?

kamidzi:
Feel free to squash it all.

PR #1043, opened 2016-05-13
TEST FIRST - Updated default PG and added nearest power_of_2

cloudm2:


erhudy:
Tested on top of an existing build first - had no effect as expected. On a fresh build, the average PG count per pool was reduced to an eighth of what it was previously, which was also expected. Since it's easy to go up but impossible to go down and because 0.94.5 fixed the PG increase explosion issue in librbd, this seems safe and reasonable to me.

PR #1044, opened 2016-05-13
Make bcpc_cephconfig provider idempotent

erhudy:
This rewrites the cephconfig provider in the following ways:
- makes it idempotent by correctly comparing the value
- flattens out the pyramid of ifs by adding a few nexts
- drops some error handling that was not likely to ever be reached


kamidzi:
Also added test for socket filetype.

PR #1046, opened 2016-05-16
Update to most current available versions of dependency cookbooks

erhudy:
Doing some housekeeping on Chef. This updates all dependency cookbooks to the most current available versions and removes the embedded copy of the chef-client cookbook, which was done to incorporate some sort of local patch that is no longer used.

(The invasion of feature/decomposed-cookbooks into master begins here.)


kelvk:
Cheffed in fine on existing cluster. Fresh build test in progress.

PR #1050, opened 2016-05-17
Break binary files off into separate cookbook

erhudy:
- updated location in local build script
- updated Ansible deployment playbook for hardware builds
- updated resources that reference binary files

The intent of this PR is to migrate binary files that we have to keep around in a separate cookbook that exists purely to wrap those binary files.

If you want to transform an existing build, you'll need to schlep stuff around manually on the bootstrap node:
- rsync **bcpc** and **bcpc-binary-files** over to the bootstrap node
- move the contents of `cookbooks/bcpc/files/default/bins` to `cookbooks/bcpc-binary-files/files/default`
- upload all cookbooks

Re-cheffing should be a no-op.


kelvk:
Recheffed fine. Testing fresh build.

PR #1053, opened 2016-05-23
Fixes #1049: generates validated hostnames in dns_fill.py

### LINKED ISSUE #1049: dns_fill.py does not filter out all invalid characters in hostnames
Noticed while fixing issue in #1048: not all invalid characters are filtered out (e.g. it allows `[` and `]` through).

erhudy:
This fixes #1049, where dns_fill.py would not scrub all invalid characters and would produce hostnames like `[blah]`.

### DIFF
--- a/cookbooks/bcpc/files/default/dns_fill_test.py
+++ b/cookbooks/bcpc/files/default/dns_fill_test.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python
+
+from dns_fill import make_rfc1123_compliant
+import unittest
+
+
+class TestMakeRFC112Compliant(unittest.TestCase):
+    def test_already_good_hostname(self):
+        name = 'testname'
+        result = name
+        self.assertEqual(
+            make_rfc1123_compliant(name),
+            result)
+
+    def test_spaces(self):
+        name = 'this is a hostname'
+        result = 'this-is-a-hostname'
+        self.assertEqual(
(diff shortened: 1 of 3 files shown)


erhudy:
I ended up doing a lot more than just fixing #1049, but all the changes should be broken up accurately by commit for easier review.

cookbooks/bcpc/files/default/dns_fill_test.py
```diff
@@ -0,0 +1,70 @@
...
+                'extremely extremely extremely extremely extremely extremely '
+                'long hostname')
+        result = ('this-is-an-extremely-extremely-extremely-extremely-'
+                  'extremely-ex')
+        self.assertEqual(
+            make_rfc1123_compliant(name),
+            result)
+
+    def test_final_hyphen_as_63rd_character(self):
+        name = ('this is a very long hostname that ends in a hyphen '
```
kamidzi: Curious - why bother preserving the hyphen instead of just truncating? I think this has a higher probability of collision than just truncating as the maximum prefix length is shortened.

erhudy: Avoiding collisions wasn't a goal, since the inputs (OpenStack server names) can be duplicates. I understand your point, though. In actual use I suspect this is an edge case that will probably never be exercised.

PR #1057, opened 2016-05-26
Replace RabbitMQ signing key

erhudy:
See http://www.rabbitmq.com/news.html#2016-05-25T18:00:00+03:00 for their news release.


kamidzi:
Hey...
```
After this operation, 43.1 MB of additional disk space will be used.
WARNING: The following packages cannot be authenticated!
  rabbitmq-server
```
Failing my builds since yesterday...

PR #1063, opened 2016-06-09
Raise default Keystone workers to 20 from 5

erhudy:
Making this change based on production experience.


erhudy:
After discussion internally, since we will probably not have another release before Liberty and Fernet tokens, abandoning this.

Issue #490, opened 2015-02-19
Fix cookbook metadata


kamidzi:
Missing information, etc.

PR #1073, opened 2016-06-21
Add issues_url and source_url to bcpc cookbook metadata

erhudy:
Super softball PR that adds `issues_url` and `source_url` to satisfy foodcritic.


erhudy:
Resolves #490. (I think.)

kamidzi:
It appears there are some missing dependencies in the new metadata, especially those used by the Basic role. Is this intentional?

erhudy:
The dependency metadata looks out of date. I will update it to synchronize it with the current dependency cookbook versions.

erhudy:
Rebased to repair merge conflict.

PR #1075, opened 2016-06-21
Return hostname based on procfs

### LINKED ISSUE #1069: Using newer than Vagrant 1.8.1 breaks the build
https://github.com/mitchellh/vagrant/issues/7488

Vagrant 1.8.3 introduced changes to how the hostname is set that causes the addition of a host to a Nova host aggregate to fail. On master right now, this will fail silently but you won't be able to launch VMs. If #1054 is merged, the build will fail on the head node with the error
```
Failed to add host bcpc-vm1 to aggregate TLVP-1: ERROR: openstack Cannot add host bcpc-vm1 in aggregate 4: not found (HTTP 404)
```

kamidzi:
Ask the kernel first, then fall back to node attribute. Fixes #1069.

### DIFF
--- a/cookbooks/bcpc/providers/host_aggregate.rb
+++ b/cookbooks/bcpc/providers/host_aggregate.rb
@@ -32,6 +32,13 @@ def openstack_cli
            "--os-password" , get_config('keystone-admin-password')]
 end
 
+# calculate the hostname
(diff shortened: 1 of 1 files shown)


erhudy:
This converged successfully on both Vagrant 1.8.1 and 1.8.4.

PR #1076, opened 2016-06-22
Don't delete old flavors automatically in order to allow for transition

erhudy:


kamidzi:
For obvious reasons, this won't work ontop of an existing 5.13.0 build.

kamidzi:
On a fresh build:
```
# openstack flavor list
+--------------------------------------+---------------------+-------+------+-----------+-------+-----------+
| ID                                   | Name                |   RAM | Disk | Ephemeral | VCPUs | Is Public |
+--------------------------------------+---------------------+-------+------+-----------+-------+-----------+
| 0dfb85b5-e7ff-4d23-809c-259309ea41df | nondurable1.tiny    |   512 |    1 |         5 |     1 | True      |
| 1                                    | m1.tiny             |   512 |    1 |         0 |     1 | True      |
| 1e905bec-a113-484d-9c8b-67ac8a78eadc | nondurable1.small   |  2048 |   20 |        20 |     1 | True      |
| 2                                    | m1.small            |  2048 |   20 |         0 |     1 | True      |
| 27aca9d1-8eda-44ad-b05a-0dd08001ff41 | generic1.medium     |  4096 |   40 |         0 |     2 | True      |
| 2b4c4a63-97e8-4ed3-b5ba-323ff8b35718 | generic1.small      |  2048 |   20 |         0 |     1 | True      |
| 3                                    | m1.medium           |  4096 |   40 |         0 |     2 | True      |
| 4                                    | m1.large            |  8192 |   80 |         0 |     4 | True      |
| 5                                    | m1.xlarge           | 16384 |  160 |         0 |     8 | True      |
| 51a7aee1-b156-4cfb-a7d8-b2f404194977 | nondurable1.xlarge  | 16384 |   40 |       160 |     8 | True      |
| 83324456-ebbd-4863-8349-18dd25633a95 | generic1.xlarge     | 16384 |   40 |         0 |     8 | True      |
| 9c6d3490-cabf-4a4a-9b1c-0d0fd4439980 | generic1.tiny       |   512 |    1 |         0 |     1 | True      |
| b8296dcc-b176-42af-9f5f-c99ca17059c3 | nondurable1.medium  |  4096 |   40 |        40 |     2 | True      |
| d8484c19-7222-4757-b99f-bceec6a19584 | generic1.large      |  8192 |   40 |         0 |     4 | True      |
| d9800ce8-e61f-44bc-919c-2386f9345c59 | nondurable1.2xlarge | 32768 |   40 |       320 |    16 | True      |
| e9534829-dfed-407e-9677-f893a33863ec | nondurable1.large   |  8192 |   40 |        80 |     4 | True      |
| ff5b0a25-3b20-41b3-a638-39c1887eaa08 | generic1.2xlarge    | 32768 |   40 |         0 |    16 | True      |
+--------------------------------------+---------------------+-------+------+-----------+-------+-----------+
```

So a few questions
 - Should the change to 2xlarge from #1039 be applied as well?
 - Is this intended to maintain the e1, m1 flavors on a fresh build? It will not currently because of the cookbook attributes.

kamidzi:
Transitionally, here is an example simulated result:

```
# openstack flavor list
+--------------------------------------+---------------------+-------+------+-----------+-------+-----------+
| ID                                   | Name                |   RAM | Disk | Ephemeral | VCPUs | Is Public |
+--------------------------------------+---------------------+-------+------+-----------+-------+-----------+
| 077a2735-c821-4754-b8af-6a622867f183 | generic1.tiny       |   512 |    1 |         0 |     1 | True      |
| 08e69cc1-1878-4216-b63c-b733f91fa0ea | e1.2xlarge          | 32768 |   40 |       320 |     8 | True      |
| 0acba828-6768-432e-8be5-bf35d7060f91 | m1.tiny             |   512 |    5 |         0 |     1 | True      |
| 167e2d62-698e-4466-be49-c1d165b45fbf | e1.medium           |  4096 |   40 |        40 |     2 | True      |
| 1acdbe8c-0e2b-4d8a-b19d-a2c77bd0fcca | m1.large            |   512 |    5 |         0 |     1 | True      |
| 20fffbfb-6c59-47d4-bb49-36f8869cac8d | nondurable1.large   |  8192 |   40 |        80 |     4 | True      |
| 21c0996d-d98d-46b3-91c9-cc6e42564b32 | generic1.xlarge     | 16384 |   40 |         0 |     8 | True      |
| 26fd2823-69e6-40bb-bfa6-ae02921a8ab6 | e1.xlarge           | 16384 |   40 |       160 |     8 | True      |
| 28122d3b-164a-4df8-9268-6dee5a2023ed | e1.small            |  2048 |   20 |        20 |     1 | True      |
| 317151c3-7a91-4c32-9ee8-fa8cb327a9ff | nondurable1.medium  |  4096 |   40 |        40 |     2 | True      |
| 492e4480-c070-4286-afde-0bb014f10c6d | nondurable1.2xlarge | 32768 |   40 |       320 |    16 | True      |
| 7f2f75af-0d03-47d9-9bc7-3a2d09d397f3 | m1.small            |   512 |    5 |         0 |     1 | True      |
| 8af68995-f3c7-4c48-b5c6-e9679232546b | e1.tiny             |   512 |    1 |         5 |     1 | True      |
| 8df12282-34af-4ddf-93c3-50fe9f192a4b | nondurable1.tiny    |   512 |    1 |         5 |     1 | True      |
| a62a39bd-3333-4612-9cf6-938fbd1754d5 | m1.medium           |   512 |    5 |         0 |     1 | True      |
| a9f26f99-c1c3-46c7-a28c-927bff555198 | nondurable1.small   |  2048 |   20 |        20 |     1 | True      |
| aaf6f4ec-791e-4833-ad96-39887f68692a | e1.large            |  8192 |   40 |        80 |     4 | True      |
| cb7d2c1c-ade2-43da-adf1-a6831500d11a | generic1.medium     |  4096 |   40 |         0 |     2 | True      |
| d78d373a-a10e-42dd-9b84-0601e27fbc6c | nondurable1.xlarge  | 16384 |   40 |       160 |     8 | True      |
| eabb65b9-8093-46e2-ace9-87a88b5d2000 | generic1.2xlarge    | 32768 |   40 |         0 |    16 | True      |
| ec46ef46-6cef-4acd-9d99-399f9c30410c | generic1.small      |  2048 |   20 |         0 |     1 | True      |
| f7c03b94-66c8-4de8-b5a4-e9e46ba97654 | generic1.large      |  8192 |   40 |         0 |     4 | True      |
| f93e4c89-eff2-4582-8746-8fd61c5d2301 | m1.xlarge           |   512 |    5 |         0 |     1 | True      |
+--------------------------------------+---------------------+-------+------+-----------+-------+-----------+
```

erhudy:
My intent here was just to stop the auto-removal logic (so that we can transition tenants according to our own schedule) without consideration towards building both the old and new flavors. Since the flavors can be overridden via chef-bcpc-prop, my expectation is that we will maintain everything there and then update that and delete the old flavors after an appropriate transition period.

erhudy:
As far as the 2xlarge change goes, we can also do that via chef-bcpc-prop, since we'll still be able to use the improved osflavor provider that can update all aspects of flavors now, not just the extra specs.

kamidzi:
I auto-removal is fine; it's just that the candidates were hard-coded. Maybe we should just be explicit and source the list of desired flavors from the attributes, whether cookbook or environmental. Anything not in that list will just get removed. This way, overriding is fairly simple and predictable.

erhudy:
We would need to extend the provider in a pretty substantial way to be able to do that and giving it such capabilities would require the provider to have an action that accesses node attributes to get the full flavor list, obtains the full flavor list from the server, and then calls its own `:delete` action to delete unmanaged flavors. While it's possible to do, it would require a fair amount of code and possibly require converting the provider to an HWRP. In my opinion the problem that we're trying to solve is not worth a complex solution, since deleting flavors is easy.

It's the problem that comes up again and again about managing resources that aren't explicitly enumerated. If you remove a file resource from Chef, Chef doesn't know about it any more. If we remove a flavor resource from Chef, Chef also doesn't know about it any more. There's no `:delete_files_i_used_to_know_about` action on the File resource.

kamidzi:
What about just deleting the flavors in the recipe as before, but just checking what's on the system versus what's in the node attributes? I don't think the provider needs to be extended.

erhudy:
I don't understand what you're proposing.

kamidzi:
I'm suggested that instead of having a hard-coded list of things to delete, enumerate the known flavors, check that against the cookbook attributes, and delete everything that's not in the list provided by the cookbook attributes.

kamidzi:
In the flavors recipe

erhudy:
Assuming "enumerate the known flavors" means asking Nova for the list of flavors it has, the provider will still need to be extended to have a new action (`:remove_unmanaged_flavors` or some such) that is able to get all flavors, examine the node attributes to get the list of flavors that we are managing, then re-execute `:delete` actions on each of the unmanaged flavors in Nova.

kamidzi:
But the flavors were previously deleted in the recipe. Is that not still a possiblity?

erhudy:
Alternatively, if you want to put it in the recipe itself, that's writing a Ruby script that runs when the recipe is executed, which seems like an abuse of Chef to me.

erhudy:
This all seems very out of scope to me. My only goal here was to not delete some flavors so that we could delete them manually at a later date instead, not rethink everything about our flavor management.

mihalis68:
Could we safely just merge this code to remove danger of dropping commonly-expected flavors and then a) manage actual clusters explicitly using the environments and b) consider a more elegant system with the points made above later? I'm mostly anxious to patch 5.13  according to the Hippocratic Oath ("first do no harm") so it's safe to deploy.

kamidzi:
Right, my concerns a two-fold:
* Leaving it like this further diverges the testing environment from what is in production; i.e., daily builds won't have the old flavors
* Defining/overriding flavors in chef-bcpc-prop adds another override that could just as well be avoided.

PR #1077, opened 2016-06-22
Move apt-get upgrade/dist-upgrade to end of Chef run

erhudy:
This change relocates the running of `apt-get upgrade` and `apt-get dist-upgrade`, if configured, to the end of the Chef run for a node. The reason for this is so that the recipes ahead of it in the Chef run can upgrade specific targeted packages in a managed fashion and handle configuration file changes appropriately, and then packages not managed by any recipe can be upgraded en masse to bring a node into a converged state with its package mirror. With the previous location, it would upgrade all packages at the beginning of the Chef run and cause various issues (restarting services with default configuration files, Chef checksum errors, etc.).


erhudy:
Closing this because I am going to subsume it into a different PR that reorganizes facets of the Basic role.

Issue #1078, opened 2016-06-22
opscode-expander-reindexer not removed


kelvk:
After upgrading Chef server to 12.6, `opscode-expander-reindexer` keeps attempting to start but fails. This causes monitoring to raise alert when `chef-server-ctl status` returns the down component.

```
run: bookshelf: (pid 55581) 253s; run: log: (pid 51876) 963s
run: nginx: (pid 55653) 252s; run: log: (pid 52131) 961s
run: oc_bifrost: (pid 55689) 252s; run: log: (pid 51682) 967s
run: oc_id: (pid 55755) 251s; run: log: (pid 51686) 966s
run: opscode-erchef: (pid 55823) 251s; run: log: (pid 52013) 962s
run: opscode-expander: (pid 55909) 250s; run: log: (pid 51796) 964s
down: opscode-expander-reindexer: 0s, normally up, want up; run: log: (pid 1596) 1817227s
run: opscode-solr4: (pid 56113) 249s; run: log: (pid 51787) 965s
run: postgresql: (pid 56250) 249s; run: log: (pid 51574) 968s
run: rabbitmq: (pid 56267) 248s; run: log: (pid 51401) 969s
run: redis_lb: (pid 56490) 248s; run: log: (pid 52086) 961s
```

According to https://github.com/chef/chef-server/issues/846 the component is no longer in use, but not cleaned up.

kamidzi:
This is resolved, no?

erhudy:
It gets cleaned up now by the Ansible path. If you upgrade it manually on a local build, you're on your own still. Should we leave it open until it's fixed properly by an upstream Chef Server release?

kelvk:
Leaving this open until Chef Server upstream is fixed or when we get to a major release (maybe 6.0).

PR #1080, opened 2016-06-23
Chef Server upgrade fixes and cleanup

kelvk:
Besides fixing #1078, this correctly uninstalls the old `knife-acl` gem and also forces gem to only install locally so we don't wait for it to timeout on Internet-disconnected clusters.


erhudy:
Tested upgrading from the old Chef server version and it cleans up the expander service as expected. One thing to note is that it will remove that service on all versions of Chef server, including ones where it is expected to be running, which is probably not desirable if Chef server isn't actually being upgraded when the playbook is being run.

kelvk:
Chef Server 12.6 is the default in 5.13.0, so I am going with the assumption that no one will run an older version with this release.

PR #1087, opened 2016-06-29
Remove zabbix RGW checks

semenovroman:
This PR removes unused RGW zabbix checks that were generating lots of errors in keystone.

Testing: Build with at least 1 monitoring node, make sure checks are not in Zabbix and that keystone log does not have lots of error messages like `WARNING keystone.common.wsgi [-] Could not find credential: ...`


erhudy:
Because Zabbix templates are not subtractive, cheffing this onto an existing cluster will require manually removing the items deleted in these templates:
* **RGW: Number buckets** trigger
* **RGW Bucket Discovery** discovery rule

erhudy:
Testing fresh build.

erhudy:
This builds fine and chefs in fine, though you need to remove the things listed above from Zabbix to complete the change. Leaving this for now in case @kelvk has any comments.

PR #1090, opened 2016-07-06
Extend cobbler and chef bootstrap functionality to monitoring nodes

kelvk:
Monitoring nodes may not have an all-encompassing role (i.e. `BCPC-Monitoring`), hence using run list addition in `knife`.


kelvk:
Withdrawing. Merging into master will be revisited.

PR #1092, opened 2016-07-07
Extend cobbler and chef bootstrap functionality to monitoring nodes

kelvk:
Monitoring nodes may not have an all-encompassing role (i.e. BCPC-Monitoring), hence using run list addition in knife. This is for 5.x release.

### DIFF
--- a/bootstrap/ansible_scripts/bootstrap_deployment/enroll-all-nodes-in-cobbler.yml
+++ b/bootstrap/ansible_scripts/bootstrap_deployment/enroll-all-nodes-in-cobbler.yml
@@ -22,14 +22,8 @@
       fail: msg="Please provide a cluster.yaml or cluster.yml in /bcpc/deployed on the bootstrap node"
       when: not cluster_yaml_stat.stat.exists and not cluster_yml_stat.stat.exists
 
-    - name: register head nodes from cluster YAML
-      command: ./bootstrap/ansible_scripts/scripts/enroll_cobbler.py add_role -t head chdir={{ bootstrap_deployed_dir }}
-
-    - name: register OSD work nodes from cluster YAML
-      command: ./bootstrap/ansible_scripts/scripts/enroll_cobbler.py add_role -t work chdir={{ bootstrap_deployed_dir }}
-
-    - name: register ephemeral work nodes from cluster YAML
-      command: ./bootstrap/ansible_scripts/scripts/enroll_cobbler.py add_role -t work-ephemeral chdir={{ bootstrap_deployed_dir }}
+    - name: register all nodes from cluster YAML
+      command: ./bootstrap/ansible_scripts/scripts/enroll_cobbler.py add_all chdir={{ bootstrap_deployed_dir }}
 
     - name: sync Cobbler
       command: cobbler sync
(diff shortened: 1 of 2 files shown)


bootstrap/ansible_scripts/bootstrap_deployment/enroll-all-nodes-in-cobbler.yml
```diff
@@ -3,7 +3,7 @@
 # first validate local environment aspects
 - include: ../common_playbooks/validate_environment.yml
 
-- hosts: bootstraps
+- hosts: "{{ target }}"
```
erhudy: Please set this to `"{{ target | default('cluster') }}"` so that if **target** is not specified, it will automatically attempt to enroll all cluster nodes, matching the current behavior.

erhudy:
Can you rename `enroll-all-nodes-in-cobbler.yml` to `enroll-target-nodes-in-cobbler.yml` to reflect that it now takes a **target** extra var?

erhudy:
One strange thing that I have noticed is that Ansible seems to fall down sometimes if the nodes to be registered are not up. I suspect this is a problem with Ansible/Paramiko instead of the playbook, but am noting it here:
```
TASK [sync Cobbler] ************************************************************
fatal: [ansible-bcpc-vm1]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh.", "unreachable": true}
fatal: [ansible-bcpc-vm2]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh.", "unreachable": true}
changed: [ansible-bcpc-vm3 -> None]
	to retry, use: --limit @bootstrap_deployment/enroll-all-nodes-in-cobbler.retry
```
Re-running the playbook typically works, which is why I think it's a problem with Ansible.

erhudy:
`cobbler sync` execution also sometimes fails when attempting to restart services, which again I don't think is a problem with the playbook itself, but rather Cobbler. Again, re-running the playbook typically works.

kelvk:
Pushed update to enroll all defined nodes in cluster YAML unconditionally.

erhudy:
Switching to `add_all` seems to work reliably. Ansible/Paramiko seems to have some problems with delegation running from OS X.

bootstrap/ansible_scripts/software_deployment/tasks-assign-roles-to-target.yml
```diff
@@ -8,35 +8,45 @@
...
       delegate_to: "{{ groups['bootstraps'][0] }}"
-      when: "'headnodes' in group_names"
 
-    - name: Give head node admin access to Chef server (set node as admin)
-      command: /opt/opscode/embedded/bin/knife group add actor admins {{ hostfqdn.stdout }}
+    - name: Give head/monitoring node admin access to Chef server (set node as admin)
+      command: /opt/opscode/embedded/bin/knife group add client {{ hostfqdn.stdout }} admins
       delegate_to: "{{ groups['bootstraps'][0] }}"
-      when: "'headnodes' in group_names"
+      when: "'headnodes' in group_names or 'monitoring' in group_names
```
erhudy: This is missing the final `"`.

erhudy: Also, this specifies that anything with **monitoring** in its group_names should get admin, but the monitoring roles referenced above are **alerting**, **logging**, and **metrics**.

kelvk: Missing double-quote added.

The monitoring group name is intentionally retained. The inventory file should therefore look like:
```
[monitoring:children]
cluster-alerting
cluster-metrics
cluster-logging
...
```

PR #1093, opened 2016-07-07
Add custom flavor sorting in horizon

kamidzi:
This is a follow-up to #1039, #1076, to ease some confusion with flavor sorting in horzon. To test, from the root of the repo:
```
git clone -b feature/test-local-flavor-sort https://github.com/kamidzi/horizon
cp ../cookbooks/bcpc/templates/default/horizon/bcpc_flavor_sort.py.erb openstack_dashboard/test/local/settings.py
./run_tests.sh -q -P
```

Alternatively, one could login to horizon, attempt to create an instance, expand the flavor drop-down and observe the sorted order.


erhudy:
Tested this on both Kilo and Liberty builds, worked properly in both.

PR #1101, opened 2016-07-22
Ensure ipmi_devintf kernel module is inserted

erhudy:
This module is required for creating `/dev/ipmi0` on platforms with IPMI support. This avoids a situation where Graphite metrics are not initially collected. On platforms without IPMI support, the module will still insert but `/dev/ipmi0` will not be created (e.g., when using VirtualBox, so in order to verify that this has the desired effect it really needs to be tested on IPMI-supporting hardware).


kelvk:
Loads as expected. Diamond's IPMI collector is also publishing stats.

PR #1110, opened 2016-08-09
Add support for rotating Fernet keys after a max age is reached

erhudy:
This adds support for rotating Fernet keys on a schedule and distributing them to other head nodes. It also makes the resources that write keys from the data bag to disk and adds keys from the disk back to the data bag generally idempotent.

**To test:**
* set **bcpc.keystone.fernet_token_max_age_seconds** to a short value and rechef, verifying that the key rotation takes place on disk and the keys and timestamp are updated in the data bag
* if testing with multiple head nodes, verify that the rotation only takes place on the first head node to be recheffed and the other head nodes write the keys out from the data bag without executing the rotation resource
* *important if testing with multiple head nodes*: if you start to rotate keys on one head node, complete the rechefs on all head nodes before starting another rotation cycle, or you will remove at least one secondary key and make it impossible to decrypt tokens created with that key

If testing on a build where Fernet tokens have never been rotated, only keys 0 (staging) and 1 (master) will exist. The recipes are set up to accommodate 3 keys, with staging in slot 0, old master in slot 1, and master in slot 2. A key rotation should promote key 0 to the master slot at 2, demote the old master key to slot 1, generate a new staging key in slot 0, and erase any secondary tokens. If you have no secondary tokens when you rotate, the key in slot 1 will stay in slot 1 but will have been superseded by the new master key in slot 2 (Keystone always treats 0 as the staging key and the highest-indexed key as the master, with any keys in between being secondary keys).

See http://docs.openstack.org/admin-guide/keystone-fernet-token-faq.html for OpenStack documentation on Fernet tokens.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -298,6 +298,11 @@ default['bcpc']['keystone']['drivers']['trust'] = 'sql'
 default['bcpc']['keystone']['drivers']['notification'] = 'log'
 # token format
 default['bcpc']['keystone']['providers']['token'] = 'fernet'
+# enable automatic key rotation for Fernet tokens
+# (recommended but just in case you want it off)
+default['bcpc']['keystone']['rotate_fernet_tokens'] = true
+# rotate Fernet tokens after they reach this age (if enabled)
+default['bcpc']['keystone']['fernet_token_max_age_seconds'] = 7*24*60*60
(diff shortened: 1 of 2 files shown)


kelvk:
Fails to rechef existing 6.0 build.

```
  * ruby_block[write-out-fernet-keys-from-data-bag] action run
    
    Error executing action `run` on resource 'ruby_block[write-out-fernet-keys-from-data-bag]'
    
    RuntimeError
    ------------
    No config found for get_config(keystone-fernet-key-2)!!!
```

PR #1114, opened 2016-08-10
Backport of PR #1113 to stable-5.x: Move imports back to top-level in dns_fill.py

erhudy:


erhudy:
See description of #1113 for rationale.

PR #1115, opened 2016-08-11
Add option to deploy Berks packaged cookbooks via Ansible

kelvk:
Deploy cookbooks that are packaged via Berkshelf, i.e. `berks package`.

### DIFF
--- a/bootstrap/ansible_scripts/software_deployment/tasks-prepare-deployed.yml
+++ b/bootstrap/ansible_scripts/software_deployment/tasks-prepare-deployed.yml
@@ -61,6 +61,14 @@
   command: mv {{ staging_location.path }}/{{ item.cookbook }}-{{ item.version }} {{ bootstrap_deployed_dir }}/cookbooks/{{ item.cookbook }}
   with_items: "{{ internal_cookbooks }}"
 
+- name: Decompress Berks-packaged cookbooks to temporary location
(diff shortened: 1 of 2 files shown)


bootstrap/ansible_scripts/software_deployment/tasks-prepare-deployed.yml
```diff
@@ -61,6 +61,14 @@
   command: mv {{ staging_location.path }}/{{ item.cookbook }}-{{ item.version }} {{ bootstrap_deployed_dir }}/cookbooks/{{ item.cookbook }}
   with_items: "{{ internal_cookbooks }}"
 
+- name: Decompress Berks-packaged cookbooks to temporary location
+  unarchive: copy=no src={{ bootstrap_git_staging_dir }}/{{ item.cookbook }}-{{ item.version }}.tar.gz dest={{ staging_location.path }}
+  with_items: "{{ berks_packages|default([]) }}"
+
+- name: Move additional Berks-packaged cookbooks into deployment directory cookbooks
+  shell: mv {{ staging_location.path }}/cookbooks/* {{ bootstrap_deployed_dir }}/cookbooks/
```
erhudy: I was probably fussy above with internal_cookbooks when I had it copy explicit cookbook/version names over instead of doing a wildcard move like this, but can you verify that this is safe to do?

kelvk: `berks package` always prefixes tarballs with `cookbooks`, so this path is exclusively the extraction destination for such cookbooks. It does not interfere with `internal_cookbooks`.

PR #1116, opened 2016-08-12
Check /vnc_auto.html for novnc status

erhudy:
Oversight: on Kilo nova-novncproxy returns 200, on Liberty 404.


kelvk:
How about probing `/vnc_auto.html` (`HEAD` at minimum) to expect a 200?

erhudy:
That's a better idea than this. I tested HEAD vs GET and found that HEAD delayed unusually long, so I will revise this to do a GET on /vnc_auto.html.

erhudy:
Revised per @kelvk's suggestion.

PR #1117, opened 2016-08-12
Correctly detect mysql ping failure

kelvk:
Current `mysql.ping` is a decimal item in Zabbix, but the check on Zabbix agent inadvertently sends the error message in STDERR, which is incompatible with the data type. As a measure to avoid altering the check on deployed Zabbix agents, we will convert the item data type to text and evaluate the output as a string.

After cheffing, the old `mysql.ping` trigger will have to be manually removed from `BCPC-Headnode` template.


erhudy:
Confirmed that without this, no error is raised, but with it the "MySQL is not pingable from [HOST]" error is raised.

PR #1120, opened 2016-08-17
Create separate ceph health warn trigger for blocked ops

kelvk:
Blocked ops is generally a more common cause of Ceph health warnings, so create a separate trigger to clearly distinguish the fault.

After cheffing, remove the old Ceph health warn trigger via the web UI.

This can be tested by altering `ceph.health` parameter in `/etc/zabbix/zabbix_agentd.d/`, then restart `zabbix-agent`.
```
#UserParameter=ceph.health,HOME=/var/lib/zabbix ceph health | head -1
UserParameter=ceph.health,HOME=/var/lib/zabbix echo HEALTH_WARN 29 requests are blocked | head -1
```

### DIFF
--- a/cookbooks/bcpc/files/default/zabbix_bcpc_templates.xml
+++ b/cookbooks/bcpc/files/default/zabbix_bcpc_templates.xml
@@ -167,6 +167,49 @@
             </applications>
             <items>
                 <item>
+                    <name>VIP Holder</name>
+                    <type>0</type>
+                    <snmp_community/>
+                    <multiplier>0</multiplier>
+                    <snmp_oid/>
+                    <key>net.vip_holder</key>
+                    <delay>90</delay>
+                    <history>3</history>
+                    <trends>3</trends>
+                    <status>0</status>
+                    <value_type>3</value_type>
+                    <allowed_hosts/>
+                    <units/>
+                    <delta>0</delta>
+                    <snmpv3_contextname/>
+                    <snmpv3_securityname/>
+                    <snmpv3_securitylevel>0</snmpv3_securitylevel>
+                    <snmpv3_authprotocol>0</snmpv3_authprotocol>
+                    <snmpv3_authpassphrase/>
+                    <snmpv3_privprotocol>0</snmpv3_privprotocol>
+                    <snmpv3_privpassphrase/>
+                    <formula>1</formula>
+                    <delay_flex/>
+                    <params/>
+                    <ipmi_sensor/>
+                    <data_type>3</data_type>
+                    <authtype>0</authtype>
+                    <username/>
+                    <password/>
+                    <publickey/>
+                    <privatekey/>
+                    <port/>
+                    <description>Check if host is VIP holder</description>
+                    <inventory_link>0</inventory_link>
+                    <applications>
+                        <application>
+                            <name>Network interfaces</name>
+                        </application>
+                    </applications>
+                    <valuemap/>
+                    <logtimefmt/>
+                </item>
+                <item>
                     <name>Ceph Health</name>
                     <type>0</type>
                     <snmp_community/>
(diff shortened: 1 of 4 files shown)


erhudy:
This doesn't work exactly as expected, because any health status change that might be accompanied by blocked operations will be classified as blocked ops. For example, if there is a major cluster failure that causes an entire rack to drop out, we will definitely see blocked operations (among other things) while Ceph begins recovering data to other locations in the cluster. With these triggers in place, though, that would get classified as a blocked ops HEALTH_WARN, and so we'd be likely to attribute the alert to simple deep scrubbing and ignore it when in fact it's much more serious.

What we need is a trigger for blocked ops that will light up only if the single reason Ceph is in HEALTH_WARN state is because of blocked ops, with nothing else impacting. I am not sure if Zabbix's string pattern matching supports full regex; if so we could probably do so with a regex that matches the entire line `HEALTH_WARN (\d)+ requests are blocked > 32 sec`. The way Ceph outputs health information makes it very difficult to accurately parse this. I checked the JSON output format of `ceph health detail` as well, but it still embeds all the different reasons onto one line, then puts that single line into a list, which is not helpful.

erhudy:
I was able to provoke a local cluster into getting angry with blocked ops and the output of `ceph health detail` looks like this under such circumstances:
```
HEALTH_WARN 2 requests are blocked > 32 sec; 1 osds have slow requests
2 ops are blocked > 32.768 sec
2 ops are blocked > 32.768 sec on osd.1
1 osds have slow requests
```

erhudy:
To infuriate your own local Ceph cluster, a combination of RADOS benchmarking and deep scrub commands was enough to get mine reasonably angry.
```
for i in $(seq 1 5); do rados bench -p vms 240 write --no-cleanup --run-name run_$i; done
```
...will generate data in the cluster, and doing so while issuing deep scrub commands to all PGs with
```
ceph pg dump pgs | tail -n +2 | awk '{ print $1 }' | xargs -I% ceph pg deep-scrub %
```
...was enough to trigger some blocked operations locally for me.

kelvk:
Force-updated commit:
* `ceph health` is now parsed by a script on the head node to determine if it is entirely a blocked ops alert or otherwise.
* This PR will also attempt to limit alerting to only the VIP holder to reduce alert noise. Since non-VIP holders are expected to always report `OK` status, VIP jumps during a `HEALTH_WARN` could lead to alerting flapping on head nodes.

cookbooks/bcpc/files/default/zabbix_bcpc_templates.xml
```diff
@@ -167,6 +167,49 @@
...
+                    <params/>
+                    <ipmi_sensor/>
+                    <data_type>3</data_type>
+                    <authtype>0</authtype>
+                    <username/>
+                    <password/>
+                    <publickey/>
+                    <privatekey/>
+                    <port/>
+                    <description>Check if host if VIP holder</description>
```
erhudy: nit: typo here

erhudy:
When testing with this script as a substitute for `ceph health detail` (editing the Zabbix agent configuration), everything seems to work as expected. Blocked ops alone are registered as a separate alert, while any other warn condition or blocked ops plus another warn condition are registered as the existing HEALTH_WARN alert.
```
#!/bin/sh

# warning without blocked ops should produce Ceph Health Warn
#echo "HEALTH_WARN noscrub flag(s) set
#noscrub flag(s) set"

# warning with blocked ops only should produce Ceph Health Warn - Blocked ops
#echo "HEALTH_WARN 1 requests are blocked > 32 sec; 1 osds have slow requests
#1 ops are blocked > 32.768 sec
#1 ops are blocked > 32.768 sec on osd.5
#1 osds have slow requests"

# warning with blocked ops and other should produce Ceph Health Warn
echo "HEALTH_WARN 1 requests are blocked > 32 sec; 1 osds have slow requests; noscrub flag(s) set
1 ops are blocked > 32.768 sec
1 ops are blocked > 32.768 sec on osd.5
1 osds have slow requests
noscrub flag(s) set"

# no warning
#echo "HEALTH_OK"
```

I will do some additional testing, plus a new build to test converging this on existing installations, but so far so good.

PR #1124, opened 2016-08-31
Use Nova API v2.1 instead of v2

erhudy:
I initially set v2 for the Nova API due to an oversight. v2.1 is the current and preferred version. This PR also contains a change to the regex used to parse the version number out of the catalog URLs so that it can handle both integers and floats, which permitted the removal of some workaround code.

Cheffing over an existing build will automatically update the service catalog and RC files in root as necessary.

### DIFF
--- a/cookbooks/bcpc/libraries/utils.rb
+++ b/cookbooks/bcpc/libraries/utils.rb
@@ -43,7 +43,7 @@ def get_api_version(service, uri_type='public')
     fail "#{service_str} is not a valid service, please select from #{node['bcpc']['catalog'].keys.join('/')}"
   end
 
-  api_version_list = node['bcpc']['catalog'][service_str]['uris'][uri_type_str].scan(/^[^\d]*(\d+)/)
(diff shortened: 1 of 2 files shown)


cookbooks/bcpc/libraries/utils.rb
```diff
@@ -43,7 +43,7 @@ def get_api_version(service, uri_type='public')
     fail "#{service_str} is not a valid service, please select from #{node['bcpc']['catalog'].keys.join('/')}"
   end
 
-  api_version_list = node['bcpc']['catalog'][service_str]['uris'][uri_type_str].scan(/^[^\d]*(\d+)/)
+  api_version_list = node['bcpc']['catalog'][service_str]['uris'][uri_type_str].scan(/(\d+(\.\d+)?)/)
```
erhudy: The new regex could theoretically fail if there are digits elsewhere in the URI, but since there are no cases where this happens, I left off the start-of-line anchor.

PR #1125, opened 2016-09-01
Ensure br_netfilter is loaded if present (kernel 3.18+)

erhudy:
Starting in kernel 3.18, br_netfilter code was moved out to a separate module. This ensures that the module is loaded if present, if you're using a new enough kernel that it's a separate module. This fixes various weird routing issues with instances on 3.19, 4.2 or 4.4 kernels (e.g., asymmetric routing if you try to access the floating IP of an instance from an instance that only has a fixed IP).

(I know that the resource is ugly, but decided that refactoring it was secondary to just getting this working.)

### DIFF
--- a/cookbooks/bcpc/recipes/networking.rb
+++ b/cookbooks/bcpc/recipes/networking.rb
@@ -69,6 +69,21 @@ bash "enable-8021q" do
     not_if "grep -e '^8021q' /etc/modules"
 end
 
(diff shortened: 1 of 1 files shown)


kelvk:
Mind adding a condition to run this on non-monitoring providers, indicated via `node['bcpc']['monitoring']['provider']`? This is probably something that we should extend to other networking configuration bits in the future.

PR #1126, opened 2016-09-02
Extend ability to manage CPU and NUMA configurations

erhudy:
This PR does the following:
* parameterizes cpu_mode, cpu_model, and vcpu_pin_set in nova.conf (but does not modify them from their current defaults)
* installs the numactl package
* adds NUMATopologyFilter to the default list of scheduler filters (note that without the relevant extra_specs information this filter will not do anything, so this is a no-op with all current configurations)


erhudy:
Hard rebooting a VM after recheffing appears to be enough to pick up the new CPU settings. I thought it might require a shelve/unshelve, but it looks like the domain XML is rewritten on start.

erhudy:
Additional relevant information: https://libvirt.org/formatdomain.html#elementsCPUAllocation

PR #1128, opened 2016-09-02
Parameterize innodb_buffer_pool_size and innodb_buffer_pool_instances

erhudy:


erhudy:
https://dev.mysql.com/doc/refman/5.6/en/innodb-multiple-buffer-pools.html for information on configuring multiple buffer pools (not that there is enough memory by default on a local build)

kelvk:
This will break #711.

erhudy:
Oh, sorry, I did not remember to update the MySQL monitoring configuration. I will do that now.

erhudy:
I re-parameterized them as template variables and defined the default values explicitly for the monitoring setup as well.

kelvk:
Note that this change will restart MySQL instances even with default innodb parameters. I don't particularly feel that will be a problem as long as we perform a rolling restart.

PR #1131, opened 2016-09-26
Parametrize mysql parameters responsible for performance

semenovroman:
'thread_cache_size' parameter is checked for existence because if it is not defined, it is calculated dynamically by mysql.

### DIFF
--- a/cookbooks/bcpc/templates/default/wsrep.cnf.erb
+++ b/cookbooks/bcpc/templates/default/wsrep.cnf.erb
@@ -38,7 +38,7 @@ innodb_buffer_pool_size=<%= @innodb_buffer_pool_size %>
 # Number of buffer pool instances. Only takes effect if pool size is 1GB+.
 innodb_buffer_pool_instances=<%= @innodb_buffer_pool_instances %>
 
-<% if not @node['bcpc']['mysql-head']['thread_cache_size'].nil? %>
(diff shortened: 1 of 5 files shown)


erhudy:
As @kelvk pointed out to me when I PR'd some previous changes to the MySQL configuration, the wsrep.cnf template is used both by MySQL on head nodes and monitoring nodes, so the new variables will need to be passed into the template in the mysql-monitoring recipe as well.

erhudy:
Never mind, these variables all apply only to mysql-head. I will double check that everything is good with the monitoring MySQL setup as is, but @kelvk may appreciate having the extra parameters made available for monitoring anyway.

cookbooks/bcpc/templates/default/wsrep.cnf.erb
```diff
@@ -38,6 +38,11 @@ innodb_buffer_pool_size=<%= @innodb_buffer_pool_size %>
 # Number of buffer pool instances. Only takes effect if pool size is 1GB+.
 innodb_buffer_pool_instances=<%= @innodb_buffer_pool_instances %>
 
+<% if not @node['bcpc']['mysql-head']['thread_cache_size'].nil? %>
```
erhudy: This references the variable in the environment and should be changed to check whether `@thread_cache_size` is `nil` so that it applies correctly for monitoring as well.

PR #1132, opened 2016-09-27
Provision private flavor access to specific projects

dkaplan34:
The following PR provisions specific project access to private flavors.

### DIFF
--- a/cookbooks/bcpc/providers/flavor_access.rb
+++ b/cookbooks/bcpc/providers/flavor_access.rb
@@ -1,0 +1,103 @@
+#
+# Cookbook Name:: bcpc
+# Provider:: flavor_access
+#
+# Copyright 2016, Bloomberg Finance L.P.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+require 'open3'
+require 'json'
+
+def whyrun_supported?
+  true
+end
+
+def openstack_cli
+  args = ["openstack",
+          "--os-tenant-name", node['bcpc']['admin_tenant'],
+          "--os-project-name", node['bcpc']['admin_tenant'],
+          "--os-username", get_config('keystone-admin-user'),
+          "--os-compute-api-version", "2",
+          "--os-auth-url", "#{node['bcpc']['protocol']['keystone']}://openstack.#{node['bcpc']['cluster_domain']}:#{node['bcpc']['catalog']['identity']['ports']['public']}/#{node['bcpc']['catalog']['identity']['uris']['public']}/",
+          "--os-region-name", node['bcpc']['region_name'],
+          "--os-password" , get_config('keystone-admin-password')]
+
+  if get_api_version(:identity) == "3"
+    args += ["--os-project-domain-name", "default", "--os-user-domain-name", "default"]
+  end
+
+  return args
+end
+
+def nova_cli
+  # Note the amazing lack of consistency between openstack CLI and nova CLI when it
+  # comes to args e.g. "--os-user-name" vs "--os-username".
+  args = ["nova",
+          "--os-tenant-name", node['bcpc']['admin_tenant'],
+          "--os-project-name", node['bcpc']['admin_tenant'],
+          "--os-user-name", get_config('keystone-admin-user'),
+          "--os-compute-api-version", "2",
+          "--os-auth-url", "#{node['bcpc']['protocol']['keystone']}://openstack.#{node['bcpc']['cluster_domain']}:#{node['bcpc']['catalog']['identity']['ports']['public']}/#{node['bcpc']['catalog']['identity']['uris']['public']}/",
+          "--os-region-name", node['bcpc']['region_name'],
+          "--os-password" , get_config('keystone-admin-password')]
+
+  if get_api_version(:identity) == "3"
+    args += ["--os-project-domain-name", "default", "--os-user-domain-name", "default"]
+  end
+
+  return args
+end
+
+action :add do
+  cur_proj = Array.new
+  project_id_list = Array.new
+  acl_add = Array.new
+  #Verify Flavor Exists and is Private
+  stdout, stderr, status = Open3.capture3(*(nova_cli + ["flavor-access-list", "--flavor",  @new_resource.flavor_name] ))
+  if status.success?
+    stdout.split("\n").each do | fline |
+      cur_proj << (fline.split("|")[2]).strip if fline.match(/(\w+\-){4}\w+/)
+    end
+  end
+  if not status.success?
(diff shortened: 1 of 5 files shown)


kamidzi:
Where do you intend to run this? Headnodes? The VIP-holding headnode?

cookbooks/bcpc/providers/flavor_access.rb
```diff
@@ -0,0 +1,103 @@
...
+  # Verfiy Flavor in Environment file is private
+  unless cur_acl.key?(@new_resource.flavor_id)
+    Chef::Log.error "Flavor,  #{@new_resource.flavor_id} does not exist or is not private"
+    raise("FlavorID does not exist or is not marked as private.") 
+  end
+
+  acl_add = @new_resource.project_id - cur_acl[@new_resource.flavor_id]
+  
+  unless acl_add.empty?
+    converge_by("Add Flavor Access") do
```
kamidzi: It might be useful to have additional information here - flavor id and project id?

kamidzi:
This all seems to function as advertised though we might want to consider using names for the flavors and some other immutable property for project identifiers (probably not `name`); maintaining the mappings consistently across regions otherwise might prove challenging. That of course then raises the question of whether the target flavors should actually be required to exist before execution? It seems that one would know at flavor creation time whether it is intended for private use..

dkaplan34:
I am using the flavor name since the flavor ID changes if the flavor is recreated.  The flavor access command expects a project ID and would not accept a project name.  The custom flavors and tenancy access would be specified in prop with the is_public flag set to false.  This will run after flavor create recipe.  The custom flavor are per cluster.

kamidzi:
As for `flavor-access-*` commands, I agree, the project_id would have to be resolved first. But for the flavors, it seems that they are being indexed by `flavor_id` rather than flavor name, for example:

```
Recipe: bcpc::flavor_access
  * bcpc_flavor_access[apriori_case1] action add

    Error executing action `add` on resource 'bcpc_flavor_access[apriori_case1]'
<...>
    bcpc_flavor_access("apriori_case1") do
      action [:add]
      retries 0
      retry_delay 2
      declared_type :bcpc_flavor_access
      cookbook_name "bcpc"
      recipe_name "flavor_access"
      flavor_id "apriori_case1"
      project_id ["AdminTenant", "nonexistent"]
    end

root@bcpc-vm1:~# openstack flavor list --private
+--------------------------------------+---------------+-----+------+-----------+-------+-----------+
| ID                                   | Name          | RAM | Disk | Ephemeral | VCPUs | Is Public |
+--------------------------------------+---------------+-----+------+-----------+-------+-----------+
| 51df6ac1-5a89-4fae-8519-cf7893461d4d | apriori_case1 |  99 |   99 |         0 |     1 | False     |
+--------------------------------------+---------------+-----+------+-----------+-------+-----------+

**Comment by dkaplan34** [comment]
recipes updated

**Comment by kamidzi** [comment]
Seems to work as advertised:
```
$ git diff
diff --git a/environments/Test-Laptop-Vagrant.json b/environments/Test-Laptop-Vagrant.json
index 8d2e80d..e6d50d2 100644
--- a/environments/Test-Laptop-Vagrant.json
+++ b/environments/Test-Laptop-Vagrant.json
@@ -5,6 +5,13 @@
       "vagrant": {
         "nat_ip": "10.0.2.15"
       },
+      "flavor_access": {
+        "apriori_case1": {
+          "project_desc": [
+            "immutable_tag"
+          ]
+        }
+      },
       "virt_type" : "qemu",
       "enabled": {
         "logging": true,

root@bcpc-vm1:~# openstack flavor create  --private apriori_case1
+----------------------------+--------------------------------------+
| Field                      | Value                                |
+----------------------------+--------------------------------------+
| OS-FLV-DISABLED:disabled   | False                                |
| OS-FLV-EXT-DATA:ephemeral  | 0                                    |
| disk                       | 0                                    |
| id                         | bcae2e20-9271-43f5-9a25-956f0506a813 |
| name                       | apriori_case1                        |
| os-flavor-access:is_public | False                                |
| ram                        | 256                                  |
| rxtx_factor                | 1.0                                  |
| swap                       |                                      |
| vcpus                      | 1                                    |
+----------------------------+--------------------------------------+
root@bcpc-vm1:~# openstack project create --description immutable_tag somename
+-------------+----------------------------------+
| Field       | Value                            |
+-------------+----------------------------------+
| description | immutable_tag                    |
| domain_id   | default                          |
| enabled     | True                             |
| id          | daa86173a6ee4374ba740d948f8dcbf9 |
| is_domain   | False                            |
| name        | somename                         |
| parent_id   | None                             |
+-------------+----------------------------------+
root@bcpc-vm1:~# nova flavor-access-list --flavor apriori_case1
/usr/lib/python2.7/dist-packages/urllib3/util/ssl_.py:97: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
+-----------+-----------+
| Flavor_ID | Tenant_ID |
+-----------+-----------+
+-----------+-----------+
root@bcpc-vm1:~# chef-client -obcpc::flavor_access
Starting Chef Client, version 12.9.41
resolving cookbooks for run list: ["bcpc::flavor_access"]
Synchronizing Cookbooks:
  - apt (3.0.0)
  - ubuntu (1.2.0)
  - chef-client (4.5.0)
  - cron (1.7.6)
  - logrotate (1.9.2)
  - windows (1.40.0)
  - ntp (2.0.0)
  - chef_handler (1.4.0)
  - hostsfile (2.4.5)
  - concat (0.3.3)
  - bcpc (6.1.0)
  - bcpc-binary-files (6.0.0)
Installing Cookbook Gems:
Compiling Cookbooks...
Converging 1 resources
Recipe: bcpc::flavor_access
  * bcpc_flavor_access[apriori_case1] action add
    - Add Flavor Access

Running handlers:
Running handlers complete
Chef Client finished, 1/1 resources updated in 18 seconds
root@bcpc-vm1:~# nova flavor-access-list --flavor apriori_case1
/usr/lib/python2.7/dist-packages/urllib3/util/ssl_.py:97: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
+--------------------------------------+----------------------------------+
| Flavor_ID                            | Tenant_ID                        |
+--------------------------------------+----------------------------------+
| bcae2e20-9271-43f5-9a25-956f0506a813 | daa86173a6ee4374ba740d948f8dcbf9 |
+--------------------------------------+----------------------------------+
```

kamidzi:
A comment, question, and a request though: 
 - I think all users of the this resource will need to be made aware that it is assumed that project  descriptions are unique and immutable.
 - It looks like the logic won't ever _remove_ flavor access mappings? Do you see any value in adding that (not necessarily in this PR)? 
 - Could you also squash the commits?

dkaplan34:
- I believe the current policies prevent users from executing keystone commands to make changes to the project like name and description. Alternatively,  we can add a function into the existing sync script to set project description equal to our custom field or sync it against ldap  

- Due to technical limitations in chef removal was not possible w/o removing all flavor access first then recreating it with the updated ACL.  This would create a potential risk of making a flavor unavailable to users when the cluster is chef'd.

dkaplan34:
commits squashed

PR #1135, opened 2016-09-28
Keepalived quorum check

kelvk:
This updates the keepalived quorum script to really check Ceph monitor quorum in true sense of the word. Logging is included, which writes to `/var/log/ceph` and hence will be log-rotated per Ceph's log rotation policy.

This should Chef cleanly into an existing cluster without any impact. When `ceph-mon` is stopped on the VIP holder and started up (after VIP has migrated), keepalived should transition from `MASTER` to `FAULT`, then `BACKUP`.


erhudy:
Verified that shutting down the monitor on the VIP holder triggers check failure, VIP migration, and VRRP FAULT state. Verified that it transitions to BACKUP state when the monitor is brought back up. Verified the quorum log (though not rotation explicitly).

PR #1136, opened 2016-09-29
Write out Fernet keys in the data bag that are not on disk

erhudy:
This guard was incorrect; if a particular key file did not exist on disk while the keys that were on disk matched the data bag, the resource would take no action. You could get into a situation where 2 keys existed on disk instead of the usual 3, and those 2 keys matched the data bag, but the 3rd key would never be written out, which could lead to spurious authentication failures due to that Keystone instance missing a Fernet key.


erhudy:
This should be reproducible even on a single head node configuration by cheffing the cluster up, verifying that all the keys in `/etc/keystone/fernet-keys` match the keys in the data bag, then deleting one of the keys (preferably not key 0, which will cause everything to explode). Without the patch in place, it should not recreate the key, but if you then touch 0/1/2, it should write the key out if there is a corresponding data bag entry. With the patch, it should write all keys regardless.

PR #1140, opened 2016-10-12
Parameterize MySQL slow query logging

erhudy:


cookbooks/bcpc/templates/default/logrotate_mysql_slow_query.erb
```diff
@@ -0,0 +1,14 @@
...
+    compress
+    nocopytruncate
+    create 660 mysql mysql
+    size 1G
+    weekly
+    missingok
+    notifempty
+    sharedscripts
+    postrotate
+       /usr/bin/mysql -uroot -p<%= @mysql_root_password %> -e 'select @@global.long_query_time into @lqt_save; set global long_query_time=2000; select sleep(2); FLUSH LOGS; select sleep(2); set global long_query_time=@lqt_save;'
```
kelvk: What are you thoughts on disabling slow log instead?

I think we want to use `FLUSH SLOW LOGS` here.

erhudy: I copied Percona's suggestion in https://www.percona.com/blog/2013/04/18/rotating-mysql-slow-logs-safely/ without really thinking about it. I will look at rewriting the statement to turn slow query logging off, flush just the appropriate log, then turn it back on.

erhudy:
Force pushed an update that replaces the query with one that turns off slow logging, flushes just the slow log, and then turns slow logging back on.

PR #1143, opened 2016-10-14
Additional monitoring band and parameterized Diamond handlers

kelvk:
Parameterize some more configuration to support pull-based monitoring solution and additional Diamond handlers. The following can be inserted into the environment for testing. When cheffed, expect `ufw` and `iptables` to include the additional CIDRs/port and dual-written Diamond metrics (if you have a Graphite listener on port `:9109`).

```
      "diamond": {
        "handlers": {
          "diamond.handler.graphite.GraphiteHandler": {
            "host": "10.0.100.11",
            "port": "9109",
            "batch": "512",
            "timeout": "15"
          }
        }
      },
      "monitoring": {
        "vip": "10.0.100.6",
        "cidrs": [
          "1.1.1.0/24",
          "2.2.2.0/27"
        ],
        "agent_tcp_ports": ["10050", "9109"]
      },
```

### DIFF
--- a/cookbooks/bcpc/recipes/ufw.rb
+++ b/cookbooks/bcpc/recipes/ufw.rb
@@ -39,6 +39,18 @@ template "/etc/ufw/before.rules" do
     notifies :restart, "service[ufw]", :delayed
 end
 
+node['bcpc']['monitoring']['cidrs'].each do |cidr|
+  node['bcpc']['monitoring']['agent_tcp_ports'].each do |port|
+    bash "setup-allow-rules-ufw-#{cidr}-#{port}" do
+      user 'root'
+        code <<-EOH
+          ufw allow from #{cidr} to #{node['bcpc']['bootstrap']['server']} \
+          port #{port} proto tcp
+        EOH
+    end
+  end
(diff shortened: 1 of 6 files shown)


cookbooks/bcpc/recipes/ufw.rb
```diff
@@ -49,7 +61,6 @@
         ufw allow 4040/tcp
         ufw allow in on #{node['bcpc']['bootstrap']['pxe_interface']} from any port 68 to any port 67 proto udp
         ufw allow in on #{node['bcpc']['bootstrap']['pxe_interface']} from any to #{node['bcpc']['bootstrap']['server']} port tftp
-        ufw allow 10050/tcp
```
erhudy: Is there any desire to remove this global allow, since we're now doing more specific allows above?

kelvk: The parameterized allows are specifically for monitoring, hence only the removal `ufw allow 10050/tcp` rule. This could at some point be extended if it does not get replaced by a firewall cookbook.

PR #1146, opened 2016-10-18
draft network architecture, for V1 architecture (layer 2 spanned)

mihalis68:


erhudy:
Do you want the doc file to go in the root of the repository, or into docs?

mihalis68:
i had moved it to /docs but open to suggestions on that

erhudy:
Can you put it into docs so that it stays with all the other Markdown docs we've written?

PR #1147, opened 2016-10-18
Switch slow query logging off during database backups

erhudy:
When applying this to an existing cluster, rechef at least one head node/monitoring node as well as the bootstrap node so that the privileges for the backup user will be updated (otherwise it will not be able to disable/enable slow query logging).


cookbooks/bcpc/recipes/mysql-monitoring-backup.rb
```diff
@@ -28,3 +28,16 @@
       %x[MYSQL_PWD=#{get_config('mysql-monitoring-root-password')} mysql -N --batch -uroot -e 'SELECT count(user) from mysql.user where user=\"#{get_config('mysql-backup-user')}\";'].to_i < 1
     }
 end
+
+# needed to disable/enable slow query logging
+ruby_block "give-mysql-backup-user-super-privileges" do
+    block do
+        %x[ export MYSQL_PWD=#{get_config('mysql-root-password')};
```
kelvk: Password should be retrieved from `mysql-monitoring-root-password`.

erhudy: Thought I caught every instance of that, but obviously not. Force pushed fix.

PR #1151, opened 2016-10-20
Add Mitaka support

erhudy:
These changesets add support for Mitaka, as well as upgrading in-place from Liberty to Mitaka. Both Liberty and Mitaka should build fresh from this branch without any errors.


erhudy:
I will be pushing an update to this branch to fix the conflicts that arose after the merge of the Kilo cleanup PR.

erhudy:
Rebased and force pushed to resolve merge conflicts.

cookbooks/bcpc/recipes/openstack.rb
```diff
@@ -20,18 +20,36 @@
...
 include_recipe "bcpc::packages-openstack"
 
-bash 'clean-old-pyc-files' do
-  code 'find /usr/lib/python2.7/dist-packages -name \*.pyc -delete'
-  action :nothing
-end
+# python-nova is used as the canary package because it's pretty fundamental
+min_version = \
+  if is_liberty?
+    '2:12.0.5'
```
erhudy: What's dangerous about it?

erhudy: It doesn't specify that only that version can be installed if you're installing Liberty, just that it has to be at least 12.0.5 (the value here is used on line 35 below in a `dpkg --compare-versions` command, where the intended version of python-nova must be greater than or equal to this minimum version). It's possible to still get a failed build if you get a version that's brand new and our patches don't apply correctly, but that's nothing new.

PR #1158, opened 2016-11-01
Fix various minor issues in the restart playbook

erhudy:
- add option to chef the node after reboot (in response to certain
  situations where sometimes after a kernel upgrade MD devices would
  not automatically reassemble)
- calculate how long to wait for nova-network network re-creation based
  on number of running instances (read usage notes for a caveat when
  using serial > 1)
- fix some syntax complaints from Ansible 2.x


kelvk:
Added commit so `control_headnode` input by user is not required. Works as expected otherwise.

erhudy:
Seems fine to me.

PR #1160, opened 2016-11-04
Add support for configuring system timezone

erhudy:


cookbooks/bcpc/recipes/system.rb
```diff
@@ -124,3 +124,9 @@
     devices_to_converge.length.zero?
   end
 end
+
+# set timezone (note that to ensure this takes full effect, you should reboot afterward)
+bash "set timezone to #{node['bcpc']['time_zone']}" do
+  code "timedatectl set-timezone #{node['bcpc']['time_zone']}"
+  not_if { %x[timedatectl status | grep Timezone | awk '{ print $2 }'].strip == node['bcpc']['time_zone'] }
```
kamidzi: minor point, but it's not really necessary to call grep before awk:

`awk '/Timezone/{print $2}'`

kamidzi:
Locally this seems to check out.

erhudy:
Rebased and squashed commits.

erhudy:
Withdrawing this PR - will redo it later when we have the opportunity to build new clusters with UTC as the clock setting from the outset.

PR #1163, opened 2016-11-07
Remove pre-Liberty clutter

erhudy:
This changeset series removes a large amount of vestigial clutter that is no longer used or needed. It is not by any means complete (I ended up undoing some of the attempted cleanup because it turned out that it would require more extensive refactoring), but it's 5000+ lines of stuff that we don't need to keep around any more.

For existing and fresh Liberty builds, nothing in here should change any behavior. Obviously Kilo will no longer build and chef-bcpc will now require a minimum python-nova version of 12.0.5, which is the current version of Nova in Liberty.


kelvk:
Chef client auto-starts when Vagrant-built VM reboots (this is not the desired behavior), as the init startup symlinks still live around after upgrading the stock Chef client. @erhudy suggests that this can be handled in a separate PR.

At a high level, Openstack services function as expected when cheffed into an existing cluster and rebuilt fresh.

erhudy:
Submitted fix to Chef client starting up on reboot as #1168.

erhudy:
Rebased this to crush all the WIP commits.

PR #1164, opened 2016-11-07
Add ability to specify readahead value for all disks

semenovroman:
`cat /sys/class/block/sd*/bdi/read_ahead_kb` should show 512 (or whatever you specify in attributes)


erhudy:
Verified that this works on a fresh build and also being added to an existing build.

PR #1166, opened 2016-11-09
Avoid restarting Openstack daemons on policy changes

kelvk:
Openstack policy changes should take effect without having to restart daemons.


erhudy:
I was unsure whether policy changes in Horizon took effect without requiring a restart, so I tested editing Horizon's copy of the Nova policy to allow resizing and reloaded the web page - changes were there.

PR #1167, opened 2016-11-09
Rework ephemeral disk check

kelvk:
An ephemeral check that takes longer than Zabbix's timeout will result in unknown trigger state, which can escape detection if notification for unknown triggers is not enabled. Additionally, Zabbix trigger functions cannot be applied when a trigger is in unknown state. This limits our ability to use historical context for detecting consistently longer-than-expected check duration. This is reworked to run via cron and push results to Zabbix.

#1165 is included here since both PRs involve the single monolithic Zabbix template.


erhudy:
Added this in to an existing build. Doing so requires deleting the old trigger from the ephemeral template, as expected.

erhudy:
Broke the LVM device purposely and verified that -1 causes the check to fail with **Ephemeral disk failure**. I replaced the script with a stub to return various numbers and verified that **Ephemeral disk high latency** is returned after an appropriate number of checks, depending on the number.

PR #1173, opened 2016-11-21
Simplify cpupower control

erhudy:
4.4 simplifies CPU power control a lot, and we never actually used the features of the provider.


cookbooks/bcpc/recipes/cpupower.rb
```diff
@@ -17,28 +17,29 @@
...
 
-# CPU frequency governor utils
-template "/etc/default/cpufrequtils" do
-  source "cpufrequtils.erb"
-  owner "root"
-  group "root"
-  mode 00644
-  notifies :restart, "service[cpufrequtils]", :delayed
+package "cpufrequtils" do
+  action :remove
```
erhudy: Sure, why not.

PR #1177, opened 2016-11-23
Enable Zabbix server parameters and add monitoring template for S3

kelvk:
This provides a template for monitoring S3 endpoints. When the template is associated with a Zabbix host that is a S3 endpoint, Zabbix will perform `PUT`, `GET` and `DELETE`. The Zabbix host will need to have the following host macros setup, and `DNS` selected as the agent interface.
* {$S3_ACCESS_KEY}
* {$S3_ACCESS_SECRET}
* {$S3_PROTO}
* {$S3_BUCKET}
* {$S3_PATH}

Additional changes included:
* Enabling of fping to enable ping checks in Zabbix
* Parameterized `CacheSize` to accommodate larger Zabbix deployments


kelvk:
Added commit to parameterize more Zabbix cache configuration.

erhudy:
I set this up by applying the S3 template to a host and the trigger came up in unknown state. The error message is **/usr/lib/zabbix/externalscripts/s3test.sh: [2] No such file or directory**. Looks like Zabbix is looking for the script in the wrong place?

kelvk:
Amended `ExternalScripts` path. However, I am looking into why the test is getting HTTP 403, which may be a local environment issue.

kelvk:
Updated test script to always attempt to create S3 bucket before object operations.

I am able to test locally against `s3.bcpc.example.com`. Some things to check in case you have issues:
* http/https proxy env vars are unset
* `$bucket.s3.bcpc.example.com` is resolvable

kelvk:
Pushed update for real.

PR #1179, opened 2016-11-25
Add sync resource on host aggregates recipe

erhudy:


erhudy:
won't work, `/root/adminrc` is not present on work nodes, looking for a different way

PR #1180, opened 2016-11-25
Add sync resource on host aggregates recipe

erhudy:
Here's a different approach; this one augments the `nova-api` restart command to wait until it is actually answering requests before proceeding.


erhudy:
As an added bonus, this also comes with some modifications I made to the Chef bootstrap script and (accidentally) included in this PR in order to help out building under certain Internet-restricted circumstances.

PR #1181, opened 2016-11-29
Tech preview for UEFI booting

erhudy:
This revisits UEFI bootstrapping by adding the Cobbler components needed to successfully boot UEFI systems. With this in place, an appropriately configured system can be bootstrapped via UEFI by changing its profile from **bcpc_host** to **bcpc_host_uefi**. This PR does *not* include anything for enabling UEFI PXE boot on VirtualBox; we will be borrowing something from our sister project at http://github.com/bloomberg/chef-bach to do that in the future.

This should be considered a tech preview, but it should not impact any existing legacy BIOS booting (unless you were using a legacy BIOS bootloader not named `/pxelinux.0`; this has been hardcoded into the DHCP server configuration).

### DIFF
--- a/bootstrap/ansible_scripts/scripts/spawn_local_vms.sh
+++ b/bootstrap/ansible_scripts/scripts/spawn_local_vms.sh
@@ -171,6 +171,12 @@ function get_node_ip {
     echo '10.0.100.12'
   elif [[ $1 == "ansible-bcpc-vm3" ]]; then
     echo '10.0.100.13'
+  elif [[ $1 == "ansible-bcpc-vm4" ]]; then
+    echo '10.0.100.14'
+  elif [[ $1 == "ansible-bcpc-vm5" ]]; then
+    echo '10.0.100.15'
+  elif [[ $1 == "ansible-bcpc-vm6" ]]; then
+    echo '10.0.100.16'
   else
     echo '169.254.1.1'
   fi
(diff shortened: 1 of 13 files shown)


erhudy:
One note that impacts existing installations: I moved **bcpc.preseed.kernel** to **bcpc.bootstrap.preseed_kernel**, as it was the only variable in the **bcpc.preseed** namespace. Existing overrides in environments will need to be updated.

mihalis68:
I had a quick go at this, firstly to verify non-UEFI still works and had one weird failure - it wanted to create the python-diamond subdirectory again:

```
fatal: destination path '/Users/cmorgan/.bcpc-cache/python-diamond' already exists and is not an empty directory.
```

Trying again

mihalis68:
Second weird failure

```
Connection to 127.0.0.1 closed.
Uploading apt          [1.10.0]
Uploading bcpc         [5.8.1]
Uploading chef-client  [4.5.0]
ERROR: Cookbook chef-client depends on cookbooks which are not currently
ERROR: being uploaded and cannot be found on the server.
ERROR: The missing cookbook(s) are: 'cron' version '>= 1.7.0'
Connection to 127.0.0.1 closed.
```
I can obviously fix this and keep going

mihalis68:
ignore above comments, was on wrong branch (uefi-pxe-boot). Derp.

erhudy:
I encountered a few unrelated issues while testing this branch to ensure legacy BIOS PXE boot was not affected, mostly in that some of the scripts in the Ansible directory have not kept up with evolutions in the `cluster.yml` format, and when creating the VMs now via `spawn_local_vms.sh`, their network interfaces were in disconnected state by default. I'll fix these on a separate PR.

mihalis68:
I verified that the default plain vanilla vagrant path is good again after the magic sleep mentioned above.

erhudy:
Closing as UEFI booting is being deferred again and we don't need to push weird hacks into Cobbler to make it work.

PR #1186, opened 2016-12-19
Allow setting custom nofile soft/hard limits for nova-compute

erhudy:
`restart_command` override is set to ensure Upstart picks up the change in `nova-compute.conf` immediately.


mihalis68:
nova-compute-upstart is an upstream file that is modified here only with the nofile statement, is that right?

erhudy:
That's correct, and it is the same in both Kilo and Liberty, so this will be backported to stable-5.x unchanged.

mihalis68:
ok then this LGTM, but let's hear what @kelvk observes in actual testing

PR #1189, opened 2016-12-22
Restart Apache as reloads raise errors

kelvk:
Reloading of Apache sometimes result in Keystone outage of approximately 5 minutes on a local build, leading to host aggregate configuration failures (depends on Keystone). It appears that this is a known issue with Keystone/Apache per https://bugs.launchpad.net/fuel/+bug/1493353. If this PR does not sufficiently work around the issue, we may have to consider other measures mentioned in the bug report.


erhudy:
Verified this with Kelvin while he was doing the PR. With the `:reload` verb, Apache crashes somewhere in the Horizon recipe, which I verified via the HAProxy stats page (error messages about a segfault appear in `/var/log/apache2/error.log`). With the `:restart` verb, everything works as expected and the Keystone backend in HAProxy does not go down.

PR #1195, opened 2017-01-26
Add concept of diskless worknodes

erhudy:
(not really diskless since an OS disk is required, but do not provision
any sort of storage, either Ceph OSDs or ephemeral)


erhudy:
For testing, you can edit `bootstrap/shared/shared_configure_chef.sh` and change the type of bcpc-vm3 to BCPC-DisklessWorknode to have the automatic build create it automatically. This should not affect monitoring and diskless worknodes should be monitored in the same way as regular OSD nodes.

kelvk:
Cluster builds clean but node does not appear in Zabbix. In Zabbix agent configuration:
```
HostMetadata=BCPC-DisklessCompute,BCPC-DisklessWorknode,BCPC-Hardware-Virtual,BCPC-Rgwnode,Basic
```
We will either have to maintain `BCPC-Worknode` or configure monitoring to accept the new metadata.

erhudy:
So how do you want to solve it?

kelvk:
Added commit to auto-register new node type in Zabbix and created associated hostgroup. Triggers are clean except RGW/HAProxy ping, as the node is not in haproxy rotation - https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/recipes/haproxy-head.rb

kelvk:
Recheffing fixes things up. Rebuilding cluster from scratch to validate.

erhudy:
Alert cleared and everything registered after recheffing everything with that new commit.

PR #1199, opened 2017-01-31
Flag to enable/disable authority for reverse zones

kelvk:
Reverse zones are fairly static and therefore, are good candidates to be hosted on an external DNS service to reduce chef-bcpc's infrastructure scope. Disabling them involves a environment file update per below:

```
    "bcpc": {
      "pdns": {
        "authority": {
          "reverse": {
            "management": false,
            "fixed": false,
            "float": false
          }
        }
      },
      ...
```
... and deleting the affected zones from `pdns.domains` and NS/SOA (optionally, PTR) from `pdns.records`.


kelvk:
Force-pushed some corrections.

kelvk:
This approach needs reconsideration. Closing for now.

PR #1202, opened 2017-02-07
Support classless float subnet

kelvk:
Current DNS zone configuration assumes float subnets are classful (/8, /16) and this could result in incorrect DNS resolution on VMs. This attempts to remediate the assumption by dividing classless float subnet into classful zones. For testing, float subnet can be reconfigured via the environment:

```
    "bcpc": {
      "floating": {
        ...
        "cidr" : "192.168.100.0/22",
        "available_subnet" : "192.168.100.128/23"
      },
```

After recheffing, expect to see following additional zones in `pdns.domains`.

```
101.168.192.in-addr.arpa
102.168.192.in-addr.arpa
103.168.192.in-addr.arpa
```

With a query like `SELECT d.name, r.name from records r, domains d WHERE d.id = r.domain_id AND r.type = 'PTR' AND r.content like 'public-%'"`, you should see the float PTRs associated with their respective reverse zones.

### DIFF
--- a/cookbooks/bcpc/libraries/utils.rb
+++ b/cookbooks/bcpc/libraries/utils.rb
@@ -21,6 +21,7 @@ require 'openssl'
 require 'base64'
 require 'thread'
 require 'ipaddr'
+require 'ipaddress'
 
 # this method deals in strings even though API versions are numbers because
 # some API versions are integers and others are floats and it would be bad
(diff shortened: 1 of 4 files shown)


cookbooks/bcpc/libraries/utils.rb
```diff
@@ -292,28 +293,22 @@ def ceph_keygen()
...
-    #           24  => 100.168.192.in-addr.arpa (1 quad removed)
-
-    reverse_ip = cidr_ip.reverse   # adds .in-addr.arpa automatically
-    (4 - (netmask.to_i/8)).times { reverse_ip = reverse_ip.split('.')[1..-1].join('.') } # drop off element 0 each time through
-
-    return reverse_ip
+# Generates an array of classful reverse DNS zone(s) by dividing provided CIDR
+# (min /24) in the form of '1.2.3.0/24'. Allows an optional prefix override to
+# divide the CIDR by that subnet class.
+def calc_reverse_dns_zone(cidr, classless=nil)
```
kamidzi: Minor point, but this is somewhat confusing when it returns multiple values... Even if you choose to alias the method, I think that would be fine

cookbooks/bcpc/recipes/powerdns.rb
```diff
@@ -114,10 +114,11 @@
     }
   end
 
-  reverse_fixed_zone = node['bcpc']['fixed']['reverse_dns_zone'] || calc_reverse_dns_zone(node['bcpc']['fixed']['cidr'])
-  reverse_float_zone = node['bcpc']['floating']['reverse_dns_zone'] || calc_reverse_dns_zone(node['bcpc']['floating']['cidr'])
-  management_zone = calc_reverse_dns_zone(node['bcpc']['management']['cidr'])
+  reverse_fixed_zone = node['bcpc']['fixed']['reverse_dns_zone'] || calc_reverse_dns_zone(node['bcpc']['fixed']['cidr']).first
+  reverse_float_zone = node['bcpc']['floating']['reverse_dns_zone'] || calc_reverse_dns_zone(node['bcpc']['floating']['cidr'], 'classless')
+  management_zone = calc_reverse_dns_zone(node['bcpc']['management']['cidr']).first
```
kamidzi: Is the assumption here for classful management networks because the expectation is that hypervisors shouldn't require more than 254 addresses? I'm not sure that's the case.

kelvk:
That is right - it is possible we have larger management subnets. I have re-pushed to include classless management subnet support and some refactoring.

kamidzi:
I ran into the following:

```
mysql> SELECT d.name, r.name from records r, domains d WHERE d.id = r.domain_id AND r.type = 'PTR' AND r.content like 'public-%' AND r.name = '100.100.168.192.in-addr.arpa';
+--------------------------+------------------------------+
| name                     | name                         |
+--------------------------+------------------------------+
| 168.192.in-addr.arpa     | 100.100.168.192.in-addr.arpa |
| 100.168.192.in-addr.arpa | 100.100.168.192.in-addr.arpa |
+--------------------------+------------------------------+
2 rows in set (0.01 sec)

mysql> SELECT d.name, r.name from records r, domains d WHERE d.id = r.domain_id AND r.type = 'PTR' AND r.content like 'public-%' AND r.name = '0.100.168.192.in-addr.arpa';
+--------------------------+----------------------------+
| name                     | name                       |
+--------------------------+----------------------------+
| 168.192.in-addr.arpa     | 0.100.168.192.in-addr.arpa |
| 100.168.192.in-addr.arpa | 0.100.168.192.in-addr.arpa |
+--------------------------+----------------------------+
2 rows in set (0.00 sec)
```

The duplicate records, I suspect are a consequence of migrating rather than building fresh. As for the network (the broadcast address as well), are these records that we would want to serve?

kelvk:
The duplicates (including domain(s)) can be safely removed. Resolution should remain unaffected at all times.

The network/broadcast addresses issue is pre-existing and IMHO it is not something we must address right away.

PR #1205, opened 2017-02-09
if_vip exit codes

kamidzi:
It's always been annoying that running commands under `if_vip` context don't return proper exit codes.


kelvk:
There are places where no-op is expected. I think introducing non-zero exit code will abort Chef runs in such places.

kamidzi:
Do you have some output from where you encountered this? I think it's best to just fix those cases.

kelvk:
For example:
```
  * execute[set-os-quota] action run

    Error executing action `run` on resource 'execute[set-os-quota]'
    
    Expected process to exit with [0], but received '1'
    ---- Begin output of . /root/adminrc; /usr/local/bin/if_vip /usr/local/bin/set-os-quota.py ----
    STDOUT: 
    STDERR: 
    ---- End output of . /root/adminrc; /usr/local/bin/if_vip /usr/local/bin/set-os-quota.py ----
    Ran . /root/adminrc; /usr/local/bin/if_vip /usr/local/bin/set-os-quota.py returned 1
    
    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/bcpc/recipes/os-quota.rb
    
     25: execute 'set-os-quota' do
     26:     action :nothing
     27:     command '. /root/adminrc; /usr/local/bin/if_vip /usr/local/bin/set-os-quota.py'
     28: end
     29: 
```
Similar usage also exists in Zabbix server recipe.

kamidzi:
I see - I'll address those and resubmit

PR #1207, opened 2017-02-16
Small change to ignore subdirectories of cacerts dir.

kamidzi:
As per ^^


erhudy:
Without this, Vagrant blows up with
```
Message: RuntimeError: All files in $CACERTS_DIR must end in .crt due to update-ca-certificates restrictions.
```

PR #1208, opened 2017-02-16
Add empty recipes that can be overridden in prop

erhudy:


erhudy:
Please withhold merge until #1151 goes in (this has a merge conflict with that branch, and that branch is more important).

PR #1209, opened 2017-02-16
Make mitaka default openstack release

kamidzi:
https://docs.openstack.org/mitaka/


kamidzi:
Clearly this is premature... see mitaka/* branches

PR #1211, opened 2017-02-21
Add instance information to DNS TXT records

erhudy:
This extends the DNS backfiller to add TXT records for qualifying
instances that contain the instance UUID and information on the
instance's Keystone project, so that a responsible party for an
instance can be easily located without involving cloud operators.

Querying the information is as easy as using dig to get the TXT record
associated with the instance's A record (if you have the CNAME, you must
first resolve that to the A record being pointed to).

This changeset also fixes a few places in the DNS backfiller where
error-handling code had mistakes in it and would fail if invoked.

### DIFF
--- a/cookbooks/bcpc/files/default/dns_fill.py
+++ b/cookbooks/bcpc/files/default/dns_fill.py
@@ -7,6 +7,7 @@ Asks openstack about all the running instances that currently have floats
 then creates a CNAME record to point to the public-X.X.X.X username
 
 """
+import json
 import keystoneclient
 from keystoneclient import exceptions as kc_exceptions
 import MySQLdb as mdb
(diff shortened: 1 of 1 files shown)


erhudy:
Example of obtaining the information programmatically using dnspython and rehydrating it. Because of the escaped backslashes and the surrounding double quotes, you need to pass the resulting string through `json.loads` twice (or you can manually strip then out, also demonstrated below).
```
>>> from dns import resolver
>>> import json
>>> r = resolver.get_default_resolver()
>>> r.nameservers
['10.0.2.3']
>>> r.nameservers = ['10.0.100.5']
>>> q = r.query('public-192-168-100-133.bcpc.example.com', 'TXT').response.answer[0].items[0].to_text()
'"{\\"project\\": {\\"name\\": \\"test project\\", \\"description\\": \\"\\"}, \\"uuid\\": \\"9c8ded14-394f-4aca-9ef8-01b77bb4618a\\"}"'
>>> json.loads(q)
u'{"project": {"name": "test project", "description": ""}, "uuid": "9c8ded14-394f-4aca-9ef8-01b77bb4618a"}'
>>> json.loads(json.loads(q))
{u'project': {u'name': u'test project', u'description': u''}, u'uuid': u'9c8ded14-394f-4aca-9ef8-01b77bb4618a'}
>>> json.loads(q.replace('"{', '{').replace('}"', '}').replace("\\", ''))
{u'project': {u'name': u'test project', u'description': u''}, u'uuid': u'9c8ded14-394f-4aca-9ef8-01b77bb4618a'}
```

cookbooks/bcpc/files/default/dns_fill.py
```diff
@@ -82,25 +83,33 @@ def generate_records_from_vms(self):
...
-                 "CNAME",
-                 "public-" + str(address).replace(".", "-") + "." +
-                 self.config["domain"])
-                )
+            rc.append((dnsname, "CNAME", float_name))
+            project_info = {
+                'uuid': server[0],
+                'project': {
+                    'name': project.name[:40],
+                    'description': project.description[:40]
```
erhudy: These two lines are intentionally truncated to avoid problems with readability and ensure that length of the dumped JSON body does not endanger the 255-character length limit of the field.

desrod:
I've reviewed commit 9db534a0 against previous 728faea for code and function, and it is clean and offers the capability that Scott Halstead was requesting.

No issues found thus far. I'll continue to test and monitor as appropriate.

```
mysql> select * from records where type='TXT'\G
id: 66094
domain_id: 1
name: public-192-168-100-129.bcpc.example.com
type: TXT
content: {\"project\":
{\"name\": \"AdminTenant\", \"description\": \"Admin services\"}

, \"uuid\": \"a47c1b15-e011-4af7-bb3f-332be132a170\"}
ttl: 300
prio: NULL
change_date: NULL
bcpc_record_type: DYNAMIC
1 row in set (0.15 sec)
```

PR #1212, opened 2017-02-23
Clean up soft-deleted secgroup-instance associations

erhudy:


erhudy:
There are additional tables beyond the vifs table and this one that need cleaned up regularly. Closing this out in favor of a more generalized solution.

PR #1214, opened 2017-02-28
General DB cleanup script

erhudy:
This supersedes vifs_cleanup.sh to provide a more general way of cleaning up tables that fill up with soft deletes over time.

The changes to wsrep_max_ws_* are to ensure that the cleanups can always run, because the tables in question can grow to millions of rows over time.

Special callout for nova.instance_system_metadata: there is a bug in early versions of Kilo that caused rows in this table to not be marked as deleted when the instance was deleted. To backfill the deleted flag onto rows for deleted instances, run
```
UPDATE nova.instance_system_metadata SET deleted = id WHERE deleted = 0 AND instance_uuid IN (SELECT uuid FROM nova.instances WHERE deleted > 0);
```

Not running the above command will not prevent the cleanup from running, but will mean that the cleanup script will not find many valid rows to move and will leave a lot of stuff behind.

### DIFF
--- a/cookbooks/bcpc/recipes/mysql-head.rb
+++ b/cookbooks/bcpc/recipes/mysql-head.rb
@@ -118,20 +118,26 @@ template '/root/.my.cnf' do
   )
 end
 
-vifs_cleanup_script = '/usr/local/bin/vifs_cleanup.sh'
-cookbook_file vifs_cleanup_script do
-  source 'vifs_cleanup.sh'
+
+# vifs-cleanup cron job removed (replaced by db_cleanup below)
+cron 'vifs-cleanup-daily' do
+  action :delete
+end
+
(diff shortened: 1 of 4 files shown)


cookbooks/bcpc/recipes/mysql-head.rb
```diff
@@ -118,20 +118,26 @@
...
+db_cleanup_script = '/usr/local/bin/db_cleanup.sh'
+cookbook_file db_cleanup_script do
+  source 'db_cleanup.sh'
   mode   '00755'
   owner  'root'
   group  'root'
 end
 
-cron 'vifs-cleanup-daily' do
+cron 'db-cleanup-daily'
```
kelvk: `do` is missing

PR #1215, opened 2017-03-01
Add custom weigher that favors nodes with <1:1 CPU commit

erhudy:
The default weigher configuration uses RAMWeigher to decide which hosts
are the most favorable for placement. This results in severe imbalances
in clusters with machines with varying memory sizes - the scheduler will
ignore empty nodes with less total memory and pile up instances on
heavily-committed nodes that have large amounts of memory.

This weigher runs alongside other weighers and returns extremely high
weights for hosts that are undercommitted on CPU, improving the chances
of spreading load. Once 1:1 commit is reached on all hosts, higher
memory hosts will become more favorable for placement again.

This algorithm will likely be tuned more in the future and is a starting
point.


kelvk:
Works as expected

Issue #1219, opened 2017-03-10
MySQL connection errors on first Chef run


kelvk:
Recipes that depend on MySQL before VIP is up raise errors on first Chef run.

```
Recipe: bcpc::mysql-backup
  * ruby_block[initial-mysql-backup-config] action runERROR 2003 (HY000): Can't connect to MySQL server on '10.0.100.5' (113)
ERROR 2003 (HY000): Can't connect to MySQL server on '10.0.100.5' (113)
ERROR 2003 (HY000): Can't connect to MySQL server on '10.0.100.5' (113)

    - execute the ruby block initial-mysql-backup-config
  * ruby_block[give-mysql-backup-user-super-privileges] action runERROR 2003 (HY000): Can't connect to MySQL server on '10.0.100.5' (113)
ERROR 2003 (HY000): Can't connect to MySQL server on '10.0.100.5' (113)
ERROR 2003 (HY000): Can't connect to MySQL server on '10.0.100.5' (113)

Recipe: bcpc::powerdns
...
    - execute the ruby block powerdns-database-creation
  * ruby_block[powerdns-table-keystone_project] action run
ERROR 2003 (HY000): Can't connect to MySQL server on '10.0.100.5' (113)
ERROR 2003 (HY000): Can't connect to MySQL server on '10.0.100.5' (113)

    - execute the ruby block powerdns-table-keystone_project
  * ruby_block[powerdns-table-domains] action runERROR 2003 (HY000): Can't connect to MySQL server on '10.0.100.5' (113)
ERROR 2003 (HY000): Can't connect to MySQL server on '10.0.100.5' (113)

...
```

Worth noting the presence of `my.cnf`.
```
root@bcpc-vm1:~# cat ~/.my.cnf 
[client]
host=10.0.100.5
user=root
password=xyz
```

kelvk:
Fixed in #1153

PR #1153, opened 2016-10-21
Refactor storage into separate roles

erhudy:
This starts the work of diverging storage and compute by adding two new roles, CephMonitorNode and CephOSDNode, and rearranging things a bit so that you can build a storage-only cluster (which can be done by running `BOOT_GO.sh -s`).

PR #1220, opened 2017-03-13
Reorder run-list so VIP comes up before anything attempts to use MySQL

kelvk:
Addresses #1219 .


kelvk:
Commit included in #1153

PR #1221, opened 2017-03-13
Add ability to enable sysstat

semenovroman:


kelvk:
Diamond is the primary collector for system metrics and ideally what we want from sar should be in Graphite. In cases where we need sar, the demand is likely temporary so it can be enabled on machines individually.

PR #1222, opened 2017-03-14
Mitaka/services

kamidzi:
Re-work keystone and all dependent services.


erhudy:
I'd like to roll up the current state of master and a few more outstanding PRs into 6.3.1/6.4 before switching to Mitaka as the default release.

kamidzi:
Resolved conflicts introduced by laziness in 6829d3521cb62dc7a12c6b61c0e67c851e059307

kamidzi:
This was only intended for upgrade not as a fresh build. Since there is evidently some time I will test and assess feasibility of that path.

kamidzi:
Necessary rebase aside, no modifications were required to make to this work as fresh build.

kamidzi:
The above is tested as fresh build. Since master is somewhat unstable at moment, it would probably unrepresentative to test migration from latest stable.

erhudy:
Keystone settings appear to be duplicated between `default.rb` and `keystone.rb` under `attributes/`.

cookbooks/bcpc/templates/default/nova/nova.conf.erb
```diff
@@ -12,9 +12,6 @@
 #qpid_username=admin
 #qpid_password=<%=get_config('rabbitmq-password')%>
 
-# memcached settings
-memcached_servers=<%=@servers.map{|x| x['bcpc']['management']['ip'] + ":11211"}.join(",")%>
```
erhudy: I think this is still necessary - the metadata service on hypervisors talks to memcached.

cookbooks/bcpc/templates/default/nova/nova.conf.erb
```diff
@@ -30,21 +27,14 @@ metadata_host=169.254.169.254
 metadata_workers=<%=node['bcpc']['nova']['workers']%>
 dmz_cidr=169.254.169.254/32
 <% if @servers.include?(node) -%>
-enabled_apis=ec2,osapi_compute,metadata
+enabled_apis=osapi_compute,metadata
```
erhudy: If we're removing EC2 as part of Mitaka (which is fine by me), we should also remove the EC2 entry from `catalog.rb`.

erhudy: Also nova-ec2 from HAProxy.

erhudy:
A larger question: while attempting to build this with Neutron enabled, it failed because the old admin account now appears to only work for Keystone and not for other services (which makes sense, because this PR introduces individual admin accounts for each service). Neutron (and all other OpenStack services that we currently run) get configured before Nova, which introduces an ordering concern if the Nova recipes are expected to set up the Nova admin account.

Does it make sense to merge `make_config`/`get_config` into a single function that will do a get-or-create?

cookbooks/bcpc/templates/default/nova/nova.conf.erb
kamidzi: This setting was deprecated: https://docs.openstack.org/mitaka/config-reference/compute/config-options.html

cookbooks/bcpc/templates/default/nova/nova.conf.erb
kamidzi: Also deprecated: https://docs.openstack.org/releasenotes/nova/mitaka.html. I'll modify HAProxy config.

kamidzi:
I haven't messed with the neutron bits, but why not just install a separate neutron account: https://docs.openstack.org/mitaka/install-guide-ubuntu/neutron-controller-install.html?

cookbooks/bcpc/templates/default/nova/nova.conf.erb
erhudy: Ah, I see. According to the deprecation note, they added a `[cache]` section for configuring oslo.cache.

PR #1223, opened 2017-03-14
Add ability to set tc_malloc_thread_cache size for ceph

semenovroman:
This creates "/etc/default/ceph" file with "TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES=128MB" line. Next time OSD is restarted, it will import line above into its environment.

This can be deployed to cluster using ansible with:
`ansible worknodes -i inventory-$cluster -b -o -m shell -a "echo 'TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES=128MB' > /etc/default/ceph"`

For change to take effect, restart OSDs one by one, node by node with:
`ls /var/lib/ceph/osd/ | xargs -I % sh -c "echo % | cut -f 2 -d '-'" | xargs -I % sh -c "sudo stop ceph-osd id=% && sleep 5 && sudo start ceph-osd id=% && sleep 3"`

you can verify that OSD has proper tcmalloc env var with:

`pgrep ceph-osd | xargs -I % sudo cat /proc/%/environ | sed 's|\x0|\n|g' | grep -i tcmalloc`


cookbooks/bcpc/recipes/ceph-common.rb
```diff
@@ -75,10 +75,19 @@
 
 template '/etc/ceph/ceph.conf' do
     source 'ceph.conf.erb'
-    mode '0644'
+    mode 00644
     variables(:servers => get_head_nodes)
```
erhudy: Can you add a `lazy` block here around the node search, as long as we're modifying this recipe for this PR anyway?

PR #1228, opened 2017-04-04
Revise TCP buffer tune to not allocate 32MB per socket descriptor (oops)

erhudy:


kamidzi:
The changes prevent vm3 from running OOM when executing a full build.

PR #1234, opened 2017-04-14
Neutron+Calico technical preview

erhudy:
This branch presents a first look at the forthcoming network architecture for the BCPC project, replacing legacy Nova networking with Neutron and Calico. As this is an early preview, a number of compromises and omissions exist at this early stage, which are enumerated below. Nova networking will remain the default and only supported networking option for BCPC through the forthcoming 7.x release series (Trusty+Mitaka).

Currently available functionality
---
* A local cluster may be built using Neutron and Calico by setting **bcpc.enabled.neutron** to **true** in your environment file. This setting is not enabled by default in the default **Test-Laptop-Vagrant** environment. Changing this flag on an existing cluster is not supported and will result in adverse consequences.
* With **bcpc.enabled.neutron** set to **true**, the resulting cluster will make use of Neutron instead of Nova networking to provide networking services to VMs. VMs launched will receive an IP address in the 192.168.101.0/24 subnet by default, per the settings in `attributes/calico.rb` (Nova networking settings as found in **bcpc.fixed** and **bcpc.floating** are not used).
* The bootstrap node is used as a BGP route reflector via the recipe `bird-false-tor.rb`. BIRD instances running on the controller and compute nodes peer to BIRD on the bootstrap node and publish route information to the bootstrap node. Instances launched on the compute nodes will have  a corresponding static host route appear on the bootstrap node.
* BIRD's internal state may be introspected via the `birdc` command with subcommands such as `show protocols all` and `show route all`.

Unavailable/future functionality
---
* No upgrade path will be provided for migrating from Nova networking to Neutron and Calico.
* The technical preview has not been tested in a scenario with multiple control plane nodes.
* The technical preview has not been tested in a scenario with monitoring nodes.
* The bootstrap node is not configured as a router for the 192.168.101.0/24 subnet. Instances are not able to route off their hypervisor or communicate with instances on other hypervisors. (In a real installation, a hypervisor's default route will redirect traffic intended for other instances up to the TOR, which will route it appropriately based on the information from BGP peering sessions with all hypervisors connected to it. In the local build installation, the BGP route reflector and the default gateway differ.)
* ~~Communicating with instances from the bootstrap does not work.~~ Communicating with instances from the bootstrap works fine, but the default security group must be updated. The default security group created by Neutron allows ingress traffic only from other instances in the same security group and does not allow it from arbitrary CIDRs. Adding additional rules (e.g., 22/TCP from 0.0.0.0/0) works as expected and allows you to SSH from the bootstrap node
* Floating IP functionality has not been added.
* Work has not yet been done to convert to a routed rack architecture (switching to Pacemaker for HA, rewriting deployment scripts, changing how rack locations are calculated and managed, adding logic for automatically locating and establishing BGP peering with the TOR).


erhudy:
One additional note: in order to avoid populating run lists with Neutron-only recipes (causing possible confusion), all new recipes are brought in via `include_recipe` directives in other recipes. This is not desirable and will be redone when Nova networking support is removed.

negronjl:
Reviewed this.  +1

PR #1244, opened 2017-05-31
Add logging and use unique filenames for S3 check

kelvk:
This should ease debugging when the check fails.


dsvua:
looks good to me.

PR #1245, opened 2017-06-06
Initial wave of bash cleanup PRs for chef-bcpc

desrod:
Here's the first of my PRs for chef-bcpc, doing some cleanup on the bash pieces, securing the variables, optimizing the loop and array logic, fixing some common issues where found. Will look at the logic flow once these are clean and build without any issues. Also looking into optimizing the packer + vagrant provisioning to speed up that process.


mihalis68:
@kamidzi would you be able to review this?

images/packer/build-bootstrap.sh
```diff
@@ -1,7 +1,20 @@
...
-cp ${KEYFILE} ../..
+keyfile="bootstrap_chef.id_rsa"
+rm "${keyfile}"*
+ssh-keygen -N "" -f "$keyfile"
+cp "${keyfile}".pub files
+cp "${keyfile}" ../..
+
+u_series=trusty
+u_url="http://archive.ubuntu.com/ubuntu/dists/$u_series-updates/main/installer-amd64/current/images"
+u_checksum=$(curl --progress -s -f -L $u_url/SHA256SUMS | awk -v f=./netboot/mini.iso '$2 == f { print $1 }')
```
kamidzi: Is this intended run on the host executing the bcpc build or on one of the the built VMs?

desrod: This runs on the host building out the environment. What I've done here, is always ensure that the image pulled remotely, matches the remote checksum, by fetching that checksum from the same location as the iso. This avoids having to edit the checksum embedded into the build_bootstrap.sh script itself, every time the upstream iso changes.

bootstrap/shared/shared_build_bins.sh
```diff
@@ -24,130 +24,130 @@ pushd $BUILD_DEST
...
 FILES="pyrabbit-1.0.1.tar.gz $FILES"
 
 # Build requests-aws package
-if [ ! -f python-requests-aws_${VER_REQUESTS_AWS}_all.deb ]; then
-  cp -v $FILECACHE_MOUNT_POINT/requests-aws-${VER_REQUESTS_AWS}.tar.gz .
-  tar zxf requests-aws-${VER_REQUESTS_AWS}.tar.gz
-  fpm -s python -t deb -f requests-aws-${VER_REQUESTS_AWS}/setup.py
-  rm -rf requests-aws-${VER_REQUESTS_AWS}.tar.gz requests-aws-${VER_REQUESTS_AWS}
+if [[ ! -f "python-requests-aws_${VER_REQUESTS_AWS}_all.deb" ]]; then
+  cp -v "$FILECACHE_MOUNT_POINT/requests-aws-${VER_REQUESTS_AWS}.tar.gz" .
```
desrod: I don't think this was one of my commits, and I don't see any subshells being called here, but there are some fundamental problems with its quoting constructs. I'll fix that up in another wave of updates/PRs.

bootstrap/shared/shared_build_bins.sh
```diff
@@ -24,130 +24,130 @@ pushd $BUILD_DEST
...
+if [[ -f diamond.deb ]]; then
+    if [[ "$(dpkg-deb -f diamond.deb Version | cut -b1)" -le 3 ]]; then
         rm -f diamond.deb
     fi
 fi
 # Make the diamond package
-if [ ! -f diamond.deb ]; then
-  git clone $FILECACHE_MOUNT_POINT/python-diamond Diamond
+if [[ ! -f diamond.deb ]]; then
+  (git clone "$FILECACHE_MOUNT_POINT/python-diamond" Diamond
```
desrod: It's redundant in this case, but I just pushed another commit that fixes this and some other bits of shared_build_bins.sh. Follow-up PR coming shortly.

PR #1255, opened 2017-07-26
Upgrade rally version

srsakhamuri:
This PR upgrades the rally version and allows us to use the latest scenarios and less buggy version of Rally.

Partially fixes: RDTIBCC-1040


cookbooks/bcpc/recipes/rally-deployments.rb
```diff
@@ -69,8 +56,9 @@
       user rally_user
       code <<-EOH
           # Another approach is to use --fromenv...
-          rally deployment create --file="#{infile}" --name=#{version}
-          unlink "#{infile}"
+          source #{rally_venv_dir}/bin/activate
+          rally deployment create --filename="#{infile}" --name=#{version}
+          sudo unlink "#{infile}"
```
kamidzi: Why is sudo needed if `infile` owner is `rally_user`?

cookbooks/bcpc/recipes/rally-deployments.rb
```diff
@@ -21,13 +21,14 @@
...
+rally_venv_dir = "/home/#{rally_user}/rally/venv"
 
 # This json file represents the current deployment of OpenStack. It is read in a later section and then
 # the information from the json file is created in Rally's database to be used for tests.
 KEYSTONE_API_VERSIONS.each do |version|
   infile = File.join(Chef::Config[:file_cache_path], "rally-existing-#{version}.json")
-  template "/var/chef/cache/rally-existing-#{version}.json" do
-      user 'root'
+  template "#{infile}" do
+      user rally_user
```
kamidzi: This is dangerous unless mode is set accordingly: https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/templates/default/rally.existing.json.erb#L17

cookbooks/bcpc/recipes/rally-deployments.rb
srsakhamuri: I will check again, I too wasn't sure why I was getting permission denied error while testing, without sudo here

cookbooks/bcpc/recipes/rally-deployments.rb
```diff
@@ -69,11 +59,19 @@
...
+          source #{rally_venv_dir}/bin/activate
+          rally deployment create --filename="#{infile}" --name=#{version}
+          sudo unlink "#{infile}"
       EOH
   end
 end
 
+bash "copy scenario files if available" do
+  code <<-EOH
+    rsync -avz --delete /chef-bcpc-files/rally/bb-rally #{rally_install_dir} || true
```
srsakhamuri: The hardcoding of the cache dir doesn't seem to work, Need to find another way.

cookbooks/bcpc/recipes/rally-run.rb
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+rally_user = node['bcpc']['rally']['user']
+rally_home_dir = node['etc']['passwd'][rally_user]['dir']
+rally_install_dir = "#{rally_home_dir}/rally"
+rally_venv_dir = "#{rally_install_dir}/venv"
+rally_deployment = "v2.0"
```
kamidzi: v2 should only really be used if necessary; is there reason to prefer over v3?

bootstrap/shared/shared_prereqs.sh
```diff
@@ -220,21 +220,12 @@ download_file "calicoctl-${VER_CALICOCTL}" "https://github.com/projectcalico/cal
...
-for package in "${!rally_packages[@]}"; do 
-        pypi_dir="${rally_packages[package]:0:1}"
-        base_package_name="${rally_packages[package]%-*}"
-        download_file "rally/${rally_packages[package]}" "$pypi_url/$pypi_dir/$base_package_name/${rally_packages[package]}"
-done
-
-# ..and for the one package that has to be a special snowflake and not fit into
-# the above scheme because of capitalization weirdness
-download_file rally/prettytable-0.7.2.tar.gz "$pypi_url/P/PrettyTable/prettytable-0.7.2.tar.gz"
+if [ $RALLY_SCENARIOS_DIR ]; then
```
srsakhamuri: Yes make sense will change that

cookbooks/bcpc/recipes/rally-run.rb
srsakhamuri: No reason, I will switch to v3.

cookbooks/bcpc/recipes/rally-deployments.rb
srsakhamuri: BOOTSTRAP_CACHE_DIR only set on the deployer node, may not be available on bootstrap node I think, but I am not 100% sure, will check it. Otherwise I think I can use chef's remote_directory construct.

cookbooks/bcpc/recipes/rally-deployments.rb
srsakhamuri: Will change the mode to more restrictive

cookbooks/bcpc/recipes/rally-deployments.rb
srsakhamuri: The /var/chef/cache is owned by root so forcing us to use sudo here

bootstrap/shared/shared_prereqs.sh
```diff
@@ -220,12 +220,14 @@ download_file "calicoctl-${VER_CALICOCTL}" "https://github.com/projectcalico/cal
 ####################################################################
 # Obtain packages for Rally. There are a lot.
 # for future reference, to install files from this cache use pip install --no-index -f file:///path/to/files rally
-if [ $RALLY_SCENARIOS_DIR ]; then
-  rsync -avz --delete --exclude=.git $RALLY_SCENARIOS_DIR $BOOTSTRAP_CACHE_DIR/rally/
-else
+if [ -z $RALLY_SCENARIOS_DIR ]; then
```
kamidzi: same issue here with the quoting

cookbooks/bcpc/recipes/rally-deployments.rb
srsakhamuri: Done

bootstrap/shared/shared_prereqs.sh
srsakhamuri: Done

kamidzi:
Ran into the following:

```
$ time ./BOOT_GO.sh
 ____   ____ ____   ____
| __ ) / ___|  _ \ / ___|
|  _ \| |   | |_) | |
| |_) | |___|  __/| |___
|____/ \____|_|    \____|

BCPC Vagrant BootstrapV2 0.2

Building a BCPC cluster of type converged

Testing configured proxies...
Not copying RALLY scenario files, set RALLY_SCENARIOS_DIR pointing to your custom rally scenarios repo
Pause 5 seconds....
DEPRECATION: pip install --download has been deprecated and will be removed in the future. Pip now has a download command that should be used instead.
Collecting virtualenv
  Downloading virtualenv-15.1.0-py2.py3-none-any.whl (1.8MB)
    100% |████████████████████████████████| 1.8MB 325kB/s
  Saved /Users/bcpc/.bcpc-cache/rally/virtualenv-15.1.0-py2.py3-none-any.whl
Collecting rally==0.9.1
  Downloading rally-0.9.1-py2-none-any.whl (754kB)
    100% |████████████████████████████████| 757kB 757kB/s
  Saved /Users/bcpc/.bcpc-cache/rally/rally-0.9.1-py2-none-any.whl
Collecting python-manilaclient<=1.14.0,>=1.12.0 (from rally==0.9.1)
  Downloading python_manilaclient-1.14.0-py2.py3-none-any.whl (263kB)
    100% |████████████████████████████████| 266kB 1.7MB/s
  Saved /Users/bcpc/.bcpc-cache/rally/python_manilaclient-1.14.0-py2.py3-none-any.whl
Collecting python-saharaclient==1.1.0 (from rally==0.9.1)
  Downloading python_saharaclient-1.1.0-py2.py3-none-any.whl (103kB)
    100% |████████████████████████████████| 106kB 3.3MB/s
  Saved /Users/bcpc/.bcpc-cache/rally/python_saharaclient-1.1.0-py2.py3-none-any.whl
Collecting oslo.log<=3.21.0,>=3.11.0 (from rally==0.9.1)
  Downloading oslo.log-3.21.0-py2.py3-none-any.whl (59kB)
    100% |████████████████████████████████| 61kB 3.8MB/s
  Saved /Users/bcpc/.bcpc-cache/rally/oslo.log-3.21.0-py2.py3-none-any.whl
Collecting PrettyTable<0.8,>=0.7.1 (from rally==0.9.1)
  Downloading prettytable-0.7.2.zip
  Saved /Users/bcpc/.bcpc-cache/rally/prettytable-0.7.2.zip
Could not import setuptools which is required to install from a source distribution.
  File "/usr/local/lib/python2.7/site-packages/pip/req/req_install.py", line 375, in setup_py
    import setuptools  # noqa
ImportError: No module named setuptools

You are using pip version 8.0.3, however version 9.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

real    0m10.121s
user    0m0.797s
sys     0m0.473s

$ git rev-parse --abbrev-ref HEAD && git log -1 HEAD
upgrade_rally_version
commit dc13f9f3b49341f4f732bb7ccde39aa4e16d8f57
Author: Srinivas Sakhamuri <ssakhamuri@bloomberg.net>
Date:   Thu Aug 3 11:37:38 2017 -0400

    Fix quoting in few more places
```

Anything special that needs to be set/happen?

kamidzi:
Also, even accounting for missing deps with something like the following:
```
From 58b99c6c251758bbfc4906da1b8a68572f75f11b Mon Sep 17 00:00:00 2001
Date: Thu, 3 Aug 2017 13:15:04 -0400
Subject: [PATCH] Specify and ensure rally build requirements.

---
 bootstrap/shared/shared_prereqs.sh | 8 ++++++++
 1 file changed, 8 insertions(+)

diff --git a/bootstrap/shared/shared_prereqs.sh b/bootstrap/shared/shared_prereqs.sh
index b1fd873..58ff70d 100755
--- a/bootstrap/shared/shared_prereqs.sh
+++ b/bootstrap/shared/shared_prereqs.sh
@@ -220,6 +220,12 @@ download_file "calicoctl-${VER_CALICOCTL}" "https://github.com/projectcalico/cal
 ####################################################################
 # Obtain packages for Rally. There are a lot.
 # for future reference, to install files from this cache use pip install --no-index -f file:///path/to/files rally
+virtualenv "$BOOTSTRAP_CACHE_DIR/rally"
+. "$BOOTSTRAP_CACHE_DIR/rally/bin/activate"
+cat > "$BOOTSTRAP_CACHE_DIR/rally/rally-build-requirements.txt" <<EoF
+pip >= 9.0.1
+setuptools
+EoF
 if [ -z "$RALLY_SCENARIOS_DIR" ]; then
   echo "Not copying RALLY scenario files, set RALLY_SCENARIOS_DIR pointing to your custom rally scenarios repo"
   echo "Pause 5 seconds...."
@@ -229,5 +235,7 @@ else
   RALLY_SCENARIOS_DIR="${RALLY_SCENARIOS_DIR%/}"
   rsync -avz --delete --exclude=.git "$RALLY_SCENARIOS_DIR" "$BOOTSTRAP_CACHE_DIR/rally/"
 fi
+pip install --upgrade -r "$BOOTSTRAP_CACHE_DIR/rally/rally-build-requirements.txt"
 pip install -d "$BOOTSTRAP_CACHE_DIR/rally/" --no-binary cffi,cryptography virtualenv rally==0.9.1
 pip install -d "$BOOTSTRAP_CACHE_DIR/rally/" pyinotify || true
+deactivate
--
2.8.4

```

There's

```
++ pip install -d /Users/bcpc/.bcpc-cache/rally/ pyinotify
DEPRECATION: pip install --download has been deprecated and will be removed in the future. Pip now has a download command that should be used instead.
Collecting pyinotify
  Downloading pyinotify-0.9.6.tar.gz (60kB)
    100% |████████████████████████████████| 61kB 3.4MB/s
  Saved /Users/bcpc/.bcpc-cache/rally/pyinotify-0.9.6.tar.gz
    Complete output from command python setup.py egg_info:
    inotify is not available on macosx-10.10-x86_64
```

srsakhamuri:
we don't need a specific pip version to setup rally, pip install and various CLI options that works across the board. Also setuptools is available on all our boxes.

srsakhamuri:
@furlongm basically this is one way to move to latest Rally, we could install manually (like before) by downloading each package from pypi. Since pip provides similar functionality of downloading, it make sense to use it. The pip/setuptools dependency is only on the local developer machine or the jump host, shouldn't be a big deal. The reason for moving to latest rally is it gives newer scenarios for testing that suites the new releases, I think we should keep moving Rally versions whenever newer release available, or pick a stable enough git commit, because last release is 0.9.1 (apr 2017) and already has few bugs working with v3 identity API.

@desrod that's neat, I will change to that.

furlongm:
A few comments:

1. It might be better to install rally into a virtualenv as per @kamidzi 's commit above? That takes care of requiring pip and setuptools, as virtualenv automatically installs them (and this approach doesn't require an extra manual step to install pip and setuptools).

2. It looks like the pyinotify requirement is already taken care of, in the oslo.log requirements.txt:

   https://github.com/openstack/oslo.log/blob/master/requirements.txt#L13

i.e. pyinotify should install automatically on linux, but not on windows, macosx (darwin) or sunos.

3. Currently this commit installs rally to the local host, rather than anywhere in the cluster. Would it be better to live on the bootstrap node? In that case, it is installed in a shared well-known location, as opposed to being installed on someone's personal desktop, or in their home dir on a jump host. Also the bootstrap node is guaranteed to have access to the APIs whereas other hosts are not.

srsakhamuri:
Will rework this PR

PR #1257, opened 2017-08-04
rom-o-matic.net moved to .eu

desrod:
Grave build failure, as rom-o-matic.net has been moved to a .eu domain upstream.


mihalis68:
I requested this change be cherry-picked into a minimal build-fixing PR, hence will merge this.

Issue #790, opened 2015-08-23
Rewrite resources to use Chef file_cache_path instead of /tmp


erhudy:
See Foodcritic rule http://acrmp.github.io/foodcritic/#FC013; basically we should just need to replace instances where files are written to `/tmp` with `Chef::Config[:file_cache_path]`, mainly so that if a box gets rebooted we're not writing out a bunch of binary files and evaluated templates to `/tmp`.

PR #1261, opened 2017-08-15
Upgrade rally version take #2

srsakhamuri:
- On laptop it uses the proxy env variables if available to access Pypi external repo
- On internal clusters this will depend on internal mirrors that need to be pre-setup

Install log: https://gist.githubusercontent.com/srsakhamuri/343bff36ec22f84b37ad72658cc4ea56/raw/0a1d1f8f3ef5a3064575bbc7c1abee32f157c1ad/gistfile1.txt

Fixes: RDTIBCC-1106


cookbooks/bcpc/recipes/rally-deployments.rb
```diff
@@ -21,17 +21,29 @@
 
 KEYSTONE_API_VERSIONS = %w{ v2.0 v3 }
 rally_user = node['bcpc']['rally']['user']
+rally_home_dir = node['etc']['passwd'][rally_user]['dir']
+rally_install_dir = "#{rally_home_dir}/rally"
+rally_venv_dir = "#{rally_install_dir}/venv"
+rally_deployment = "v2.0"
+
+# Have the image file ready for image based tests
+cookbook_file "/tmp/rally/cirros-0.3.4-x86_64-disk.img" do
```
kamidzi: Using the Chef cache dir might be a better bet. See #790.

desrod: +1 on this one; we should strive to keep the build clean and contained where possible with the ```Chef::Config[:file_cache_path]``` construct.

srsakhamuri: This image file is needed in the tests that we run later, and need to be a well known path. I find the chef cache path to be different between vagrant and prod deployments.

One option is to write the chef-cache path to a config file and access from the tests.

kamidzi: I'm not sure I've had the same experience with varying paths:
```
Ohai2u vagrant@bcpc-vm1.hypervisor-bcpc.example.com!
chef (12.9.41)> Chef::Config[:file_cache_path]
 => "/var/chef/cache"
```
```
Ohai2u operations@****.bloomberg.com!
chef (12.9.41)> Chef::Config[:file_cache_path]
 => "/var/chef/cache"
```

srsakhamuri: Ok, I was thinking of /bcpc on prod vs /chef-bcpc-files on vagrant, thanks for clarification

the file is in 
/var/chef/cache/cookbooks/bcpc-binary-files/files/default/cirros-0.3.4-x86_64-disk.img
is this ok if i refer this path directly from Rally tests?

kamidzi: I think that's fine, though the best situation would be if such a path were configurable with the above set as the default.

srsakhamuri: yep, thinking the same, will configure this as default. No hard-coding in the tests

cookbooks/bcpc/recipes/rally.rb
```diff
@@ -17,39 +16,78 @@
...
+     libpq-dev
+     libxml2-dev
+     libxslt1-dev
+}.each do |pkg|
+    package pkg do
+        action :upgrade
+    end
+end
+
+bash 'create virtual env for rally' do
```
srsakhamuri: Looks like the re-cheffing missing some of the pip configuration happening in Vagrantfile, if I add this manually the migration from master works. i.e. 
$ cat /home/vagrant/.pip/pip.conf
[global]
proxy = https://proxy.bloomberg.com:81

kamidzi: Same issue after re-provisioning to get pip config
```
    Error executing action `run` on resource 'bash[create virtual env for rally]'

    Expected process to exit with [0], but received '127'
    ---- Begin output of "bash"  "/tmp/chef-script20170818-32549-1b2dhim" ----
    STDOUT: Downloading/unpacking virtualenv
      Cannot fetch index base URL https://pypi.python.org/simple/
      Could not find any downloads that satisfy the requirement virtualenv
    Cleaning up...
    No distributions at all found for virtualenv
    STDERR: mkdir: cannot create directory �/home/vagrant/rally�: File exists
      File "/usr/bin/pip", line 9, in <module>
        load_entry_point('pip==1.5.4', 'console_scripts', 'pip')()
      File "/usr/lib/python2.7/dist-packages/pip/__init__.py", line 235, in main
        return command.main(cmd_args)
      File "/usr/lib/python2.7/dist-packages/pip/basecommand.py", line 163, in main
        log_file_fp = open_logfile(log_file_fn, 'w')
      File "/usr/lib/python2.7/dist-packages/pip/basecommand.py", line 194, in open_logfile
        os.makedirs(dirname)
      File "/usr/lib/python2.7/os.py", line 157, in makedirs
        mkdir(name, mode)
    OSError: [Errno 13] Permission denied: '/root/.pip'
    /tmp/chef-script20170818-32549-1b2dhim: line 3: /home/vagrant/.local/bin/virtualenv: No such file or directory
    ---- End output of "bash"  "/tmp/chef-script20170818-32549-1b2dhim" ----
    Ran "bash"  "/tmp/chef-script20170818-32549-1b2dhim" returned
```

srsakhamuri: sudo -i loads the roots environment, it is causing some issues. Seems like the proper way to invoke chef-client is sudo chef-client according to chef docs. https://docs.chef.io/ctl_chef_client.html

But the latest commit should fix the issue since we don't need to run the rally recipe on bootstrap immediately.

kamidzi: The chef docs claim either `su` which will modify HOME,SHELL, and maybe USER; or `sudo`, which without `-i`, doesn't modify HOME, amongst other things.

srsakhamuri: They seem to recommend running chef-client with sudo without -i, using -i may cause issues like what we are seeing. 

Do you still think there are changes to be made to this PR? Since we don't run rally recipe part of the bootstrap run initially.

kamidzi:
Minor point though; it should be noted somewhere that the `rally_user` is required to have sudo privs.

PR #1267, opened 2017-08-29
add screen recipe

furlongm:
Installs screen and adds a basic configuration file

### DIFF
--- a/cookbooks/bcpc-extra/files/default/screenrc
+++ b/cookbooks/bcpc-extra/files/default/screenrc
@@ -1,0 +1,44 @@
+# This is the global screenrc file. Handle with care.
+
+startup_message off
+defscrollback 30720
+
+# Turn visual bell on
+vbell on
+vbell_msg "   Wuff  ----  Wuff!!  "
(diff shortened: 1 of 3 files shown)


srsakhamuri:
Good idea.

Just a thought, shall we keep these auxiliary recipes into their own cookbook, without adding to bcpc cookbook. Rally etc also will be moved to this new cookbook.

kamidzi:
Agree with @srsakhamuri on separate cookbook in this case.

furlongm:
Some of the current PRs (kexec, etckeeper) are very useful to have from the very beginning of the chef-bcpc build (especially with vagrant).

Two questions:

1) To clarify, you mean another cookbook within chef-bcpc?
2) What are the benefits of moving to another cookbook vs remaining in the bcpc cookbook?

kamidzi:
1. Doesn't really matter, just not the bcpc cookbook.
2. The bcpc cookbook is currently bloated and is/has been a nightmare to maintain. Decomposing grants options for the future.

PR #1268, opened 2017-08-29
add kexec recipe

furlongm:
Installs kexec-tools and configures kexec to load on reboot. This makes reboots take less than 30 seconds as the kernel is loaded directly from shutdown sequence by kexec.


furlongm:
No output as per PR description

PR #1271, opened 2017-08-31
add bcpc-extra cookbook

furlongm:


srsakhamuri:
Do we need the CHANGELOG? or is it auto generated some how

furlongm:
Autogenerated and foodcritic complains if it's not there. Usually the correct thing is to have a changelog.

PR #1280, opened 2017-09-21
use rfc1918 ip addresses for default values

furlongm:
Tested and confirmed working locally.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -140,8 +140,8 @@ default['bcpc']['management']['firewall_tcp_ports'] = [
 default['bcpc']['metadata']['ip'] = "169.254.169.254"
 
 default['bcpc']['storage']['netmask'] = "255.255.255.0"
-default['bcpc']['storage']['cidr'] = "100.100.0.0/24"
-default['bcpc']['storage']['gateway'] = "100.100.0.1"
+default['bcpc']['storage']['cidr'] = "10.17.2.0/24"
+default['bcpc']['storage']['gateway'] = "10.17.2.1"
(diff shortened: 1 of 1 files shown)


kamidzi:
Please provide some output for your testing.

cookbooks/bcpc/attributes/default.rb
```diff
@@ -157,9 +157,9 @@
 # to be set properly
 default['bcpc']['floating']['interface-parent'] = nil
 
-default['bcpc']['fixed']['cidr'] = "1.127.0.0/16"
+default['bcpc']['fixed']['cidr'] = "10.17.16.0/20"
```
kamidzi: Is there reason to use a smaller network?

furlongm: The larger the network, the greater the risk of overlap in rfc1918 address space. We don't actually need a /20 for testing (which is where this value is used mostly).

kelvk:
This will impact population of DNS records for fixed network. See #1202. It still is assumed to be classful.

furlongm:
"Current DNS zone configuration assumes float subnets are classful (/8, /16) and this could result in incorrect DNS resolution on VMs." - is this something that could be fixed in the DNS zone configuration?

kelvk:
Yes. https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/recipes/powerdns.rb#L121 is probably the place to start.

bcpc:
@furlongm #1280 rebased on master with singlehead: is successful! click https://cloud-infra.jaas.dev.bloomberg.com/job/bcpc/job/jenkins/job/master/177/

PR #1282, opened 2017-09-25
Run rally only when opted with VERIFY_INSTALL env var

srsakhamuri:
* Currently defaults to False


bootstrap/config/bootstrap_config.sh.defaults
```diff
@@ -85,3 +85,5 @@ export FILECACHE_MOUNT_POINT=/chef-bcpc-files
 # Leave empty for sensible defaults in shared_prereqs.sh
 export CHEF_CLIENT_DEB=
 export CHEF_SERVER_DEB=
+
+export VERIFY_CLUSTER=0
```
kamidzi: So the default is to run rally?

srsakhamuri: Fixed it, I was thinking in python while writing the if...

PR #1283, opened 2017-09-29
First-pass cleanup of current shell provisioning

desrod:
Cleaning up some of the leftover breadcrumbs from previous closed PRs. 

This changeset ensures that vars are correctly quoted ($REPO_ROOT w/spaces, for example) and that we're not leaving any vars exposed to splitting or masking return values. Each file includes its own commit details.

I've run two full builds and this should merge cleanly with bloomberg/master.


mihalis68:
Tested this pulled into master, built clean, looks fine

PR #1285, opened 2017-10-02
More minor cleanup, rewrites of binary and apt package version detection

desrod:
Small fixes to add some error checking, exit codes, rewrote the checks for required binaries, apt package version comparison, pulling things closer to the bash style guide with regard to variable naming conventions and syntax. Merges clean with bloomberg/master at the time of this PR.

### DIFF
--- a/bootstrap/shared/shared_configure_chef.sh
+++ b/bootstrap/shared/shared_configure_chef.sh
@@ -55,7 +55,8 @@ do_on_node vm-bootstrap "$CHEF_SERVER_INSTALL_CMD \
 # configure knife on the bootstrap node and perform a knife bootstrap to create the bootstrap node in Chef
 echo "Configuring Knife on bootstrap node..."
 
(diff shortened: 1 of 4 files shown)


mihalis68:
I'm in favor of no tabs and/or using that tool. Letting Marcus and David finish that convo, but otherwise ok to merge.

mihalis68:
@desrod @furlongm did you finish the debate about tabs/spaces?

desrod:
I'm fine either way; I'm starting to get used to Python, where I have to adhere to the constructs already present in the codebase. If I have a choice, as long as it's consistent across the project, I'm all for using spaces instead of tabs. 

https://www.jwz.org/doc/tabs-vs-spaces.html

furlongm:
I don't mind really, it was more of a "nit" comment. My preference is for tabs but the PR is already approved and should just be merged if there is not a follow-up PR to convert to spaces. Don't want to get into a tabs-vs-spaces debate!

PR #1286, opened 2017-10-04
add postfix recipe

furlongm:


mihalis68:
I'd like this to be optional. Can be another part of the "enabled" structure in the attributes. I would disable postfix on throwaway test cluster builds that will never have a long life

furlongm:
Makes sense, disabled by default in latest commit.

PR #1288, opened 2017-10-06
Fix proxy tests on linux 

kamidzi:
If building on linux (maybe Mac OS X as well?) behind a proxy, the build is likely to fail with something like

```
 ____   ____ ____   ____
| __ ) / ___|  _ \ / ___|
|  _ \| |   | |_) | |
| |_) | |___|  __/| |___
|____/ \____|_|    \____|

BCPC Vagrant BootstrapV2 0.2

Building a BCPC cluster of type converged

Testing configured proxies...
Error: proxy http://intrusive.proxy.example.com:8082 non-functional for HTTPS requests
```

This fixes such that similar such output does not appear.

### DIFF
--- a/bootstrap/shared/shared_prereqs.sh
+++ b/bootstrap/shared/shared_prereqs.sh
@@ -44,7 +44,18 @@ vbox_url="http://download.virtualbox.org/virtualbox"
 # List of binary versions to download
 source "$REPO_ROOT/bootstrap/config/build_bins_versions.sh"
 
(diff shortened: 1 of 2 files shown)


kamidzi:
One minor issue is that for two functions, the variable tests are performed each time as opposed to conditionally defining the function.

furlongm:
Note that curl on OS X El Capitan is too old. The following is needed:

```
brew install curl ; echo 'export PATH="/usr/local/opt/curl/bin:$PATH"' >> ~/.bash_profile
```

PR #1292, opened 2017-10-19
remove ephemerals

furlongm:


srsakhamuri:
Can you add the output of the run, with output from `nova aggregate-list`

furlongm:
```
root@bcpc-vm1:~# openstack aggregate list
+----+-----------------------+-----------------------+
| ID | Name                  | Availability Zone     |
+----+-----------------------+-----------------------+
|  1 | general_compute       | None                  |
|  2 | maintenance           | None                  |
|  3 | Test-Laptop-Vagrant-1 | Test-Laptop-Vagrant-1 |
+----+-----------------------+-----------------------+
```

furlongm:
```
Everything looks like it's been installed successfully!

Access the BCPC landing page at https://10.0.100.5
Use these users and passwords to access the different resources:

Horizon: admin / VjacQMZchCYevJBieynQ
HAProxy: haproxy / 4bMVhF86MvkthKaAnBR3
RabbitMQ: guest / 4CMcZcfNd1DlEIA1KAZm

Here are a few additional passwords:
System root password: BxiX09PCik3gxM7EPpan
MySQL root password: q_iS68xEOzjqWxveUxpj

Thanks for using BCPC!
Finished in 2795 seconds (0h:46m:35s)
```

kamidzi:
I suspect this will leave nondurable flavors hanging around. Should it be the case, I'm not sure what the effect will be if one of them is selected for a new instance.

mihalis68:
On our operational clusters we have no ephemeral hosts deployed any more, so such a choice would fail with a no host error (from memory). We can manually delete them if that becomes an issue.

PR #1296, opened 2017-10-23
Revert "Fix proxy tests on linux "

### LINKED ISSUE #1294: Proxy test failing
Build fails after implementing #1288. I explicitly reverted it and build succeeds.

```
BCPC Vagrant BootstrapV2 0.2
Testing configured proxies...
Error: proxy http://someproxy.corporate.domain:3128 non-functional for HTTPS requests
```

kelvk:
Reverts bloomberg/chef-bcpc#1288 to fix #1294. Build is expected to succeed on a Linux build machine with preinstalled internal certificates if it is behind a proxy.

### DIFF
--- a/bootstrap/shared/shared_prereqs.sh
+++ b/bootstrap/shared/shared_prereqs.sh
@@ -44,18 +44,7 @@ vbox_url="http://download.virtualbox.org/virtualbox"
 # List of binary versions to download
 source "$REPO_ROOT/bootstrap/config/build_bins_versions.sh"
 
-# This is obscure to prevent cert dir test each time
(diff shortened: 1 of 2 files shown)


kamidzi:
For a bit more context, this was reverted due to previously undetected incompatibilities with curl tls libraries.  See original issue, but essentially curl behaves differently based upon whether it is linked against gnutls; on xenial this is the default, whereas it is openssl on trusty.

Another approach would be to simply append the system capath as well:
```
... --capath <custom_path>:/etc/ssl/certs
```

though it was decided to let the system handle the certificates.

PR #1299, opened 2017-10-31
remove i386 foreign arch from chef client

furlongm:


srsakhamuri:
Can you add a Jira or some context around why this change is needed? Otherwise LGTM

furlongm:
Multiarch support seems to be enabled on the physical installations, this adds both i386 and amd64 package sources. chef/apt get confused when i386 packages don't exist for corresponding amd64 packages, and the chef run fails. This happens inconsistently, but definitely happens. This patchset removes the inconsistency and ensures the system is amd64 only.

PR #1310, opened 2017-12-15
Add unique internal network names.

kamidzi:
Currently, since the network names are hard-coded, it's not possible to have multiple bcpc builds on the same host (on Linux -- OS X probably suffers the same issue, but it is untested), due to Virtualbox internal network permissions. See:

 - https://www.virtualbox.org/manual/ch06.html#network_internal
 - https://www.vagrantup.com/docs/virtualbox/networking.html

Attempting multiple builds _even under different user ids_ would result in error such as:

```
$ time ./BOOT_GO.sh
<...>
==> bootstrap: Booting VM...
There was an error while executing `VBoxManage`, a CLI used by Vagrant
for controlling VirtualBox. The command and stderr is shown below.

Command: ["startvm", "468c136e-23bd-4485-b7a0-e513e129ecd0", "--type", "headless"]

Stderr: VBoxManage: error: Failed to open/create the internal network 'management' (VERR_PERMISSION_DENIED).
VBoxManage: error: Failed to attach the network LUN (VERR_PERMISSION_DENIED)
VBoxManage: error: Details: code NS_ERROR_FAILURE (0x80004005), component ConsoleWrap, interface IConsole
```

This patch fixes the issue for multiple builds on same host under different user ids

### DIFF
--- a/bootstrap/config/examples/singlehead.networks.json
+++ b/bootstrap/config/examples/singlehead.networks.json
@@ -1,0 +1,31 @@
+{
+  "cluster_name": "Test-Laptop-Vagrant",
+  "nodes": {
+    "bcpc-dev-bootstrap": {
+      "ip_address": "10.0.100.3",
+      "gateway": "10.0.2.2",
+      "chef_role": "BCPC-Bootstrap"
+    },
+    "bcpc-dev-r1n1": {
+      "ip_address": "10.0.100.11",
+      "gateway": "10.0.2.2",
+      "chef_role": "BCPC-Headnode"
+    },
+    "bcpc-dev-r1n2": {
+      "ip_address": "10.0.100.12",
+      "gateway": "10.0.2.2",
+      "chef_role": "BCPC-Worknode"
+    }
+  },
+  "networks":{
+    "management":{
+      "name": "management-from-config"
+    },
+    "storage":{
+      "name": "storage-from-config"
+    },
+    "tenant":{
+      "name": "tenant-from-config"
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


kamidzi:
Spoken offline with @kelvk: there is an outstanding request for ability to specify network names apriori, or least have them generated in a stable, reproducible way.

kamidzi:
To demonstrate name generation
```
$ cd chef-bcpc/bootstrap/vagrant_scripts
$ irb
irb(main):001:0>  d = 'Vagrantfile.d' ; Dir.foreach(d){|f| p=File.expand_path(File.join d,f) ; load p if File.ftype(p) == 'file' }
=> nil
irb(main):002:0> generate_network_name 'management'
=> "management-ad0831bd131df8036bc4f76703e489d081092caa4869645102bd6aafa940c5a0"
irb(main):003:0> generate_network_name 'storage'
=> "storage-47e92eec0dac3b962fe51baa1316fb93ab08a90f814382d49f5b908055f895ac"
irb(main):004:0> generate_network_name 'foo'
=> "foo-6de56efb2fa758e11686e33b6ce89a7f1681ee36877a0ad4d9bad81596fe19c9"
```

^^^ Note the `generate_network_name` does not care whether the label is a member of `%w[ management storage tenant]`. Also note that `cluster_name` is not factored into the id generation.

Also, demonstrating that the correct name is used; i.e, sourced from config or generated, is a bit more involved due to deeply coupled nature of `Vagrantfile`. Try add a `networks` section to the cluster config as per `examples/singlehead.networks.json`.

kamidzi:
Not entirely happy with namespacing. Will re-factor.

PR #1315, opened 2017-12-20
remove ceph radosgw

furlongm:


furlongm:
Manual removal of pools can be done using:

```
pools=".rgw .rgw.control .rgw.gc .rgw.root .users.email .users .usage .log .intent-log .rgw.buckets .rgw.buckets.index .rgw.buckets.extra .users.uid"
for i in ${pools} ; do rados rmpool ${i} ${i} --yes-i-really-really-mean-it ; done
```

PR #1321, opened 2017-12-22
Use Keystone v3 for Nova functional check

kelvk:
`check nova` currently fails as it uses v2 authentication. Updated to use v3 and some `flake8` cleanup.

When applied, the check should be able to build an instance successfully within 60 seconds.

```
root@bcpc-dev-r1n1:~# check -f timeonly nova
11.8811299801
```

### DIFF
--- a/cookbooks/bcpc/templates/default/checks/default.yml.erb
+++ b/cookbooks/bcpc/templates/default/checks/default.yml.erb
@@ -1,6 +1,8 @@
 ---
 
-tenant: <%=@node['bcpc']['keystone']['admin_tenant']%>
-user: '<%="#{get_config('keystone-admin-user')}"%>'
+project_name: <%=@node['bcpc']['keystone']['admin_tenant']%>
+username: '<%="#{get_config('keystone-admin-user')}"%>'
 password: '<%="#{get_config('keystone-admin-password')}"%>'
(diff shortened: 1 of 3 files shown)


cookbooks/bcpc/templates/default/checks/default.yml.erb
```diff
@@ -1,6 +1,8 @@
 ---
 
-tenant: <%=@node['bcpc']['keystone']['admin_tenant']%>
-user: '<%="#{get_config('keystone-admin-user')}"%>'
+project_name: <%=@node['bcpc']['keystone']['admin_tenant']%>
+username: '<%="#{get_config('keystone-admin-user')}"%>'
 password: '<%="#{get_config('keystone-admin-password')}"%>'
-auth_url: https://<%="#{node[:bcpc][:management][:vip]}"%>:5000/v2.0/
+user_domain_name: '<%="#{get_config('keystone-admin-user-domain')}"%>'
+project_domain_name: '<%=@node['bcpc']['keystone']['default_domain']%>'
```
furlongm: keystone-admin-project-domain doesn't exist? should it?

kelvk:
Pushed fix to migration branch - https://github.com/bloomberg/chef-bcpc/commit/57ff10a5e98e0803c9e59f77eb0beadb154e37c8

PR #1324, opened 2018-01-08
Add default domain for horizon

srsakhamuri:
- Specify the domain we want to use for Horizon
- Use `domains` hash to extract default domain

### DIFF
--- a/cookbooks/bcpc/attributes/keystone.rb
+++ b/cookbooks/bcpc/attributes/keystone.rb
@@ -69,8 +69,6 @@ default['bcpc']['keystone']['admin']['enable_admin_project'] = true
 default['bcpc']['keystone']['service_project']['name'] = 'service'
 default['bcpc']['keystone']['service_project']['domain'] = 'local'
 
-default['bcpc']['keystone']['default_domain'] = 'default'
-
 # LDAP credentials used by Keystone
 default['bcpc']['ldap']['admin_user'] = nil
 default['bcpc']['ldap']['admin_pass'] = nil
--- a/cookbooks/bcpc/templates/default/horizon/horizon.local_settings.py.erb
+++ b/cookbooks/bcpc/templates/default/horizon/horizon.local_settings.py.erb
@@ -59,7 +59,7 @@ OPENSTACK_API_VERSIONS = {
 
 # Overrides the default domain used when running on single-domain model
 # with Keystone V3. All entities will be created in the default domain.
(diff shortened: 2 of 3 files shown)


srsakhamuri:
```
    - update content in file /etc/openstack-dashboard/local_settings.py from 3be084 to 998ee8
    --- /etc/openstack-dashboard/local_settings.py	2018-01-08 19:51:13.462586797 +0000
    +++ /etc/openstack-dashboard/.chef-local_settings.py20180108-294294-6n6m95	2018-01-08 19:51:41.937477959 +0000
    @@ -59,6 +59,7 @@

     # Overrides the default domain used when running on single-domain model
     # with Keystone V3. All entities will be created in the default domain.
    +OPENSTACK_KEYSTONE_DEFAULT_DOMAIN = local
```


Without this set horizon assumes default domain which is not available in Vanilla Vagrant builds, and causes 401 errors
screen shot 2018-01-08 at 3 08 52 pm

cookbooks/bcpc/attributes/keystone.rb
```diff
@@ -91,3 +89,6 @@
     }
   }
 }
+
+# set the default domain to the last domain defined, assumes we specify service domain in the beginning
+default['bcpc']['keystone']['default_domain'] = default['bcpc']['keystone']['domains'].to_a.last[0]
```
kamidzi: https://ruby-doc.org/core-2.2.0/Hash.html#method-i-keys ?

srsakhamuri:
Adding extra domain seems not very straight forward, We have `domains` dict in attributes, and we derive admin_config from values from `ldap` dictionary. which are used to enable admin user/project etc. We need a bit of refactoring in order to add two `sql` driver based domains. And it felt it will complicate things a bit.

https://github.com/bloomberg/chef-bcpc/blob/feature/l-to-m-migration/cookbooks/bcpc/recipes/keystone.rb#L533-L644

One way we can simplify things is to call `sql` domain `default` instead of `local`. But it can be confusing for us later.

cookbooks/bcpc/attributes/keystone.rb
```diff
@@ -91,3 +89,6 @@
     }
   }
 }
+
+# set the default domain to the last domain defined, assumes we specify service domain in the beginning
+default['bcpc']['keystone']['default_domain'] = default['bcpc']['keystone']['domains'].keys[-1]
```
furlongm: Could we instead override the default domain in the vagrant environment?

PR #1327, opened 2018-01-11
added enable flag for rally to control installation

justinjpacheco:


srsakhamuri:
https://github.com/bloomberg/chef-bcpc/commit/201f91500b9d78c3b84add6319127966e5254844

Looks similar to the above, may be back-port this one.

justinjpacheco:
@srsakhamuri ah, that is a better implementation. i agree, this should be back-ported.

justinjpacheco:
i've updated my branch to only include the cherry-picked contents of 201f915 as per @srsakhamuri recommendation

PR #1328, opened 2018-01-18
migration from liberty to mitaka

furlongm:


srsakhamuri:
Tested with multi-node deployment on local vagrant. It is working as expected, with one small change to the environment file. 

Make sure to use `Default`, not `default` in the domains dict in the environment file since else where we expect capitalized domain name.

PR #1330, opened 2018-01-24
Deprecate unnecessary/redundant vagrant script

desrod:
Since we're checking for needed programs/binaries elsewhere, and the vagrant_test.sh wasn't doing anything intelligent except checking that the binary existed, we can deprecate that, remove the check from BOOT.sh, and fold it into the list of other bins we need for bootstrap. Also rewrote the check for binaries.


bcpc:
@desrod multihead on master: is successful! click https://cloud-infra.jaas.dev.bloomberg.com/job/bcpc/job/jenkins/job/master/168/

PR #1338, opened 2018-03-20
fix: test timeout must exceed OSD heartbeat grace

bryanapperson:
The default osd heartbeat grace value in ceph is 20s this must be higher otherwise we will get false positives.

### DIFF
--- a/cookbooks/bcpc/files/default/zabbix_s3test.sh
+++ b/cookbooks/bcpc/files/default/zabbix_s3test.sh
@@ -16,7 +16,7 @@ filename="$script-$(hostname -s)-$fqdn-$start_time"
 file_content="$(uuidgen) $date"
 content_type='application/x-compressed-tar'
 acl="x-amz-acl:private"
-curl_cmd='curl -q -f --max-time 10'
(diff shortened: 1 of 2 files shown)


kelvk:
https://github.com/bloomberg/chef-bcpc/blob/master/cookbooks/bcpc/attributes/monitoring.rb#L124 needs to be updated to `25` as well.

bryanapperson:
@kelvk - Thanks. Made that change as well.

cookbooks/bcpc/files/default/zabbix_s3test.sh
```diff
@@ -16,7 +16,7 @@ filename="$script-$(hostname -s)-$fqdn-$start_time"
 file_content="$(uuidgen) $date"
 content_type='application/x-compressed-tar'
 acl="x-amz-acl:private"
-curl_cmd='curl -q -f --max-time 10'
+curl_cmd='curl -q -f --max-time 25'
```
furlongm: Should this file be converted to a template that uses default['bcpc']['zabbix']['agentd_timeout'] instead of hard-coding 25?

PR #1345, opened 2018-04-13
added modprobe.d/nf_conntrack.conf

justinjpacheco:
used to store nf_conntrack module changes that need to persist across reboots

### DIFF
--- a/cookbooks/bcpc/recipes/system.rb
+++ b/cookbooks/bcpc/recipes/system.rb
@@ -83,11 +83,23 @@ template "/etc/udev/rules.d/99-readahead.rules" do
     mode 00644
 end
 
-ruby_block "set-nf_conntrack-hashsize" do
-    block do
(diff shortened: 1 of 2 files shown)


cookbooks/bcpc/recipes/system.rb
```diff
@@ -83,11 +83,23 @@
...
-    end
-    not_if { system "grep -q ^$((#{node['bcpc']['system']['parameters']['net.nf_conntrack_max']}/8))$ /sys/module/nf_conntrack/parameters/hashsize" }
+begin
+  sys_params = node['bcpc']['system']['parameters']
+  nf_conntrack_max = sys_params['net.nf_conntrack_max']
+  hashsize = nf_conntrack_max / 8
+
+  template '/etc/modprobe.d/nf_conntrack.conf' do
+    source 'modprobe.d/nf_conntrack.conf.erb'
+    variables(
```
furlongm: Will this result in nf_conntrack.conf that doesn't actually contain the nf_conntrack param? That might be confusing. Maybe rename to avoid confusion?

furlongm: My bad, thought it was a sysctl file, not a modprobe file.

PR #1368, opened 2018-06-29
general cleanup and additional refactoring

justinjpacheco:
- added the following to role:node
    - apport
    - cpupower
    - etckeeper
    - getty
    - hwrng
    - kexec

  - removed old BCPC-* roles
  - removed old templates no longer in use


chef/cookbooks/bcpc/recipes/host-aggregates.rb
```diff
@@ -1,63 +0,0 @@
-# Cookbook Name:: bcpc
```
furlongm: How are host aggregates handled now?

chef/environments/virtual.json
```diff
@@ -5,16 +5,22 @@
...
+        ]
+      },
       "proxy": {
-        "enabled": false,
+        "enabled": true,
         "proxies": {
-          "http": "",
-          "https": ""
+          "http": "http://devproxy.bloomberg.com:82",
+          "https": "http://devproxy.bloomberg.com:82"
```
furlongm: This shouldn't be included?

chef/roles/bootstrap.json
```diff
@@ -8,5 +8,9 @@
     "recipe[bcpc::ssl]",
     "recipe[bcpc::ssh]",
     "recipe[bcpc::file-server]"
+    /*
+    "recipe[bcpc::ufw]"
+    "recipe[bcpc::backup-cluster]"
+    */
```
furlongm: Why commented out?

chef/roles/headnode.json
```diff
@@ -26,5 +26,8 @@
     "recipe[bcpc::horizon]",
     "recipe[bcpc::os-quota]",
     "recipe[bcpc::flavors]"
+    /*
+    "recipe[bcpc::mysql-backup]"
+    */
```
furlongm: Why commented out?

chef/environments/virtual.json
justinjpacheco: mistakenly committed, i'll remove them

chef/roles/bootstrap.json
justinjpacheco: left as a reminder that these needed to still be implemented and/or looked at

chef/roles/headnode.json
justinjpacheco: same reason as above

justinjpacheco:
i agree that smaller commits are ideal but this branch is still under heavy development and likely to change rapidly

chef/cookbooks/bcpc/recipes/host-aggregates.rb
justinjpacheco: likely to be reviewed as a whole. they were largely used prior to the refactor to prevent vms from ending up on headnodes because they were running the nova-compute service, this is no longer possible.

furlongm: The maintenance aggregate was used for putting nodes into maintenance to avoid scheduling. My preference is to use `nova-manage service disable` for that purpose, so I don't mind seeing that go.

This commit also removes the functionality which auto-creates AZs. Is it the intention that operators will create those AZs manually in future?

@dkaplan34 , comments on the above from an operator perspective?

justinjpacheco: the AZ's will be created automatically. the code for that functionality will be pushed shortly. as it stands now, this code does not work as intended, so it is being removed until it is reworked.

furlongm: OK cool, it was not clear from the commit or commit message that it was the case that there was replacement code, so it just looked like functionality was being removed.

furlongm:
Regarding smaller commits, `git add -p` can be useful in that case.

PR #1369, opened 2018-07-01
refactored virtual environment

justinjpacheco:
- removed paramiko and yaml as a build time dependancy
  - updated Makefile to use tabs over spaces
  - replaced ansible virtual environment build code process with shell
    - create-virtual-environment
    - destroy-virtual-environment
  - reimplemented yaml virtual topology files in json
  - updated Vagrantfile to parse json topology files


Makefile
```diff
@@ -1,96 +1,85 @@
...
-    -t chef-node
+	ansible-playbook -v -i ${inventory} ansible/site.yml -t chef-node
 
 chef-client :
 
-  ansible-playbook -v \
-    -i ${inventory} ansible/site.yml \
-    -t chef-client --limit bootstraps
+	ansible-playbook -v \
+		-i ${inventory} ansible/site.yml \
```
comay: On this and other such continuation lines (following a line with an ending backslash), the convention I use is half the number of spaces (so in this case, four spaces instead of the tab stop).

virtual/bin/create-virtual-environment
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
comay: General question on new files like this, particular those that include "logic" rather than just configuration. Shouldn't they include a copyright and appropriate license block?

virtual/bin/generate-ansible-inventory
```diff
@@ -13,39 +13,36 @@ def is_valid_file(parser, arg):
...
 
-      # We want the ssh config to point to the real hostname, but we dont want to
-      # set ansible_ssh_host to the real name, but the ssh_config alias
-      if options['hostname'] == 'localhost':
-        options['hostname'] = hostname
+      line = line.strip()
 
-      ssh_config[hostname] = options
+      if line != '':
+        key,value = line.split(' ')
```
comay: Nit - space after the comma?

virtual/bin/generate-ansible-inventory
```diff
@@ -13,39 +13,36 @@ def is_valid_file(parser, arg):
...
+  return hosts
 
 def parse_topology_config(topology_config_file):
 
   with open(topology_config_file) as f:
-    return yaml.load(f)
+    return json.load(f)
 
-def get_group_hosts(group,hosts,nodes):
+def get_group_hosts(group,ssh_config,nodes):
```
comay: Space after the comma as well? In general, it makes lists like this easier to read.

Makefile
justinjpacheco: the use of tabs makes it difficult to know how many spaces will be equal to half the tab amount. some people might have it set to 8 some might have it at 4

virtual/bin/generate-ansible-inventory
```diff
@@ -59,19 +56,19 @@ def get_group_hosts(group,hosts,nodes):
...
       if len(primary_ip):
         host_vars['primary_ip'] = primary_ip[0]
 
-      group_hosts.update({ host: host_vars })
+      group_hosts.update({ host['Host']: host_vars })
 
   return group_hosts
 
-def generate_inventory(ssh_config,nodes):
+def get_inventory_data(ssh_config,nodes):
```
comay: Same here - might be good to check this throughout the source? "egrep ',[^ ]'" or something like that.

virtual/bin/generate-ansible-inventory
```diff
@@ -123,9 +140,10 @@ def main():
...
   topology = parse_topology_config(args.topology_conf)
-  inventory = generate_inventory(ssh_config,topology['nodes'])
-
-  yaml.dump(inventory, sys.stdout, default_flow_style=False)
+  inventory_data = get_inventory_data(ssh_config,topology['nodes'])
+  print generate_inventory(inventory_data)
 
 if __name__ == "__main__":
   main()
+
```
comay: Extraneous blank lines.

virtual/bin/generate-ansible-inventory
```diff
@@ -2,8 +2,8 @@
```
comay: General comment on Python code in bcpc - should we be standardizing on PEP-8 as OpenStack itself does?

virtual/bin/create-virtual-environment
furlongm: I think all files are covered by the LICENSE.txt in the top level of the project? Not sure about copyright notice per file.

comay: I'm not quite sure of the convention here. I know that Apache suggests having each source code file (I've always interpreted that to mean something that includes IP) should have a license block:

        Each original source document (code and documentation, but excluding the LICENSE
        and NOTICE files) SHOULD include a short license header at the top.

(from https://www.apache.org/dev/apply-license.html)

Makefile
comay: Understood. I don't know what the convention is for bcpc but I just know one common convention is for continuation lines to be indented at a smaller level than the primary one (specifying, for example, a new block of code).

justinjpacheco: i wouldn't know how to do that with tab characters and have it look readable across different configurations

virtual/bin/generate-ansible-inventory
furlongm: We could set up e.g. travis to enforce flake8?

PR #1373, opened 2018-07-03
add pod support

justinjpacheco:
- update the vagrant host definitions to include pod information
  - parse pod id out of hostname and use it to locate the correct pod
    networking information in the networking topology

### DIFF
--- a/chef/cookbooks/bcpc/attributes/networking.rb
+++ b/chef/cookbooks/bcpc/attributes/networking.rb
@@ -6,36 +6,36 @@ require 'ipaddr'
 require 'ipaddress'
 
 default['bcpc']['networking']['topology'] = {
-  'networks': {
-    'primary': {'cidr': '10.121.84.0/22', 'dev': 'eth1'},
-    'storage': {'cidr': '10.121.88.0/22', 'dev': 'eth2'}
+  'networks' => {
+    'primary' => { 'cidr' => '10.121.84.0/22', 'dev' => 'eth1' },
+    'storage' => { 'cidr' => '10.121.88.0/22', 'dev' => 'eth2' }
   },
-  'racks': [
+  'racks' => [
     {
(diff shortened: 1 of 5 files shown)


chef/cookbooks/bcpc/attributes/networking.rb
```diff
@@ -6,50 +6,73 @@
...
     }
-  }
+  ]
 }
 
 # try to get node information via the hostname
-# eg. ob-cloud-r(01)(a)n(01) = rack_id,pod_id,node_id
+# eg. cloud-r(01)(a)n(01) = rack_id,pod_id,node_id
 #
+rack_id,pod_id,node_id = [nil,nil,nil]
```
comay: Please add spaces after the commas on both sides of the assignment. It looks like there are other instances of this in the file too which should be corrected.

chef/cookbooks/bcpc/attributes/networking.rb
```diff
@@ -6,50 +6,73 @@
...
+racks = default['bcpc']['networking']['topology']['racks']
 
 if match = node['hostname'].match(/.*r(\d+)(\w+)?n(\d+)$/i)
+
   rack_id,pod_id,node_id = match.captures
```
comay: For example, here. 😉

virtual/topology/1h1w.json
```diff
@@ -33,7 +33,7 @@
       }
     },
     {
-      "host": "r1n1",
+      "host": "r1an1",
```
comay: Oh noes, need to fight against muscle memory. 🙃

justinjpacheco: lol, i had the same thought

justinjpacheco:
adding another example topology to show an additional pod wouldn't be a problem. just need to make sure the TOR can support it

PR #1375, opened 2018-07-05
networking fixes and configuration file cleanup

justinjpacheco:
- apt
    - added apt attributes file for apt specific configuration
    - updated apt.conf.d/00-bcpc to accept all apt config as template
      variable
    - added support for customized sources.list contents
  - calico
    - removed old calico config
    - moved calico config from default into attributes/calico.rb
    - added arch amd64 to apt_repository resource
  - networking
    - fixed incorrect array indexing from regex capture
    - moved systemd-resolved dns configuration to networking.rb recipe
  - mysql
    - added arch amd64 to apt_repository resource
  - systemd-resolved
    - added configuration option for fallback dns servers
  - virtual environment
    - removed proxy,dns_servers and ntp configuration. defaults are now in
      defaults.rb

### DIFF
--- a/chef/cookbooks/bcpc/attributes/default.rb
+++ b/chef/cookbooks/bcpc/attributes/default.rb
@@ -9,7 +9,12 @@ default['bcpc']['cloud']['fqdn'] = "openstack.#{default['bcpc']['cloud']['domain
 default['bcpc']['cloud']['region'] = node.chef_environment
 default['bcpc']['cloud']['vip'] = {'ip': '10.10.254.254/32'}
 
+# list of dns servers to use
 default['bcpc']['dns_servers'] = ["8.8.8.8", "8.8.4.4"]
+
+# list of ntp servers to use
+default['bcpc']['ntp']['servers'] = ["time.google.com"]
+
 default['bcpc']['file_server']['url'] = 'http://bootstrap:8080'
 
 # Hypervisor domain (domain used by actual machines)
@@ -18,6 +23,16 @@ default['bcpc']['hypervisor_domain'] = "hypervisor-bcpc.example.com"
 # convenience variable
 #
 vip = IPAddress(node['bcpc']['cloud']['vip']['ip']).address
+
+###############################################################################
+# ubuntu
+###############################################################################
+
+default['bcpc']['ubuntu']['archive_url'] = 'http://archive.ubuntu.com/ubuntu'
+default['bcpc']['ubuntu']['security_url'] = 'http://security.ubuntu.com/ubuntu'
+default['bcpc']['ubuntu']['codename'] = node['lsb']['codename']
+default['bcpc']['ubuntu']['components'] = ['main','restricted','universe','multiverse']
+default['bcpc']['ubuntu']['arch'] = 'amd64'
 
 ###############################################################################
 # proxy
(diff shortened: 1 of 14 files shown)


chef/cookbooks/bcpc/attributes/default.rb
```diff
@@ -9,7 +9,12 @@
 default['bcpc']['cloud']['region'] = node.chef_environment
 default['bcpc']['cloud']['vip'] = {'ip': '10.10.254.254/32'}
 
+# list of dns servers to use
 default['bcpc']['dns_servers'] = ["8.8.8.8", "8.8.4.4"]
+
+# list of ntp servers to use
+default['bcpc']['ntp']['servers'] = ["time.google.com"]
```
bryanapperson: Only one timeserver?

justinjpacheco: one by default, it can take an array if the user wants to add more

mihalis68: a real datacenter cluster would do better

justinjpacheco: @mihalis68 we can add time1.google.com, time2.google.com as defaults if you like

mihalis68: sure, hinting to people to use more than 1 is good practice

chef/cookbooks/bcpc/recipes/calico-apt.rb
```diff
@@ -37,6 +37,7 @@
 return unless node['bcpc']['calico']['repo']['enabled']
 
 apt_repository "calico" do
+  arch 'amd64'
   uri node['bcpc']['calico']['repo']['url']
   distribution 'xenial'
```
comay: Is there a 'bionic' version available for Calico?

chef/cookbooks/bcpc/attributes/default.rb
```diff
@@ -19,6 +24,16 @@
...
 vip = IPAddress(node['bcpc']['cloud']['vip']['ip']).address
 
+###############################################################################
+# ubuntu
+###############################################################################
+
+default['bcpc']['ubuntu']['archive_url'] = 'http://archive.ubuntu.com/ubuntu'
+default['bcpc']['ubuntu']['security_url'] = 'http://security.ubuntu.com/ubuntu'
+default['bcpc']['ubuntu']['codename'] = node['lsb']['codename']
+default['bcpc']['ubuntu']['components'] = ['main','restricted','universe','multiverse']
```
comay: Spaces after each comma?

chef/cookbooks/bcpc/recipes/calico-apt.rb
justinjpacheco: there does not appear to be a bionic version for calico yet

chef/cookbooks/bcpc/attributes/default.rb
justinjpacheco: i will address these and others when i implement pre-commit hooks for linting ruby files

justinjpacheco:
Documentation seems like the best choice at the moment. Something like a "howto" section showing examples of common things to override.

PR #1377, opened 2018-07-09
ansible refactoring and chef role generation

justinjpacheco:
- ansible
    - moved playbooks into playbooks/
    - flattened group_vars/all for easier overrides in sub-children
      group_vars
  - chef
    - roles are now generated via python


comay:
Perhaps I'm missing something obvious but what is the intent for generating the roles via Python?

justinjpacheco:
@comay it provides an easy way for other projects that wish to inherit these roles to override various aspects of the role data structure. eg: adding items to the base run_list

Issue #791, opened 2015-08-25
getty processes crash repeatedly in virtualized local builds


erhudy:
With the merge of #735, when building locally, the `ttyS0` and `ttyS1` processes repeatedly crash and are respawned by Upstart:
```
Aug 25 13:43:05 vagrant-ubuntu-trusty-64 kernel: [  662.010796] init: ttyS0 main process (14186) terminated with status 1
Aug 25 13:43:05 vagrant-ubuntu-trusty-64 kernel: [  662.010808] init: ttyS0 main process ended, respawning
Aug 25 13:43:05 vagrant-ubuntu-trusty-64 kernel: [  662.054730] init: ttyS1 main process (14194) terminated with status 1
Aug 25 13:43:05 vagrant-ubuntu-trusty-64 kernel: [  662.054749] init: ttyS1 main process ended, respawning
```
This doesn't appear to occur on hardware.

kamidzi:
Confirm. Only started seeing this yesterday.

erhudy:
After stracing the processes on the VM versus a hardware system, it looks like this is getty's reaction to the serial port not actually existing (you can reproduce on hardware as well by trying to set up a getty on ttyS2 and beyond). I made it happy again by adding fake serial ports to the VMs that don't go anywhere:
```
➜  vagrant_scripts git:(master) VBoxManage modifyvm bcpc-vm3 --uart1 0x3F8 4    
➜  vagrant_scripts git:(master) VBoxManage modifyvm bcpc-vm3 --uart2 0x2F8 3
➜  vagrant_scripts git:(master) VBoxManage modifyvm bcpc-vm3 --uartmode1 disconnected
➜  vagrant_scripts git:(master) VBoxManage modifyvm bcpc-vm3 --uartmode2 disconnected
```
I will work this into a PR to modify the Vagrantfile so that these serial ports are added to each VM.

PR #1380, opened 2018-07-10
Align VirtualBox (5.1 or 5.2) VM configuration with V7 settings.

comay:
With this change, the VMs created by VirtualBox via Vagrant will have the same optimizations that the ones under V7 did. Most of the latter settings are already the default in VirtualBox 5.1 and 5.2 so this only sets the ones that need to be changed.

### DIFF
--- a/virtual/Vagrantfile
+++ b/virtual/Vagrantfile
@@ -49,6 +49,7 @@ Vagrant.configure("2") do |config|
 
         args = {}.tap do |args|
           args[:virtualbox__intnet] = n['network']
+          args[:nic_type] = '82543GC'
           args[:auto_config] = false
         end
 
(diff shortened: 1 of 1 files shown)


virtual/Vagrantfile
```diff
@@ -67,6 +68,11 @@ Vagrant.configure("2") do |config|
         vb.cpus = hw_profile['cpus']
         vb.memory = hw_profile['ram_gb'] * 1024
         vb.customize ['modifyvm', :id, '--groups', project_name]
+        vb.customize ['modifyvm', :id, '--uart1', '0x3F8', '4']
+        vb.customize ['modifyvm', :id, '--uart2', '0x2F8', '3']
+        vb.customize ['modifyvm', :id, '--uartmode1', 'disconnected']
+        vb.customize ['modifyvm', :id, '--uartmode2', 'disconnected']
```
furlongm: Do we actually need a UART?

virtual/Vagrantfile
```diff
@@ -49,6 +49,7 @@ Vagrant.configure("2") do |config|
 
         args = {}.tap do |args|
           args[:virtualbox__intnet] = n['network']
+          args[:nic_type] = '82543GC'
```
furlongm: Does this NIC provide better performance than the others?

virtual/Vagrantfile
comay: I believe the changes were put in to address

https://github.com/bloomberg/chef-bcpc/issues/791

but I don't know if the underlying issue still would be a problem.

virtual/Vagrantfile
comay: Supposedly although I wasn't able to see a performance testing when this change was deployed on the BIRD-based leafy-spines implementation. It looks like this was integrated into BCPC quite awhile ago.

PR #1384, opened 2018-07-12
az and ha support

justinjpacheco:
- availability zones
    - created for each rack listed in the network topology

  - host aggregates
    - each compute node is added to the az for its rack

  - ansible
    - upload-bcpc tag
      - corrected rsync options to properly ignore .swp and .git
      - use delete attribute on syncronize module


justinjpacheco:
@comay yep, that works

PR #1387, opened 2018-07-13
Anycast address should only be exported by Consul leader.

comay:
The checks in the DNS and haproxy watch scripts don't actually check whether a node is the Consul leader before enabling or disabling BGP. After checking this, only one of the head nodes will end up exporting its routes (in particular any Anycast addresses) to the TOR in the rack.

This was tested by manually resetting the Consul services on the leader causing a new election and then observing a new head node becomes the node exporting the Anycast routes.

### DIFF
--- a/chef/cookbooks/bcpc/templates/default/consul/dns-watch.erb
+++ b/chef/cookbooks/bcpc/templates/default/consul/dns-watch.erb
@@ -1,11 +1,45 @@
 #!/bin/bash
 
+#
+# Copyright 2018, Bloomberg Finance L.P.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+disable_bgp() {
+    /usr/sbin/birdc 'disable bgp1'
+}
+
+enable_bgp() {
+    /usr/sbin/birdc 'enable bgp1'
+}
+
+self_ip="<%= node['ipaddress'] %>"
+leader=$(curl -qs http://localhost:8500/v1/status/leader | jq -r '.')
+if [ "${self_ip}" != "${leader%:8300}" ]; then
+    disable_bgp
+    exit 0
+fi
+
 status="`cat`"
 self="<%= node['hostname'] %>"
 echo "$status" > /tmp/dns-handler.log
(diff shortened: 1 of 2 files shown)


chef/cookbooks/bcpc/templates/default/consul/dns-watch.erb
```diff
@@ -1,11 +1,45 @@
...
+disable_bgp() {
+    sudo /usr/sbin/birdc 'disable bgp1'
+}
+
+enable_bgp() {
+    sudo /usr/sbin/birdc 'enable bgp1'
+}
+
+self_ip="<%= node['ipaddress'] %>"
+leader=$(eval echo $(curl -qs http://localhost:8500/v1/status/leader))
```
justinjpacheco: do we need the eval and echo?

chef/cookbooks/bcpc/templates/default/consul/dns-watch.erb
```diff
@@ -1,11 +1,45 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+disable_bgp() {
+    sudo /usr/sbin/birdc 'disable bgp1'
```
justinjpacheco: i believe consul is running as a root owned process, so sudo won't be required

chef/cookbooks/bcpc/templates/default/consul/dns-watch.erb
comay: We do because the leader IP address comes out quoted, so the eval echo is to get the string inside.

justinjpacheco:
you could pipe to jq -r

chef/cookbooks/bcpc/templates/default/consul/dns-watch.erb
comay: Indeed - just tested that. I'll make that change.

comay:
Cool trick using 'jq' - I've made the update.

chef/cookbooks/bcpc/templates/default/consul/dns-watch.erb
justinjpacheco: how about:
leader=$( curl -qs http://localhost:8500/v1/status/leader | tr -d '"')

justinjpacheco:
on a 3h3w build, the consul leader is the second headnode (r2an1) but primary.mysql.service.consul is pointing to the first headnode (r1an1). im assuming during the second pass of chef-client, consul elected a new leader when the current leader (r1an1) rebooted because of an update to its configuration. the election of a new leader didn't change the DNS entry for primary.mysql.service.consul to point to the new leader.

comay:
This particular pull request only resolves the BGP issue and not the MySQL DNS issue. I'd prefer address these separately.

justinjpacheco:
understood. this won't break current functionality because primary.mysql.service.consul points to the primary address of the node and not the local anycast address.

PR #1388, opened 2018-07-15
linting fixes

justinjpacheco:


chef/cookbooks/bcpc/attributes/catalog.rb
```diff
@@ -1,16 +1,10 @@
...
-###########################################
-
 default['bcpc']['catalog'] = {
   'identity' => {
     'name' => 'keystone',
     'project' => 'keystone',
     'description' => 'OpenStack Identity',
     'ports' => {
-      'admin' => 35357,
+      'admin' => 35_357,
```
furlongm: Should this not be a port number?

chef/cookbooks/bcpc/attributes/default.rb
```diff
@@ -196,54 +162,50 @@
 # (generally stuff served via HAProxy)
 # some ports are hardcoded - see bcpc-firewall.erb template
 default['bcpc']['management']['firewall_tcp_ports'] = [
-  8088,7480,35357,8004,8000
+  8088, 7480, 35_357, 8004, 8000
```
furlongm: Ditto here

chef/cookbooks/bcpc/attributes/default.rb
```diff
@@ -196,54 +162,50 @@
...
 ]
 
 # use this to *add* more reserved ports; i.e. modify value of
 # net.ipv4.ip_local_reserved_ports
 default['bcpc']['system']['additional_reserved_ports'] = []
 
 # any other sysctl parameters (register under parameters)
-default['bcpc']['system']['parameters']['kernel.pid_max'] = 4194303
+default['bcpc']['system']['parameters']['kernel.pid_max'] = 4_194_303
```
furlongm: Should be an integer?

chef/cookbooks/bcpc/attributes/default.rb
```diff
@@ -196,54 +162,50 @@
...
 
 # connection tracking table max size
-default['bcpc']['system']['parameters']['net.nf_conntrack_max'] = 262144
+default['bcpc']['system']['parameters']['net.nf_conntrack_max'] = 262_144
```
furlongm: And here?

chef/cookbooks/bcpc/attributes/catalog.rb
justinjpacheco: it is. ruby prefers numbers larger than 4 characters use underscores to separate groups of three (for readability). the underscores are ignored when used

chef/cookbooks/bcpc/attributes/default.rb
```diff
@@ -252,8 +214,7 @@
 # Enable memcached double verbose logging.
 default['bcpc']['memcached']['debug'] = false
 # Set number of memcached connections.
-default['bcpc']['memcached']['connections'] = 10240
-
+default['bcpc']['memcached']['connections'] = 10_240
```
furlongm: And here?

chef/cookbooks/bcpc/attributes/default.rb
justinjpacheco: read above

chef/cookbooks/bcpc/attributes/default.rb
justinjpacheco: read above

chef/cookbooks/bcpc/attributes/default.rb
justinjpacheco: read above

chef/cookbooks/bcpc/attributes/default.rb
justinjpacheco: read above

chef/cookbooks/bcpc/attributes/catalog.rb
furlongm: Rubocop prefers, not ruby prefers? Ruby supports it, as other languages do, but it is by no means a requirement.

I can see this being problematic with no great advantages (simple example: grepping 35357, a well known port in OpenStack and finding nothing).

justinjpacheco: that is correct, rubocop prefers it, i don't believe it was stated as a requirement.

furlongm: Given that it's not a requirement, and it can be problematic, can we revert to using non-underscored integers?

PR #1389, opened 2018-07-17
added vlan and mtu support

justinjpacheco:
- interfaces
    - network interfaces now support mtu and vlan settings from config
  - networking
    - refactored to consolidate networking logic into util functions
  - chef recipes
    - updated to account for networking data structure shuffling
  - Vangrantfile
    - updated topology files to create bootstrap node as r1an0 but refer to it
      as 'bootstrap' to the user.
    - updated inventory generator to use hostname from topology file vs. ssh
      config file
  - rubocop
    - added config file for easier configuration overrides

### DIFF
--- a/chef/cookbooks/bcpc/libraries/utils.rb
+++ b/chef/cookbooks/bcpc/libraries/utils.rb
@@ -17,6 +17,7 @@
 
 return Chef::Log.warn('requires chef-server') if Chef::Config[:solo]
 
+require 'ipaddr'
 require 'ipaddress'
 
 def init_cloud?
@@ -31,6 +32,11 @@ end
 
 def worknode?(node)
   search(:node, "role:worknode AND hostname:#{node['hostname']}").any?
+end
+
+def bootstraps
+  nodes = search(:node, 'role:bootstrap')
+  nodes.sort! { |a, b| a['hostname'] <=> b['hostname'] }
 end
 
 def get_headnodes(exclude: nil, all: false)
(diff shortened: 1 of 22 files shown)


chef/cookbooks/bcpc/libraries/utils.rb
```diff
@@ -103,12 +111,93 @@ def get_address(cidr)
...
+
+  {
+    'rack_id' => match.captures[0].to_i,
+    'pod_id' => match.captures[1],
+    'node_id' => match.captures[2]
+  }
+end
+
+def node_networking
+  {
```
comay: Is this actually used?

chef/cookbooks/bcpc/recipes/networking.rb
```diff
@@ -39,10 +39,8 @@
 # primary interface configuration
 #
 begin
-  ifaces = node['bcpc']['networking']['ifaces']
-  raise 'could not find a primary interface' unless ifaces.key?('primary')
-
-  primary = ifaces['primary']
+  primary = node_interfaces.find { |i| i['type'] == 'primary' }
+  raise 'unable to find primary interface' if primary.nil?
```
comay: Nit - I suggest adding "the " in front of "primary" since it's a bit clearer and matches the same message below for the "storage" network.

virtual/Vagrantfile
```diff
@@ -32,7 +32,9 @@ Vagrant.configure('2') do |config|
   config.vm.box_download_insecure = true
 
   topology['nodes'].each do |node|
-    config.vm.define node['host'] do |subconfig|
+    vm_name = node['name'] || node['host']
+
+    config.vm.define vm_name do |subconfig|
       networks = node['networking']['networks']
       gateway = node['networking']['gateway']
       primary = networks.find { |n| n['type'] == 'primary' }
```
comay: Any existing issue but can't these assignments be moved outside the subconfig loop

justinjpacheco: yep, definitely

chef/cookbooks/bcpc/libraries/utils.rb
justinjpacheco: wouldn't appear to be. old idea that didn't pan out and i forgot to remove the function

PR #1392, opened 2018-07-20
networking refactoring

justinjpacheco:
- Makefile
    - chef-client-worknodes
      - updated to run_once=true for first pass of worknodes, then run
        everywhere
  - chef
    - add recipes updated to account for networking util refactoring
    - updated databag ssl to use a 1024 bit ssl cert

  - sshd
    - updated sshd_config to listen on virtualbox nat ip if virtualized
    - updated to only listen on primary/storage network vs. all interfaces

  - network interfaces
    - updated storage interface to fix static route to storage network


chef/cookbooks/bcpc/files/default/vim/vimrc
```diff
@@ -0,0 +1,11 @@
...
+set tabstop=4
+set softtabstop=4
+set shiftwidth=4
+set expandtab
+set autoindent
+set ruler
+set ls=2
+set background=dark
+set nocompatible
+set term=screen-256color
```
comay: Nit - sort the file so the options appear in alphabetical order.

chef/cookbooks/bcpc/libraries/utils.rb
```diff
@@ -175,23 +200,25 @@ def node_primary_interface
 end
 
 def node_storage_interface
```
comay: node_primary_interface and node_storage_interface have a very similar structure. Could you change one to look more like the other so that a "diff" between them only shows really their differences. For example, the former sets cloud_network right after pod_network but the latter does so later. Also, the former uses a "gateway" variable which is then included in "iface" but the latter just references the corresponding base value directly in "iface".

chef/cookbooks/bcpc/recipes/ceph-osd.rb
```diff
@@ -91,5 +89,5 @@
     end
   end
 
-  only_if { headnode?(node) }
+  only_if { headnode? }
```
comay: Didn't you say this PR was going to remove putting OSDs on headnodes or did I misunderstand?

chef/cookbooks/bcpc/recipes/ssh.rb
```diff
@@ -39,3 +41,12 @@
     nodes: all_nodes
   )
 end
+
+template '/etc/ssh/sshd_config' do
+  source 'ssh/sshd_config.erb'
+  variables(
+    primary: node_interfaces(type: 'primary'),
+    storage: node_interfaces(type: 'storage')
```
furlongm: Why primary and storage only? It can be sometimes be useful to have ssh listening everywhere (e.g. dead NIC)

chef/cookbooks/bcpc/templates/default/ssh/sshd_config.erb
```diff
@@ -1,18 +1,16 @@
...
+###############################################################################
 
 # What ports, IPs and protocols we listen for
 Port 22
 # Use these options to restrict which interfaces/protocols sshd will bind to
 #ListenAddress ::
-<% if node['bcpc']['vagrant'] and node['bcpc']['vagrant']['nat_ip'] %>
-ListenAddress <%= node['bcpc']['vagrant']['nat_ip'] %>
+<% if node['virtualization']['system'] == 'vbox' %>
+ListenAddress 10.0.2.15
```
furlongm: Should this be hardcoded or should it come from a variable?

chef/cookbooks/bcpc/recipes/ssh.rb
justinjpacheco: existing functionality ported over. not sure if we want to listen on all interfaces

chef/cookbooks/bcpc/templates/default/ssh/sshd_config.erb
justinjpacheco: which part? the ip or where im looking?

chef/cookbooks/bcpc/recipes/ssh.rb
furlongm: Maybe all or one? Primary and Storage seems like a non-intuitive selection of interfaces.

chef/cookbooks/bcpc/recipes/ceph-osd.rb
justinjpacheco: not in this PR

chef/cookbooks/bcpc/templates/default/ssh/sshd_config.erb
furlongm: The IP address

chef/cookbooks/bcpc/recipes/ssh.rb
justinjpacheco: im good with the primary interface

chef/cookbooks/bcpc/templates/default/ssh/sshd_config.erb
justinjpacheco: ok

PR #1394, opened 2018-07-23
don't auto-upgrade packages

furlongm:


cookbooks/bcpc-extra/recipes/kexec.rb
```diff
@@ -18,7 +18,7 @@
 #
 
 package 'kexec-tools' do
-  action :upgrade
+  action :install
```
justinjpacheco: the default action is to install if you want to save a few lines

furlongm: Just trying to minimize changes. Not sure which diff would be worse for a rebase though?

justinjpacheco: it's easy enough to identify either way. i'll fix the rebase after this is merged

PR #1395, opened 2018-07-24
disable unused services

furlongm:


chef/cookbooks/bcpc/recipes/system.rb
```diff
@@ -137,3 +137,10 @@
 cookbook_file '/etc/profile.d/bcpc.sh' do
   source 'profile.d/bcpc.sh'
 end
+
+disable_services = ['rpcbind', 'lxcfs', 'snapd', 'lxd', 'iscsid']
```
comay: Nit - could you sort the list of services?

justinjpacheco: would the removal of iscsid impact cinder's ability to utilize SANs? how much memory are these services actually consuming?

furlongm: I don't think so. iscsid is the server component, iscsi is the client.

It's not solely about memory utilization, also want to disable services that are not required. I will update PR title.

PR #1398, opened 2018-07-24
remaining cookstyle fixes

furlongm:


justinjpacheco:
has this been run though a build?

furlongm:
Building now, but if it we know it will fail, happy to cancel.

PR #1402, opened 2018-08-01
ceph, nova, glance and make improvements

justinjpacheco:
- make
    - new target: adjust-ceph-pool-pgs
        - calculates pool pgs based on size and osd count
    - new target: add-cloud-images
        - adds cloud images to glance
    - updated target: register-compute-nodes
        - updated ansible target to enable nova compute service
- ceph
    - removed osd from headnodes
    - added rack crush map buckets
        - osd hosts are now added under their racks

- nova
    - updated nova-compute config to disable compute services on startup
        - compute services are enable by operator: register-compute-nodes


ansible/playbooks/roles/headnode/tasks/adjust-ceph-pool-pgs.yml
```diff
@@ -0,0 +1,55 @@
...
+      pool="$1"
+      echo $(ceph osd pool get ${pool} pg_num -f json | jq '.pg_num')
+    }
+
+    pool_pg_num_calc() {
+      pool="$1"
+      size=$(pool_size ${pool})
+      pg_num=$(pool_pg_num ${pool})
+      num_osd=$(ceph osd ls 2> /dev/null | wc -l)
+      target_pg_num=$(power2 $(echo "(100 * $num_osd * 80*0.01) / $size" | bc))
```
comay: Nit - for readability, add white space so the middle expression is of the form (100 * $num_osd * (80 * 0.01))

(Is the last part of the expression supposed to represent 80%? Maybe some comments are in order here for the math?)

ansible/playbooks/roles/headnode/tasks/adjust-ceph-pool-pgs.yml
```diff
@@ -0,0 +1,55 @@
...
+
+    for pool in ${pools}; do
+      pg_num=$(pool_pg_num $pool)
+      target_pg_num=$(pool_pg_num_calc $pool)
+
+      if [ $pg_num -eq $target_pg_num ]; then
+        echo "pool: ${pool} (${target_pg_num} target pgs) (${pg_num} pgs)"
+        continue
+      else
+        echo "pool: ${pool} (${target_pg_num} target pgs) (${pg_num} pgs)"
```
comay: Can lines 37 and 40 be pulled out of the conditional since they're identical?

chef/environments/virtual.json
```diff
@@ -5,6 +5,7 @@
   "chef_type": "environment",
   "override_attributes": {
     "bcpc": {
+      "dns_servers": ["10.43.0.1"],
```
comay: Where does this address come from? Is it just an example?

chef/cookbooks/bcpc/attributes/nova.rb
```diff
@@ -73,12 +72,6 @@
   ServerGroupAffinityFilter
 )
 
-# configure optional Nova notification system
-default['bcpc']['nova']['notifications']['enabled'] = false
-default['bcpc']['nova']['notifications']['notification_topics'] = 'notifications'
-default['bcpc']['nova']['notifications']['notification_driver'] = 'messagingv2'
-default['bcpc']['nova']['notifications']['notify_on_state_change'] = 'vm_state'
-
```
comay: Is there a reason we're hard-wiring these in nova.conf now rather than leaving them as something that can be overridden?

chef/cookbooks/bcpc/templates/default/nova/nova.conf.erb
```diff
@@ -10,10 +10,7 @@ ssl_only = false
 key = /etc/nova/ssl-bcpc.key
 cert = /etc/nova/ssl-bcpc.pem
 my_ip = <%= node['ipaddress'] %>
-
-<% if node['bcpc']['nova']['cpu_config']['vcpu_pin_set'] %>
-vcpu_pin_set = <%= node['bcpc']['nova']['cpu_config']['vcpu_pin_set'] %>
```
furlongm: I think this is worth keeping.

Makefile
```diff
@@ -32,15 +32,21 @@ operator :
...
+	ansible-playbook -v \
+		-i ${inventory} ${playbooks}/site.yml \
+		-t chef-server --limit bootstraps
 
 chef-workstation :
 
-	ansible-playbook -v -i ${inventory} ${playbooks}/site.yml -t chef-workstation
+	ansible-playbook -v \
+		-i ${inventory} ${playbooks}/site.yml \
+		-t chef-workstation --limit bootstraps
```
bryanapperson: you can probably just use the group name without --limit, but same idea.

ansible/playbooks/roles/headnode/tasks/adjust-ceph-pool-pgs.yml
```diff
@@ -0,0 +1,55 @@
+- name: adjust ceph pool pgs
```
bryanapperson: nice, I assume this will be run when needed.

chef/cookbooks/bcpc/attributes/ceph.rb
```diff
@@ -5,12 +5,10 @@
 default['bcpc']['ceph']['repo']['enabled'] = false
 default['bcpc']['ceph']['repo']['url'] = ''
 
+default['bcpc']['ceph']['pg_num'] = 64
```
bryanapperson: better

chef/cookbooks/bcpc/recipes/ceph-osd.rb
```diff
@@ -58,36 +58,24 @@
...
+      EOH
+      only_if "lsblk /dev/#{osd}"
+      not_if "blkid /dev/#{osd}1 | grep 'ceph data'"
     end
   end
 
-  only_if { headnode? }
+  bash "move #{host} host to ceph rack bucket" do
+    code <<-EOH
+      ceph osd crush move #{host} rack=#{rack}
```
bryanapperson: Nice.

chef/cookbooks/bcpc/templates/default/nova/nova.conf.erb
furlongm: +1

furlongm: Can we leave the settings here in that case, rather than removing them?

PR #1405, opened 2018-08-12
add openstack designate, powerdns and misc fixes

justinjpacheco:
- openstack designate
    - provides dns as a service
    - using pdns4 (PowerDNS 4) as the DNS service
  - powerdns 4
    - added powerdns automation
  - chef
    - added powerdns and designate blocks to the databag
  - haproxy
    - updated to blance designate-api requests

  - misc fixes
    - default
      - change cloud vip from a string to a hash with ip and prefix vs.
        cidr string
    - consul
      - pulled out config from attributes/default.rb into attributes/consul.rb
    - unbound
      - pulled out config from attributes/default.rb into attributes/unbound.rb
    - keystone
      - enabled caching


chef/cookbooks/bcpc/templates/default/powerdns/create-db.sql.erb
```diff
@@ -0,0 +1,104 @@
+CREATE DATABASE IF NOT EXISTS <%= @db['dbname'] %>;
```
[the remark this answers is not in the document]
justinjpacheco: i figured the relative location of the file would be good enough to give context rather than putting $service in the file name when it's already in a directory named after that service

chef/cookbooks/bcpc/attributes/default.rb
```diff
@@ -1,13 +1,11 @@
...
-
 ###############################################################################
 # cloud
 ###############################################################################
 
 default['bcpc']['cloud']['domain'] = 'bcpc.example.com'
 default['bcpc']['cloud']['fqdn'] = "openstack.#{node['bcpc']['cloud']['domain']}"
 default['bcpc']['cloud']['region'] = node.chef_environment
-default['bcpc']['cloud']['vip'] = { 'ip' => '10.10.254.254/32' }
+default['bcpc']['cloud']['vip'] = { 'ip' => '10.10.254.254', 'prefix' => '32' }
```
furlongm: Is there ever a case where the VIP prefix is not 32?

justinjpacheco: none that i can think of

furlongm: So we could probably just hardcode it?

justinjpacheco: seems ok to me unless anyone objects

chef/cookbooks/bcpc/templates/default/powerdns/create-db.sql.erb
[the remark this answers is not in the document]
justinjpacheco: i agree

PR #1407, opened 2018-08-19
Add cert to rally config

srsakhamuri:


cookbooks/bcpc-extra/templates/default/rally.existing.json.erb
```diff
@@ -20,4 +20,5 @@
         "project_domain_name": "<%=@domain_name%>"
     },
     "https_insecure": false,
+    "https_cacert": "/usr/local/share/ca-certificates/ssl-bcpc.crt",
```
justinjpacheco: we should probably make this file path a variable in attributes/default.rb as it is referenced directly in multiple locations. but we should probably do that in a separate commit

srsakhamuri: Agree

PR #1410, opened 2018-08-20
remove v2.0 keystone rally config

srsakhamuri:
No need to setup v2.0 config from mitaka

### DIFF
--- a/cookbooks/bcpc-extra/recipes/rally-deployments.rb
+++ b/cookbooks/bcpc-extra/recipes/rally-deployments.rb
@@ -19,7 +19,7 @@
 # Note: The rally.rb recipe must have already been executed before running this one.
 # IMPORTANT: The head nodes MUST have already been installed and the keystone endpoints working. Rally verifies.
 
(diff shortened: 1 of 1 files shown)


cookbooks/bcpc-extra/recipes/rally-deployments.rb
```diff
@@ -19,7 +19,7 @@
 # Note: The rally.rb recipe must have already been executed before running this one.
 # IMPORTANT: The head nodes MUST have already been installed and the keystone endpoints working. Rally verifies.
 
-KEYSTONE_API_VERSIONS = %w{ v2.0 v3 }
+KEYSTONE_API_VERSIONS = %w{ v3 }
```
furlongm: If there's only one version, can we remove the parameterization and hardcode v3? That would simplify the rest of the code in this recipe.

srsakhamuri: Make sense will clean that up

PR #1412, opened 2018-08-27
use bionic repos for haproxy, rabbitmq and erlang

furlongm:


justinjpacheco:
the system is already configured to use the bionic repos for haproxy, rabbitmq and erland. they all have enabed set to false, this is the flagged that's used to determine if it should use an external repository or not. this was done intentionally to provide hints on how to use the external repositories if needed.

furlongm:
If the repos are not in use, is there any point in keeping the related code? The code will still exist in the git history if these repos are ever required in the future, whereas removing it via this commit simplifies the code by removing unused cruft.

justinjpacheco:
the repos aren't in use now, that doesn't mean they won't be in the future. and the sections you've removed are just the configuration to enable/disable their use. the actual code is in their respective recipes. also, i don't view toggles as "cruft". we may want to use the latest versions in the future if the bionic packages fall to far behind.

furlongm:
This commit removes more than just toggles? It removes unused attributes (including toggles), unused repo keys and unused code in the recipes. As noted, all can be simply recovered from git history in the future if required.

furlongm:
This is a dev branch so it is unlikely that there are any other users . As it's a development branch we can make breaking changes, and if required we can always revert the commits if requested.

Are these currently unused codepaths tested?

justinjpacheco:
> Are these currently unused codepaths tested
tested during development. they haven't been used in some time.

> This is a dev branch so it is unlikely that there are any other users . As it's a development branch we can make breaking changes, and if required we can always revert the commits if requested.

I don't follow the reasoning here. Yes, this is a development branch, but it will soon become a production branch, so why does that impact code features that are designed to import software from additional repositories once newer versions are available? HAproxy and RabbitMQ will eventually have newer releases that won't be included in the vanilla distribution. When that happens, we will have the option of using those newer versions by changing our config. This is preferable to digging though the git history to find deleted code blocks.

I see no technical or practical reason to remove these code blocks. They were implemented precisely for the reasons listed above and they have no impact on the system or the automation if the operator decides to not use them.

bcpc:
@furlongm #1412 rebased on feature/v8-bionic with multihead with 1412 <https://github.com/bloomberg/chef-bcpc/pull/1412|use bionic repos for haproxy, rabbitmq and erlang> by @furlongm failed during converge <https://cloud-infra.jaas.dev.bloomberg.com/job/bcpc/job/jenkins/job/master/754/|Jenkins log>

mihalis68:
I don't want to terminate this discussion, so consider this a non-binding resolution, however I see code which allows for the dependencies to be acquired from a range of miscellaneous upstreams as important and customary for our project. It's nice that right now our distro has all the good stuff but some of it will likely move faster elsewhere in future. For example haproxy now has an ambitious commercial maintainer, it might well be that we want stuff from their "community edition" before Canonical merges it into a main ubuntu release, we might be grabbing from a ppa in that case

bcpc:
@furlongm #1412 rebased on feature/v8-bionic with singlehead with 1412 <https://github.com/bloomberg/chef-bcpc/pull/1412|use bionic repos for haproxy, rabbitmq and erlang> by @furlongm failed during converge <https://cloud-infra.jaas.dev.bloomberg.com/job/bcpc/job/jenkins/job/master/755/|Jenkins log>

furlongm:
The history of adding erlang/rabbitmq/haproxy repos was for features that were not in then-current versions. There are currently no outstanding features in the non-standard packages.

The reasons have been outlined above, these are unused and untested code blocks, that can be easily added back if they are needed.

For a technical/practical reason: with these repos enabled, chef-bcpc fails to build.

bcpc:
@furlongm #1412 rebased on feature/v8-bionic with multihead with 1412 <https://github.com/bloomberg/chef-bcpc/pull/1412|use bionic repos for haproxy, rabbitmq and erlang> by @furlongm failed during converge <https://cloud-infra.jaas.dev.bloomberg.com/job/bcpc/job/jenkins/job/master/756/|Jenkins log>

bcpc:
@furlongm #1412 rebased on feature/v8-bionic with multihead with 1412 <https://github.com/bloomberg/chef-bcpc/pull/1412|use bionic repos for haproxy, rabbitmq and erlang> by @furlongm failed during converge <https://cloud-infra.jaas.dev.bloomberg.com/job/bcpc/job/jenkins/job/master/758/|Jenkins log>

bcpc:
@furlongm #1412 rebased on feature/v8-bionic with singlehead with 1412 <https://github.com/bloomberg/chef-bcpc/pull/1412|use bionic repos for haproxy, rabbitmq and erlang> by @furlongm failed during converge <https://cloud-infra.jaas.dev.bloomberg.com/job/bcpc/job/jenkins/job/master/759/|Jenkins log>

justinjpacheco:
these build errors are for v7 not v8

```
TASK [deployment : Cleanup VMs] ************************************************
fatal: [120-lab-cloud-r01n06.lab.bcc.bloomberg.com]: FAILED! => {"changed": true, "cmd": "REPO_ROOT=/home/infra_jenkins/builds/chef-bcpc VAGRANT_VAGRANTFILE=/home/infra_jenkins/builds/chef-bcpc/bootstrap/vagrant_scripts/Vagrantfile_todestroy /bin/bash -lc ./vagrant_clean.sh", "delta": "0:00:00.006868", "end": "2018-09-18 12:57:43.053942", "msg": "non-zero return code", "rc": 127, "start": "2018-09-18 12:57:43.047074", "stderr": "/bin/bash: ./vagrant_clean.sh: No such file or directory", "stderr_lines": ["/bin/bash: ./vagrant_clean.sh: No such file or directory"], "stdout": "", "stdout_lines": []}
...ignoring

TASK [deployment : Run deployment script] **************************************
fatal: [120-lab-cloud-r01n06.lab.bcc.bloomberg.com]: FAILED! => {"changed": true, "cmd": "/bin/bash -lc ./BOOT_GO.sh", "delta": "0:00:00.006172", "end": "2018-09-18 12:57:43.687963", "msg": "non-zero return code", "rc": 127, "start": "2018-09-18 12:57:43.681791", "stderr": "/bin/bash: ./BOOT_GO.sh: No such file or directory", "stderr_lines": ["/bin/bash: ./BOOT_GO.sh: No such file or directory"], "stdout": "", "stdout_lines": []}
```

it's failing because those scripts no long exist, not because they are using newer versions of haproxy,rabbitmq or erlang

justinjpacheco:
> The history of adding erlang/rabbitmq/haproxy repos was for features that were not in then-current versions. There are currently no outstanding features that we need in the non-standard packages.

The history of adding those repos is to use newer versions that fixed bugs, added features or made performance improvements. These are all possible once our version of those software packages gets older as time moves on and we remain to the current LTS (18.04)

> The reasons have been outlined above, these are unused and untested code blocks, that can be easily added back if they are needed.

It has been used and it has been tested, then it was turned off. I think I said as much above. How did you come to this conclusion?

mattp-:
it seems reasonable to provide an avenue to override repo source; considering even internally mirror servers are used to work around firewall issues and upstream dependence on the existence of said mirrors (which over time may be deleted, etc).

furlongm:
I was not referring to the build errors above.

Building v8 with these toggles enabled does not succeed. I came to the conclusion that they are unused and untested codeblocks precisely because nobody has used or tested them in v8. If they had, someone would have noticed that chef-bcpc fails to build with them enabled.

justinjpacheco:
so the only option is to remove the feature entirely? also, what was the source of the error?

mihalis68:
This is an override mechanism. It is not expected that currently unused overrides should build successfully if you enable them for no reason. If they don't build, that does NOT prove that the override mechanism is useless. 

Also, I question calling the non-canonical packages "non-standard". That becomes a fairly loaded term here. The standard version of any package is whatever we deem to be the preferred one right now. It may well come from Canonical today and some other upstream at a later date. I actually think it unlikely that we'll stay as purely bionic based for very long.

furlongm:
I did not investigate the error as the code was clearly unused and untested. The other option was to spend time maintaining unused and untested code.

Please feel free to close this PR, as it's clearly a source of contention, and I was only trying to clean up our codebase.

PR #1413, opened 2018-08-28
dns zone creation/fill, misc fixes/updates

justinjpacheco:
- dns zone creation/fill
    - create zone for cloud domain
    - populate zone with ip information from fixed networks defined in neutron

  - designate
    - removed from runlist
    - very poor performance when importing >1k recordset zone file

  - ansible
    - register-compute-nodes task
      - updated selector to use '==' vs. 'contains' for exact matching

  - keystone
    - update admin email attribute to use domain variable

  - neutron
    - added support for floating ips
    - refactored 'networks' strucuture to support mutliple fixed/float subnets
      per network

  - unbound
    - updated cloud_domain forward-zone to point to powerdns over designate

  - powerdns
    - operating as master

### DIFF
--- a/chef/cookbooks/bcpc/templates/default/powerdns/zone.erb
+++ b/chef/cookbooks/bcpc/templates/default/powerdns/zone.erb
@@ -1,0 +1,13 @@
+$ORIGIN <%= node['bcpc']['cloud']['domain'] %>.
+$TTL 3600
+
+@  IN SOA ns1.<%= node['bcpc']['cloud']['domain'] %>. <%= @admin_email %>. <%= @serial_number %> 3559 600 86400 3600
+@  IN NS ns1.<%= node['bcpc']['cloud']['domain'] %>.
+
+<% @networks.each do |network| %>
+<% network['cidrs'].each do |cidr| %>
+<% cidr.each_host do |ip| %>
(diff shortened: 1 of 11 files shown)


chef/cookbooks/bcpc/templates/default/powerdns/zone.erb
```diff
@@ -0,0 +1,13 @@
+$ORIGIN <%= node['bcpc']['cloud']['domain'] %>.
+$TTL 3600
+
+@  IN SOA ns1.<%= node['bcpc']['cloud']['domain'] %>. <%= @admin_email %>. <%= @serial_number %> 3559 600 86400 3600
+@  IN NS ns1.<%= node['bcpc']['cloud']['domain'] %>.
+
+<% @networks.each do |network| %>
+<% network['cidrs'].each do |cidr| %>
+<% cidr.each_host do |ip| %>
+<%= network['name'] %>-<%= ip.octets.join('-') %> IN A <%= ip %>
```
comay: Should we be adding associated PTR records here?

comay: Well in particular, should we be creating a zone (or zones) for the corresponding <x>.<y>[.<z>].in-addr.arpa domains and populating the corresponding PTR records under it?

PR #1430, opened 2018-10-06
Add bcpc-extra cookbook to v8

srsakhamuri:
Add bcpc-extra for rally and other misc recipes


justinjpacheco:
bcpc extra has been imported into chef-bcpc. this cookbook is no longer required

PR #1435, opened 2018-10-11
haproxy configuration refactor, misc fixes

justinjpacheco:
- haproxy
    - update haproxy to load configuration fragments from /etc/haproxy/haproxy.d
    - updated each openstack service to provide it's own haproxy configuration

  - misc fixes
    - os quota fix
      - prepended quota limit grep with -- so values such as -1 would not be
        treated as command switches to grep
    - horizon
      - removed policy file/directory creation


justinjpacheco:
sure

PR #1440, opened 2018-10-15
powerdns, horizon, neutron, calico and cinder updates

justinjpacheco:
- powerdns
    - bind to local anycast address vs. physical interface ip
  - horizon
    - domain dropdown will now show first domain as option
    - preconfigured security groups for security group downdown
    - removed commented material design theme option
  - cinder
    - add retry for ceph backend property modification
  - calico
    - removed unused calico-felix.service file
  - neutron
    - improved cloud and admin project default security group alteration


chef/cookbooks/bcpc/recipes/neutron-head.rb
```diff
@@ -320,40 +320,27 @@
...
+      fi
 
-  code <<-DOC
-    project=#{node['bcpc']['openstack']['admin']['project']}
-    sec_groups=$(openstack security group list -f value --project ${project})
-    group_id=$(echo ${sec_groups} | grep default | awk '{print $1}')
-    openstack security group rule create --proto tcp --dst-port 22 $group_id
+      if ! openstack security group rule list --protocol tcp $group_id | grep '22:22'; then
+        openstack security group rule create $group_id \
+          --protocol tcp --dst-port 22:22 --remote-ip 0.0.0.0/0
```
comay: Shouldn't we also be providing the IPv6-equivalent security group rule here?

Also, checking against the JSON output might be better in case the output includes an IPv6 address that includes the substring "22:22" in the "IP Range" field.

justinjpacheco: this is how we were modifying it before, we can improve it after this version is merged. i'd like to unblock unrelated dependents

justinjpacheco: also, we're not currently using ipv6

comay: Both, good points. I guess they can wait for the time we support IPv6 (tbh, I didn't realize that bcpc didn't support IPv6!)

PR #1451, opened 2018-11-27
new chef/ansible role: storagenode

justinjpacheco:
- storagenode
    - provides ceph storage

  - Makefile
    - added new target to chef storagenodes

  - ansible
    - added storagenode playbooks and chef role generation

  - chef
    - nova, glance, cinder
      - updated recipes to use ceph auth imports for user creation
    - nova-compute
      - added ceph to list of required packages

  - vagrant
    - new topology file: 1h1w1s.yml
      - defines a dedicated storage node


justinjpacheco:
> ... recipes can call to do the ceph-specific creation?

which creation are you referring to?

> And to confirm, with this change the ability to use the compute node for OSDs is removed? Is that something we want to preserve as an option?

compute nodes can still act as storage nodes provided they have the proper role. see 1h1w.yml

> does it make sense to have the ceph-mon move away from the headnodes,

i thought about that and it is certainly possible, we should handle that in another commit imo

comay:
By ceoh-specific creation, I mean the pool creation. It seems the same block of code (modulo name changes) is being used with the cinder, glance, and nova recipes and I'm wondering whether or not having this in some sort of function in the libraries directory would make sense.

comay:
Also, missed the update to 1h1w.yml - can the other "legacy" configurations be updated as well to add the storage rule to their worknodes?

PR #1466, opened 2019-01-28
Tunnel support

mihalis68:
Rudimentary tunnel support so you can load horizon from the host after doing a build by simply

```
make tunnel
```

from the root dir after make create all.

### DIFF
--- a/virtual/bin/make-tunnel.sh
+++ b/virtual/bin/make-tunnel.sh
@@ -31,9 +31,21 @@ fi
 
 cd $virtual_dir
 
-vagrant ssh-config ${bootstrap_name} > /tmp/ssh-config.$$
+CONFFILE="/tmp/ssh-config.$$"
+vagrant ssh-config ${bootstrap_name} > $CONFFILE
 
-sudo ssh -C -F /tmp/ssh-config.$$ \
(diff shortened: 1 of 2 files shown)


virtual/bin/make-tunnel.sh
```diff
@@ -0,0 +1,39 @@
...
+else
+    host_ip=`hostname -i`
+fi
+
+cd $virtual_dir
+
+vagrant ssh-config ${bootstrap_name} > /tmp/ssh-config.$$
+
+sudo ssh -C -F /tmp/ssh-config.$$ \
+     -L ${host_ip}:8443:10.10.254.254:443 \
```
justinjpacheco: is there any way we can get the vip dynamically? this value will change depending on the environment

mihalis68: Probably. I'll look into that

kamidzi: I maintain ports should not be hardcoded and the port forwarding should be dynamic instead; in order to use horizon, one would then need to configure the ssh connection as the socks proxy. This way any dns resolutions should work reliably within horizon.

mihalis68:
Replaced by https://github.com/bloomberg/chef-bcpc/pull/1487 which is better

PR #1469, opened 2019-02-01
Timed make targets

mihalis68:
There is no test procedure per se. All make targets are timed and the results are always printed. Of course this could be made more refined and/or optional.

Sample output (the last line of text in the following make invocation) : 

```
$ make destroy-virtual-network
virtual/bin/destroy-virtual-network.sh
++ git rev-parse --show-toplevel
+ root_dir=/Users/cmorgan/work/v8/submission
+ virtual_dir=/Users/cmorgan/work/v8/submission/virtual
+ network_dir=/Users/cmorgan/work/v8/submission/virtual/network
+ cd /Users/cmorgan/work/v8/submission/virtual/network
+ vagrant destroy -f
==> spine2: VM not created. Moving on...
==> spine1: VM not created. Moving on...
==> tor3: VM not created. Moving on...
==> tor2: VM not created. Moving on...
==> tor1: VM not created. Moving on...
        2.49 real         2.09 user         0.18 sys
```


rusage
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
justinjpacheco: i think the location of this file shouldn't be in the root directory. possibly something like /bin?

Makefile
```diff
@@ -8,6 +8,8 @@ export ANSIBLE_CONFIG = ansible/ansible.cfg
 headnodes = $$(ansible headnodes -i ${inventory} --list | tail -n +2 | wc -l)
 storagenodes = $$(ansible storagenodes -i ${inventory} --list | tail -n +2 | wc -l)
 
+SHELL = ./rusage sh
```
justinjpacheco: we might want to ensure that make is always able to find this script regardless of it's location in the filesystem

rusage
mihalis68: I'm open to suggestions however there's no local bin, other than virtual/bin, but this change does not depend on the build being virtual. I don't want to assume we can install this in the developers own bin dir somewhere.

mihalis68:
test this techique in a minimal setup with makefile includes:
```
/Users/cmorgan/maketest $ ls -l
total 8
-rw-r--r--  1 cmorgan  domain users  153 Feb  6 14:03 Makefile
drwxr-xr-x  3 cmorgan  domain users  102 Feb  6 14:11 bin
drwxr-xr-x  3 cmorgan  domain users  102 Feb  6 14:02 sub
/Users/cmorgan/maketest $ more Makefile 
HERE = $(dir $(realpath $(lastword $(MAKEFILE_LIST))))

SHELL = $(HERE)/bin/rusage sh


A :
        echo A
        $(warning $(HERE))

B :
        echo B
        $(warning $(HERE))
/Users/cmorgan/maketest $ ls bin
rusage
/Users/cmorgan/maketest $ more bin/rusage 
#!/bin/sh
exec /usr/bin/time "$@"
/Users/cmorgan/maketest $ ls sub
Makefile
/Users/cmorgan/maketest $ more sub/Makefile 
include ../Makefile

C :
        echo C
        $(warning $(HERE))
/Users/cmorgan/maketest $ make A B C
Makefile:7: /Users/cmorgan/maketest/
echo A
A
        0.00 real         0.00 user         0.00 sys
Makefile:11: /Users/cmorgan/maketest/
echo B
B
        0.00 real         0.00 user         0.00 sys
make: *** No rule to make target `C'.  Stop.
/Users/cmorgan/maketest $ cd sub
/Users/cmorgan/maketest/sub $ make A B C
../Makefile:7: /Users/cmorgan/maketest/
echo A
A
        0.00 real         0.00 user         0.00 sys
../Makefile:11: /Users/cmorgan/maketest/
echo B
B
        0.00 real         0.00 user         0.00 sys
Makefile:4: /Users/cmorgan/maketest/
echo C
C
        0.00 real         0.00 user         0.00 sys
/Users/cmorgan/maketest/sub $ 

```

bin/rusage
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
comay: I realize the shell script is trivial but shouldn't it include the standard copyright/license block as seen in files such as `virtual/network/proxyconfig.sh`?

mihalis68: Yes, will fix

mihalis68: Fixed

mihalis68:
This turned out to be not very useful. Sorry to occupy the reviewers time.

PR #1481, opened 2019-02-26
all in one network vm

justinjpacheco:
- non leaf/spine network implementation that provides bgp network
    functionality without requiring multiple vms

  - drop-in replacement for current network configuration
  - disabled by default.

  - enabled via environment variable:

    ```
    export DEPLOY_NETWORK_VM=true
    make create-virtual-network
    ```


mihalis68:
Could we have a way to persistently chose the all-in-one networking vm with some setting in a file?

justinjpacheco:
for the time being i've solved that problem by putting the environment variable in my .bashrc

mihalis68:
I like to be able to have multiple build areas which are self-contained.

justinjpacheco:
i see. we don't have a facility yet for that kind of functionality.

virtual/network/Vagrantfile
```diff
@@ -11,68 +11,75 @@ def setup_proxy(node)
...
-      node.vm.network "private_network", virtualbox__intnet: "tor1_spine#{i}",
-       auto_config: false
-      node.vm.network "private_network", virtualbox__intnet: "tor2_spine#{i}",
-       auto_config: false
-      node.vm.network "private_network", virtualbox__intnet: "tor3_spine#{i}",
-       auto_config: false
+
+      node.vm.network "private_network", virtualbox__intnet: "management1", auto_config: false
+      node.vm.network "private_network", virtualbox__intnet: "management2", auto_config: false
+      node.vm.network "private_network", virtualbox__intnet: "management3", auto_config: false
```
comay: Could you break these >80 character lines as was done below? Namely,

      node.vm.network "private_network", virtualbox__intnet: "management1",
       auto_config: false
      node.vm.network "private_network", virtualbox__intnet: "management2",
       auto_config: false
      node.vm.network "private_network", virtualbox__intnet: "management3",
       auto_config: false

virtual/network/bird/network.conf
```diff
@@ -0,0 +1,138 @@
...
+        if net ~ mynetworks_nets then accept;
+        if net = DEFAULT then accept;
+        reject;
+}
+
+# The Kernel protocol is not a real routing protocol. Instead of communicating
+# with other routers in the network, it performs synchronization of BIRD's
+# routing tables with the OS kernel.
+protocol kernel {
+        learn;
```
comay: Could you re-add the

# necessary to include DHCP-derived default route

comment before this line as seen in the spine configuration files?

PR #1489, opened 2019-03-14
Script to force clean the VMs

srsakhamuri:


justinjpacheco:
what's the purpose of this script? we already have code that removes vms via vagrant

srsakhamuri:
Vagrant doesn't seem to work sometimes when the make create fails for some reason, and if the builds terminated for any reason, i ran in to that situation several times locally and with jenkins. so these commands help bring it to clean state.

justinjpacheco:
I think it would be much better to understand why that failure is happening and solve the problem properly vs. adding another vm cleanup script (a much more destructive one btw) to our code base just because something happens sometimes for unknown reasons.

srsakhamuri:
I think we did encounter this on xob afaik, and when switching between builds (multi/single) in jenkins. And also when destroy fails leaving vms around. Yes it is destructive, that is purpose of this script. When vagrant destroy fails sometimes this is the only option. Since I ran into this several times rather than remembering the commands a script like this comes handy.

justinjpacheco:
>  and when switching between builds (multi/single) in jenkins. And also when destroy fails leaving vms around
if you're switching between single and multi, then you should make you're properly destroying your environment before moving on.  if the current cleanup script isn't work, then make it better. adding another script to cover-up a problem happening in previous steps isn't the way this should be solved.

> Since I ran into this several times rather than remembering the commands a script like this comes handy.
that's fine that you have a solution for a corner case, by why does that mean it should be committed to our code base? you can store these files in github gists. the codebase isn't the appropriate place for one-offs.

srsakhamuri:
Well I can either put this script as part of jenkins build repo or here. Since this happened more than once and require manual intervention, it just the nature of vagrant/virtualbox integration. Will let you know when i run into this again.

justinjpacheco:
scripts of this nature aren't really appropriate for this codebase.

furthermore, one-liner solutions that pkill virtualbox process indiscriminately because the build server doesn't clean up the env properly is a **build server** problem. **build server** solutions should say with the **build server**. we can't be expected to maintain a plethora of random **build server** fix-it scripts for anyone/everyone just because they run into every problem under the sun. that's not the responsibility of this project/codebase.

srsakhamuri:
One off scripts are not new to this repo, we did have several https://github.com/bloomberg/chef-bcpc/blob/master/bootstrap/ansible_scripts/scripts/vbox_clean.sh
And this
https://github.com/bloomberg/chef-bcpc/blob/master/bootstrap/vagrant_scripts/dump_config.sh

The pkill is needed because the VBoxManage uregistervm failed to clean it up. I am not particular where this lives, and this is not specific to build server, I am sure we have seen this on xob and our local machines.

mihalis68:
I think this would be better with the jenkins pipeline stuff. It's a bit dangerous/brutal to go into chef-bcpc but I understand you need to have it for wedged jenkins builds from time to time.

PR #1498, opened 2019-03-25
calico 3.6

justinjpacheco:
- updated calico to 3.6
  - updated calicoctl to 3.6
  - updated admin default security group to include ssh, http and https

### DIFF
--- a/chef/cookbooks/bcpc/recipes/neutron-head.rb
+++ b/chef/cookbooks/bcpc/recipes/neutron-head.rb
@@ -345,9 +345,11 @@ bash 'update admin default security group' do
       if ! openstack security group rule list ${sec_id} \
             --protocol icmp \
             --long -c Ethertype -f value | grep -q ${ethertype}; then
+
         openstack security group rule create ${sec_id} \
           --protocol icmp \
           --ethertype ${ethertype}
+
       fi
 
       # allow ssh, http and https
--- a/chef/cookbooks/bcpc/recipes/neutron-head.rb
+++ b/chef/cookbooks/bcpc/recipes/neutron-head.rb
@@ -339,7 +339,7 @@ bash 'update admin default security group' do
 
     sec_groups=$(openstack security group list --project ${id} -f json)
     sec_id=$(echo ${sec_groups} | \
-      jq -r --arg id "${id}" '.[] | select(.Name == "Default") .ID')
+      jq -r --arg id "${id}" '.[] | select(.Name == "default") .ID')
 
     # allow icmp
     if ! openstack security group rule list ${sec_id} | grep -q icmp; then
--- a/chef/cookbooks/bcpc/recipes/neutron-head.rb
+++ b/chef/cookbooks/bcpc/recipes/neutron-head.rb
@@ -338,8 +338,7 @@ bash 'update admin default security group' do
     id=$(openstack project show ${admin_project} -f value -c id)
 
     sec_groups=$(openstack security group list --project ${id} -f json)
-    sec_id=$(echo ${sec_groups} | \
-      jq -r --arg id "${id}" '.[] | select(.Name == "default") .ID')
+    sec_id=$(echo ${sec_groups} | jq -r '.[] | select(.Name == "default") .ID')
 
     # allow icmp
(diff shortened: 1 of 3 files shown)


chef/cookbooks/bcpc/recipes/neutron-head.rb
```diff
@@ -329,26 +329,30 @@
...
     admin_project=#{node['bcpc']['openstack']['admin']['project']}
-    admin_project_id=$(openstack project show ${admin_project} -f value -c id)
+    id=$(openstack project show ${admin_project} -f value -c id)
 
-    for id in ${cloud_project_id} ${admin_project_id}; do
-      sec_groups=$(openstack security group list -f json)
-      group_id=$(echo ${sec_groups} | jq -r --arg id "${id}" '.[] | select(.Project == $id) .ID')
+    sec_groups=$(openstack security group list -f json)
+    sec_id=$(echo ${sec_groups} | \
+      jq -r --arg id "${id}" '.[] | select(.Project == $id) .ID')
```
furlongm: ```openstack security group list --project ${id} -f value -c ID``` ?

justinjpacheco: updated

furlongm: jq not required at all if using `openstack security group list --project ${id} -f value -c ID`

justinjpacheco: there can be multiple security groups defined for a given project. The one named "Default" is the target

furlongm: Ah I see. Pity there's no `--name` option to `list`, and `show` doesn't have `--project` option.

Is it `Default` or `default`?

justinjpacheco: yah, that would be a handy option. updated to use the correct name, nice catch

chef/cookbooks/bcpc/recipes/neutron-head.rb
```diff
@@ -329,26 +329,29 @@
...
-      group_id=$(echo ${sec_groups} | jq -r --arg id "${id}" '.[] | select(.Project == $id) .ID')
+    sec_groups=$(openstack security group list --project ${id} -f json)
+    sec_id=$(echo ${sec_groups} | jq -r '.[] | select(.Name == "default") .ID')
 
-      if ! openstack security group rule list $group_id | grep icmp; then
-        openstack security group rule create --protocol icmp $group_id
-      fi
+    # allow icmp
+    if ! openstack security group rule list ${sec_id} | grep -q icmp; then
+      openstack security group rule create ${sec_id} --protocol icmp
```
comay: I realize we don't yet have IPv6 deployed in our infrastructure but Neutron does create IPv6 compliant base egress and ingress rules automatically. so I did the same in my recent change. Anyway, since you mentioned one of the goals was consistency.

chef/cookbooks/bcpc/recipes/neutron-head.rb
```diff
@@ -329,26 +329,29 @@
...
+    fi
+
+    # allow ssh, http and https
+    for port_range in 22:22 80:80 443:443; do
```
comay: The `--dst-port` argument automatically sets `port_range_min` and `port_range_max` by just specifying a simple integer, so you can change the list to just `22 80 443`.

chef/cookbooks/bcpc/recipes/neutron-head.rb
```diff
@@ -329,26 +329,29 @@
...
+      if ! openstack security group rule list ${sec_id} --protocol tcp \
+            -c "Port Range" -f value | grep -q "${port_range}"; then
 
-      if ! openstack security group rule list --protocol tcp $group_id | grep '22:22'; then
-        openstack security group rule create $group_id \
-          --protocol tcp --dst-port 22:22 --remote-ip 0.0.0.0/0
+        openstack security group rule create ${sec_id} \
+          --protocol tcp --dst-port ${port_range} --remote-ip 0.0.0.0/0
```
comay: Same comment as above regarding IPv6 - just a comment not a must-do.

PR #1510, opened 2019-05-16
Feature/v8 bionic rebased

justinjpacheco:


ansible/playbooks/roles/common/filter_plugins/util.py
```diff
@@ -0,0 +1,106 @@
+from __future__ import (absolute_import, division, print_function)
```
comay: As this file is code, it should include the standard copyright/license block as in `virtual/bin/generate-ansible-inventory.py`.

PR #1524, opened 2019-06-06
add tunnel support

furlongm:
Alternative tunnel implementation


furlongm:
commit included in #1529

PR #1531, opened 2019-06-12
add support for openstack heat

furlongm:


chef/cookbooks/bcpc/templates/default/heat/heat.conf.erb
```diff
@@ -0,0 +1,2646 @@
+[DEFAULT]
```
justinjpacheco: could we remove all the options not used?

justinjpacheco:
this looks good so far, i'll do a build and try it out. do you have some steps i can use to try this feature out?

furlongm:
Create the template `demo.yml` with the following content:

```
heat_template_version: 2015-10-15
description: Launch a basic instance with CirrOS image and print instance name and IP

resources:
  server:
    type: OS::Nova::Server
    properties:
      image: cirros
      flavor: generic1.tiny
      networks:
      - network: ext1

outputs:
  instance_name:
    description: Name of the instance.
    value: { get_attr: [ server, name ] }
  instance_ip:
    description: IP address of the instance.
    value: { get_attr: [ server, first_address ] }
```

Then create the stack from the template:

```shell
$ openstack stack create -t demo.yml demo1
```

PR #1540, opened 2019-06-28
configure libvirt to use ssh

justinjpacheco:
- configure nova-compute to use the qemu+ssh schema when executing
    live migrations
  - configure libvirt to no longer listen for incoming tcp connections


furlongm:
Does this affect the time to live migrate an instance?

justinjpacheco:
i'll test with the largest flavor available, but it seems to be as fast as the default method

justinjpacheco:
@comay yah, im not sure

justinjpacheco:
@furlongm migration time seems to be unaffected. i tested with the largest flavor between two hypervisors, they appear to be the same

furlongm:
Line 18 seems to have an extra space at the start of the line, perhaps that's throwing things off? Although, it should just be showing an extra space there in the diff?

justinjpacheco:
@furlongm doesn't look like it made a difference :(

furlongm:
Actually it seems like a bunch of whitespace has been added at the end of those lines too.

furlongm:
Looks like a tab at the end of each line.

justinjpacheco:
@furlongm fixed

PR #1547, opened 2019-07-02
simplify rally installation

furlongm:


justinjpacheco:
i believe the later version of rally provided by pip is required which is why it via pip vs. apt

furlongm:
> i believe the later version of rally provided by pip is required which is why it via pip and not apt

Installation now uses pip again, modified to use yaml config, create a rally homedir and be ready to run tempest test suite.

PR #1549, opened 2019-07-03
reorder network configuration tasks

furlongm:


ansible/playbooks/roles/common/tasks/configure-networking.yml
```diff
@@ -89,25 +97,17 @@
...
-    state: link
-    force: true
-  register: resolved
-
-- name: restart systemd-resolved
-  service:
-    name: systemd-resolved
-    state: restarted
-  when: resolved.changed
+- name: check for default gateway from tor
```
justinjpacheco: verification of a valid bird configuration is required before applying the netplan config. this is to prevent severing your connection to the transit interfaces in the event that a gateway from the tors cannot be established via bird.

PR #1556, opened 2019-07-11
wait for heat api after restarting either heat api

furlongm:


chef/cookbooks/bcpc/recipes/heat.rb
```diff
@@ -315,7 +311,6 @@
     headnodes: headnodes(all: true),
     vip: node['bcpc']['cloud']['vip']
   )
-  notifies :restart, 'service[heat-api-apache2]', :immediately
-  notifies :restart, 'service[heat-api-cfn-apache2]', :immediately
+  notifies :restart, 'service[heat-apis-apache2]', :delayed
```
justinjpacheco: you don't want to restart apache immediately after a heat.conf change?

furlongm: I do, but there's an ordering issue. If I restart here immediately, the apache wsgi configs are not in place yet, so waiting for heat api never succeeds.

furlongm:
Updated dependency chaining in order to not delay heat-api restarts.

PR #1557, opened 2019-07-12
added updated nova-network model.py that includes wait method

justinjpacheco:


srsakhamuri:
@justinjpacheco I think we should use patch here?

justinjpacheco:
using patch now

PR #1560, opened 2019-07-19
add global default workers variable

furlongm:
This PR adds to the use of the existing `default['bcpc']['openstack']['services']['workers']` variable. Currently this variable is not used by default by all projects.

This commit changes that so `default['bcpc']['openstack']['services']['workers']` is used by default unless a project-specific workers value is defined. If a project-specific variable, e.g. `default['bcpc']['cinder']['workers']`, is set, then it has priority. This gives us consistency over how workers are defined in each project, and also allows for fine-tuning the workers on a per-project basis.

The default for `default['bcpc']['openstack']['services']['workers']` is now set to `1`, which reduces the RAM used by r1n1 in the open source build (reduced to 5G now), and speeds up build time.


chef/cookbooks/bcpc/attributes/openstack.rb
```diff
@@ -12,7 +12,7 @@
 default['bcpc']['openstack']['admin']['username'] = 'admin'
 default['bcpc']['openstack']['admin']['project'] = 'admin'
 
-default['bcpc']['openstack']['services']['workers'] = nil
+default['bcpc']['openstack']['services']['workers'] = 1
```
srsakhamuri: is this going to be overridden in each cluster config
it is currently set to 1
which is very low number of workers?

furlongm: Yes, it will be overridden. The default is just for the vagrant build - where it reduces memory

PR #1565, opened 2019-07-22
add support for tinyproxy

justinjpacheco:


ansible/playbooks/roles/common/templates/tinyproxy/tinyproxy.conf.j2
```diff
@@ -0,0 +1,172 @@
...
+# The LAST matching rule wins the route decision.  As you can see, you
+# can use a host, or a domain:
+#  name     matches host exactly
+#  .name    matches any host in domain "name"
+#  .        matches any host with no domain (in 'empty' domain)
+#  IP/bits  matches network/mask
+#  IP/mask  matches network/mask
+
+{% for upstream in local_proxy['config']['upstream'] %}
+Upstream {{ upstream }}
```
furlongm: Should this be lowercase `upstream` per the examples?

Not sure if it makes any difference, but it might.

ansible/playbooks/roles/common/templates/tinyproxy/tinyproxy.conf.j2
```diff
@@ -0,0 +1,172 @@
...
+#  upstream testproxy:8008 ".test.domain.invalid"
+#  upstream testproxy:8008 ".our_testbed.example.com"
+#  upstream testproxy:8008 "192.168.128.0/255.255.254.0"
+#
+#  # no upstream proxy for internal websites and unqualified hosts
+#  no upstream ".internal.example.com"
+#  no upstream "www.example.com"
+#  no upstream "10.0.0.0/8"
+#  no upstream "192.168.0.0/255.255.254.0"
+#  no upstream "."
```
furlongm: Should we add a `no upstream` for the local cloud domain?

ansible/playbooks/roles/common/templates/tinyproxy/tinyproxy.conf.j2
justinjpacheco: i went back and forth on this. {U,u}stream works, but the example in the original file uses Upstream for its example of the default route and all the other config params use the uppercase version as well 🤷‍♂

PR #1575, opened 2019-08-07
tinyproxy.conf removed connectport config

justinjpacheco:


justinjpacheco:
it allows tinproxy to accept ssl connections on whatever port was requested. this helps tinyproxy expand its use to handling requests for keystone (port 35357) and other openstack services.

srsakhamuri:
LGTM, seems convenient to allow all the ports

PR #1584, opened 2019-08-23
libvirt build support

furlongm:
This PR adds support for building chef-bcpc vagrant-libvirt provider.

This allows guest VMs to run with nested kvm, which means no longer having emulated performance.

With longer build timeouts on travis, it will also allow for full builds on each commit (using qemu though, no nested kvm on travis).

It also has the benefit of allowing builds of other vagrant-libvirt projects on the same machine (which doesn't work when the virtualbox drivers are loaded).

Building with the vagrant-libvirt proivider is an option, the default remains building with the virtualbox provider.

To enable libvirt, use the following commands:

```
export VAGRANT_DEFAULT_PROVIDER=libvirt
make create all
```

Note that building with the vagrant-libvirt provider only builds a single network VM.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ language: generic
 matrix:
   include:
     - os: linux
-      dist: xenial
+      dist: bionic
       sudo: required
     - os: osx
-      osx_image: xcode10.2
+      osx_image: xcode11
 
 before_install: .travis/install-deps.sh
 
--- a/.travis/install-deps.sh
+++ b/.travis/install-deps.sh
@@ -1,6 +1,18 @@
 #!/bin/bash -x
 
-print_debug_info () {
+print_debug_info_osx () {
+    printenv
+    sysctl machdep.cpu
+    kextstat
+    sudo dmesg
+    vm_stat
+    df -h
+    diskutil list
+    ls -la
+    groups
+}
+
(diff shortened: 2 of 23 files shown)


comay:
Is the hard-coding of the network addresses in virtual/network/Vagrantfile.libvirt actually necessary? It would be nice if there was one source of truth between that and the files under netplan/*.

Also, I'd prefer to see parity between the network type (traditional leaf-spine or combined) across the virtualization methods. If everyone is happy with using combined, perhaps the traditional one should be removed at this point. Alternatively, I'd be happy to see if left in but with support via the more-efficient `libvirt` mechanism.

furlongm:
I rarely use the traditional leaf spine build any more, it takes a lot longer to build and uses up more resources. So I just left it out for simplicity's sake and because I almost never use it.

For the hard-coded network addresses, we could parse the netplan files? Alternatively, we could include a network group in the topology and/or integrate the network stuff as an optional extra to the top-level Vagrantfile?

furlongm:
@comay - now loading the network topology from netplan yaml.

For the multi-vm leaf-spine builds, I think that is a bigger question, and can be pushed to a different PR if it is required? In any case, it should probably go into a different Vagrantfile, just to avoid unnecessary complexity in the Vagrantfile in this commit?

PR #1591, opened 2019-08-30
add bootstrap deb repo with ansible repo installation example

furlongm:
This PR creates an apt repo on the bootstrap sever, adds the chef debs into it, and adds the repo to apt on all nodes. It replaces the file-based deb installation with a repo-based installation of the chef debs as an example.

With this basis, we can convert all existing repos in the chef cookbooks to install during this ansible stage, so that all apt repos are configured by the time the chef client starts.


Makefile
```diff
@@ -59,6 +59,12 @@ sync-assets :
 		-i ${inventory} ${playbooks}/site.yml \
 		-t sync-assets --limit localhost
 
+configure-apt-repos :
```
justinjpacheco: can we make `configure-apt-repos` singular seeing as how it's only configuring a single apt repo?

justinjpacheco: also, shouldn't this target be added to the rest of the targets under `make all`?

justinjpacheco: i think we'll need to rethink how we're configuring apt as a whole with this change. we're currently configuring the network aspects of apt in the `configure-networking` target and i think it would be best to have `apt` configured in a central place

ansible/playbooks/roles/common/tasks/main.yml
```diff
@@ -20,3 +20,9 @@
   vars:
     ansible_host: "{{ interfaces['transit'] | primary_ip(hostvars[inventory_hostname]) }}"
   tags: [never,configure-networking]
+
+- import_tasks: configure-apt-repos.yml
+  become: true
+  vars:
```
justinjpacheco: this is not required after the `configure-networking` task because the service ip is now available

ansible/playbooks/roles/common/defaults/main/base.yml
```diff
@@ -211,3 +220,11 @@ virtualbox:
 ###############################################################################
 
 host_aggregates: []
+
+###############################################################################
+# bootstrap
+###############################################################################
+
+bootstrap_host: "{{ groups['bootstraps'][0] }}"
+bootstrap_ip: "{{ hostvars[bootstrap_host][service][ip] | ipaddr('address') }}"
```
justinjpacheco: you don't need to grab the service ip directly because the host ip is the service ip and it will be available by the time the bootstrap related tasks are running.

Makefile
furlongm: Yep, it should be under `make all`, will add it there.

Ideally the apt repos would be configured before networking, including the proxy config.

justinjpacheco: we can put all the apt related stuff into an apt yml and import the task under tiny proxy in the configure-networking task

ansible/playbooks/roles/common/defaults/main/base.yml
furlongm: Just copied these lines from the `chef_server_` vars. Should it be changed there too?

ansible/playbooks/roles/common/tasks/main.yml
furlongm: If we move the apt-repos configuration before networking configuration, we will need it though?

ansible/playbooks/roles/common/defaults/main/base.yml
justinjpacheco: yah, makes sense

ansible/playbooks/roles/common/tasks/main.yml
justinjpacheco: correct, i think we can rename this task to `configure-apt.yml` and use it in the `configure-networking` task via `import_tasks` under the `tinyproxy` configuration

Makefile
```diff
@@ -13,6 +13,7 @@ all : \
 	sync-assets \
 	configure-operator \
 	configure-networking \
+	configure-apt-repos \
```
justinjpacheco: there are debs that are installed via apt in the `configure-networking` task (eg. bird,tinyproxy). so we'll need to have apt configured before those packages installed, except for tinyproxy which is in a bit of a chicken and egg situation.

Makefile
```diff
@@ -59,6 +59,12 @@ sync-assets :
 		-i ${inventory} ${playbooks}/site.yml \
 		-t sync-assets --limit localhost
 
+configure-apt :
```
justinjpacheco: is this target still needed? the `configure-apt` ansible tag doesn't exist anymore

Makefile
```diff
@@ -12,11 +12,11 @@ storagenodes = \
 all : \
 	sync-assets \
 	configure-operator \
+	configure-web-server \
```
justinjpacheco: i don't believe we should be running this task before configure-networking because we won't have a service ip up at this point which means we'd have to go in through the transit interfaces

ansible/playbooks/roles/common/tasks/configure-apt.yml
```diff
@@ -0,0 +1,15 @@
...
+    src: apt/proxy.j2
+    dest: /etc/apt/apt.conf.d/proxy
+  when: local_proxy is defined and local_proxy['enabled']
+
+- name: configure apt translations
+  copy:
+    src: apt/translations
+    dest: /etc/apt/apt.conf.d/translations
+
+- name: add bootstrap apt repo
```
justinjpacheco: i think this task might have to be split off into its own thing and run after the `configure-web-server` tag

furlongm: This is why the `configure-web-server` tasks were moved.

What is the issue with the downloaded files being transferred over the transit interface rather than the service interface?

comay: I believe the intention is that nothing traverses over the transit link except for BGP traffic and the usual house keeping such as ARP between the node and the TOR.

Makefile
furlongm: Updated PR will cover this case by installing tinyproxy from deb

Makefile
furlongm: Updated PR still has this tag

ansible/playbooks/roles/common/tasks/configure-apt.yml
furlongm: Marking resolved as I think this is addressed in an updated PR

ansible/playbooks/roles/common/filter_plugins/util.py
```diff
@@ -82,6 +82,21 @@ def find_asset(a, *args, **kw):
     raise ValueError("could not find {}".format(asset_to_find))
 
 
+def find_assets(a, *args, **kw):
```
justinjpacheco: what does `find_assets` provide over `find_asset`?

ansible/playbooks/roles/common/defaults/main/base.yml
```diff
@@ -186,6 +197,26 @@ assets_files:
     checksum: sha256:2800962092ead67747ed2cd2087b0e254eb5e1a1b169cdc162c384598e4caed5
     filename: chef-server-core_12.17.33-1_amd64.deb
 
+  - name: aptly
```
justinjpacheco: why are we downloading these debs over the ones already provided in apt?

ansible/playbooks/roles/common/filter_plugins/util.py
furlongm: It returns a list of matching packages, due to tinyproxy being distributed as two debs rather than a single deb.

furlongm: I tried to modify `find_asset` to return a list if there are multiple matching packages, but the `etcd` glob mismatches two packages in that case.

ansible/playbooks/roles/common/defaults/main/base.yml
furlongm: In order to solve the chicken and egg problem you mentioned previously. With these downloaded, bird, tinyproxy and aptly are all that is needed to bootstrap without requiring `apt` access before installing bird and tinyproxy.

justinjpacheco:
I don't believe it will be in our best interest to accept this PR as-is for the following reasons:

1. the introduction of a local deb server
I don't think the added complexity and increased dependance on the bootstrap node justifies
the use of a local deb server. We only have 2 debs (chef-server, chef-client) that we install
locally via DPKG and this has never caused any problems. Also, as we move away from chef,
our dependance on the bootstrap server will go with it. I don't believe that now is a good
time to increase our reliance on the concept of a bootstrap node. Additionally,
the modifications needed to support this feature has caused the following to happen:

  a. Downloading external deb files
  In order to satisfy the chicken and eggs problems brought on by the introduction of
  a locally running deb server, we are now required to download external copies of tinyproxy
  and bird to solve these problems. Again, i don't think there is much value in this additional
  complexity

  b. Duplication of code with marginal differences
  In order to satisfy the requirements for installing tinyproxy manually, a `find_assets`
  Ansible method was added to search for similarly named asset files. Duplicating a method
  and altering its behavior in a minor way is not an optimal way to develop software. A better
  approach would have been to extend the existing method with an additional parameter to
  activate the desired behavior or update the existing data structure to support the idea
  of dependent files.

I'd just like to point out that I am not suggesting that the points raised in a) and b)
should be fixed, updated or solved. I am simply pointing out that these are problems that will
exist because of the existence the parent problem (the introduction of a local deb server).

2. Alteration of the networking recipe
This PR touches on aspects of the networking recipe which are currently being reworked to have nodes provisioned with service ip support from the beginning. This shift in networking configuration
responsibility will require these sections to be removed entirely as they are competing implementations trying to achieve the same goal.

I think this PR should try to achieve what we originally discussed which was
a simple update to allow us to configure the apt sources list before the chef-client
run. In general, I think our PRs should have a clear and defined purpose and
be re-evaluated when they become too large and unwieldy. I propose that we close
this PR and discuss what to do next.

PR #1600, opened 2019-09-24
Added some memory tuning to attempt to alleviate memory pressure issues

desrod:
Gently easing up the memory watermark to 50% of host RAM, and to start swapping when 75% of that 50% has been reached. This will attempt to address rabbitmq from spiraling off and consuming excessive amounts of memory when the queue gets a burst of messages. Also reformatted the ERB output for easier readability.


mihalis68:
stable-7.x deployments are being retired, this wont be deployed.

PR #1603, opened 2019-09-27
be more specific about libvirt builds

furlongm:
`ansible_virtualization_type` can be 'kvm' on physical hardware, whereas we only want this to match when running under QEMU and when the machine is VM (i.e. a kvm guest).

### DIFF
--- a/ansible/playbooks/roles/common/tasks/configure-networking.yml
+++ b/ansible/playbooks/roles/common/tasks/configure-networking.yml
@@ -135,7 +135,7 @@
     src: netplan/libvirt.yaml.j2
     dest: "/etc/netplan/libvirt.yaml"
   register: netplan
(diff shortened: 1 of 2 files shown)


ansible/playbooks/roles/common/tasks/configure-networking.yml
```diff
@@ -135,7 +135,7 @@
     src: netplan/libvirt.yaml.j2
     dest: "/etc/netplan/libvirt.yaml"
   register: netplan
-  when: ansible_virtualization_type == 'kvm' or ansible_system_vendor == 'QEMU'
+  when: ansible_virtualization_type == 'kvm' and ansible_virtualization_role == 'guest' and ansible_system_vendor == 'QEMU'
```
justinjpacheco: can you update this to use set_fact so that we don't repeat the condition logic?

PR #1610, opened 2019-10-18
disable placement api healthcheck

furlongm:


furlongm:
Also submitted https://storyboard.openstack.org/#!/story/2006741

PR #1614, opened 2019-11-04
Revert health checks

justinjpacheco:
these changes don't work on multi-head deployments. reverting until resolved


furlongm:
Can you give an example of how they do not work?

justinjpacheco:
during the second pass of the headnode installation the services failed to respond to any calls. eg openstack user list despite all services reporting up including mysql and rabbit. i reverted the above commits, re-ran automation and it worked. i suspect everything worked until haproxy was updated to use all 3 headnodes instead of the local headnode

furlongm:
Can you reproduce this issue? I tried to reproduce with the `3h3w.yml` topology but the build completed successfully 3 times in a row.

If you can reproduce, it would be helpful to provide the output of `openstack user list --debug` or `openstack token issue --debug`, if possible.

PR #1620, opened 2019-11-22
A quick 1-line patch to support disabling multipathing for NVME devices

desrod:
Small, 1-line fix to address re-enumerating NVME storage devices, which breaks Ceph OSDs when underlying block devices fail or are replaced.

### DIFF
--- a/ansible/playbooks/roles/common/defaults/main/chef.yml
+++ b/ansible/playbooks/roles/common/defaults/main/chef.yml
@@ -18,6 +18,7 @@ chef_environment:
         cmdline_linux:
           - net.ifnames=0
           - biosdevname=0
+          - nvme_core.multipath=0
       nova:
         cpu_config:
           cpu_mode: custom


ansible/playbooks/roles/common/defaults/main/chef.yml
```diff
@@ -18,6 +18,7 @@ chef_environment:
         cmdline_linux:
           - net.ifnames=0
           - biosdevname=0
+          - nvme_core.multipath=0
```
comay: I don't think this is going to actually have any effect on real clusters, only virtual ones. Instead, I think a section of the form

```
      grub:
        cmdline_linux:
          - nvme_core.multipath=0
```

needs to be added to the appropriate cluster's `chef_environment.yml` file. I would double check with Justin though.

desrod: I'm reverting this and replacing it with another in system.rb, PR inflight shortly.

PR #1622, opened 2019-12-09
use correct jinja2 boolean logic

furlongm:


ansible/playbooks/roles/common/templates/ssh/sshd_config.j2
```diff
@@ -19,7 +19,7 @@ Protocol 2
 Port 22
 {% if ansible_virtualization_type == 'virtualbox' %}
 ListenAddress {{ virtualbox['nat_ip'] }}
-{% elif libvirt_kvm_guest == True %}
+{% elif libvirt_kvm_guest %}
```
justinjpacheco: this is not correct, this fails in metal environments, it treats the existence as true, it needs to be tested to see that it is false

furlongm: Odd. `== True` does not work in a libvirt env. Maybe it should be `== true`?

furlongm: From the jina2 docs:

```
The if statement in Jinja is comparable with the Python if statement. In the simplest form, you can use it to test if a variable is defined, not empty and not false
```

justinjpacheco: i printed the value, it showed "False" but the code was still executing for that condition (so ansible was failing because it was trying to look up eth0 which is what it does when the condition is true). so despite the condition evaluating to "False" it was still acting as true

furlongm: It looks like in a metal environment, this is evaluating to "False\n" as a result of 

https://github.com/ansible/ansible/issues/18142

`false` + "\n" = "False\n" which then evaluates to `true` in jinja2

The workaround described in that bug works as expected.

furlongm:
Reopening pending testing, but it looks like working around the ansible set_fact bug means this should now work correctly.

PR #1625, opened 2019-12-13
generate libvirt host uuid

justinjpacheco:
- this update will produce a unique, deterministic uuid for each host
    based on the node's fqdn and use that for libvirts system uuid vs.
    relying on the one provided by the system hardware

### DIFF
--- a/chef/cookbooks/bcpc/templates/default/libvirt/libvirtd.conf.erb
+++ b/chef/cookbooks/bcpc/templates/default/libvirt/libvirtd.conf.erb
@@ -55,7 +55,7 @@ unix_sock_rw_perms = "0770"
 # In case 'dmidecode' does not provide a valid UUID and none is
 # provided here, a temporary UUID will be generated.
 #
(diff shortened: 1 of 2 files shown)


chef/cookbooks/bcpc/templates/default/libvirt/libvirtd.conf.erb
```diff
@@ -46,3 +46,10 @@ unix_sock_group = "libvirt"
 # If not using PolicyKit and setting group ownership for access
 # control then you may want to relax this to:
 unix_sock_rw_perms = "0770"
+
+# UUID of the host:
+# Provide the UUID of the host here in case the command
+# 'dmidecode -s system-uuid' does not provide a valid uuid. In case
+# 'dmidecode' does not provide a valid UUID and none is provided here, a
+# temporary UUID will be generated.
```
mihalis68: is it really a temporary UUID? Seems to me it's a permanent uuid using a different generator?

PR #1629, opened 2020-01-10
Journal capacity tuning

desrod:
Add some tuning to systemd-journald to allow for greater local journal capacity to help with any triage, debugging or diagnosis that needs to be made prior to the most-current, truncated boot journals.


justinjpacheco:
could i see some testing output please? something that involves a fresh build showing the recipe run and the changes taking effect.

chef/cookbooks/bcpc/files/default/journal/journal.conf
```diff
@@ -0,0 +1,43 @@
+#  This file is part of systemd.
```
justinjpacheco: how is this file being deployed? Also, could we have the options that aren't used removed?

desrod: Both of your requests have been implemented in the latest commit to this branch. Let me know if that addresses your concerns.

PR #1631, opened 2020-01-23
Updating default grub options to include deterministic nvme device na…

desrod:
This addresses the issue with controller vs.  disk/device naming issues, providing consistent, deterministic device naming conventions.

See the following for context:

https://github.com/linux-nvme/nvme-cli/issues/455
https://bugs.launchpad.net/ubuntu-power-systems/+bug/1778844
https://patchwork.kernel.org/patch/10039191/


chef/cookbooks/bcpc/recipes/system.rb
```diff
@@ -56,8 +56,11 @@
   source 'grub/default.erb'
 
   cmdline = []
-  io_scheduler = node['bcpc']['hardware']['io_scheduler']
+  # This addresses the issue with controller vs.  disk/device naming issues,
```
comay: Could you please remove the extra space between "vs." an "disk/device"?

PR #1632, opened 2020-01-30
replaced missing disable_bgp function call with remove_cloud_ip()

justinjpacheco:


chef/cookbooks/bcpc/templates/default/consul/dns-watch.erb
```diff
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-#
 # Copyright 2018, Bloomberg Finance L.P.
```
comay: Year should be updated.

PR #1635, opened 2020-02-01
Move automation for `chrony` from Chef to Ansible

comay:
The proposed change moves the automation for provisioning `chrony` from Chef to Ansible. By doing this, `chrony` can start early enough on all of the nodes in the cluster so that the clocks on the instances are synced up with a time source before most other services start. This benefits services that are time critical (such as `etcd`) as well keeping clocks in sync during the long provisioning time of head nodes.


justinjpacheco:
nice work

PR #1636, opened 2020-02-04
separated osd creation and activation into two separate operations

justinjpacheco:


chef/cookbooks/bcpc/recipes/ceph-osd.rb
```diff
@@ -64,6 +64,14 @@
         ceph-volume lvm zap --destroy /dev/#{osd}
         ceph-volume lvm create --bluestore --data /dev/#{osd}
```
tstachecki: From Ceph docs:
```
[ceph-volume lvm] create: Wraps the two-step process to provision a new osd (calling prepare first and then activate) into a single one. The reason to prefer prepare and then activate is to gradually introduce new OSDs into a cluster, and avoiding large amounts of data being rebalanced.
```

Does it make sense to `s/create/prepare` here and follow up with an `ceph-volume lvm activate --all` as an additional preventative measure?

justinjpacheco: i assume this would mean that on large clusters (100+ hosts), the act of enabling each osd would involve stepping onto each host and running the activation step. this would achieve the same goal, but require actions taken on multiple hosts vs. a single host

ansible/playbooks/roles/headnode/tasks/enable-ceph-osds.yml
```diff
@@ -0,0 +1,11 @@
+- name: enable ceph osds
+  changed_when: false
+  shell: |
+    set -o pipefail
+    set -xe
+
+    osd_tree=$(ceph osd tree --format json)
+    osd_ids=$(echo ${osd_tree} | jq '.nodes[] | select(.type == "osd") .id')
+    echo "${osd_ids}" | xargs -I% -n1 ceph osd in osd.%
```
comay: The usage statement for `ceph osd in` seems to indicate that `ceph osd in any` or `ceph osd in all` rather than iterating over the OSDs in the tree.

```Invalid command: saw 0 of ids(<string>) [<string>...], expected at least 1
osd in <ids> [<ids>...] :  set osd(s) <id> [<id>...] in, can use <any|all> to automatically set all previously out osds in
Error EINVAL: invalid command

File: chef/cookbooks/bcpc/recipes/ceph-osd.rb [hunk 3]
```diff
@@ -64,6 +64,14 @@
...
         sleep 5
+
+        # get osd id from ceph-volume lvm list
+        osd_info=$(ceph-volume lvm list /dev/#{osd} --format json)
+        osd_id=$(echo ${osd_info} | jq -r '.[][] .tags ."ceph.osd_id"')
+
+        # mark osd as out
+        # this osd will be added back in a later task
+        ceph osd out osd.${osd_id}
```
comay: Nit but you don't need the `osd.` prefix there.

PR #1638, opened 2020-02-06
add suffix to virtualbox VMs and network names

peter-at:
Allow multiple users to create multiple chef-bcpc builds by adding a suffix to virtualbox object names during VM creation. This feature is control by `ENABLE_VBOX_SUFFIX` environment variable. Set to enable, unset to run as  current. As this affects how VM are 'named' it is best place this in profile.

Testing
`make create` - resulted in same virtualbox names as before this PR
`ENABLE_VBOX_SUFFIX='true' make create` - created suffix added virtualbox object names (VMs and intnets - eg: tor2_ssV1i, management1_ssV1i)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 *.swo
 assets/
 ansible/inventory.yml
+ansible/group_vars/all/overrides.yml
 chef/roles/node.json
 chef/roles/bootstrap.json
 chef/roles/headnode.json
@@ -14,3 +15,5 @@ chef/roles/storagenode.json
 chef/environments/
 chef/databags/
 virtual/files/ssh/id_*
+hardware.overrides.yml
+topology.overrides.yml
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 *.swo
 assets/
 ansible/inventory.yml
-ansible/group_vars/all/overrides.yml
 chef/roles/node.json
 chef/roles/bootstrap.json
 chef/roles/headnode.json
@@ -15,5 +14,3 @@ chef/roles/storagenode.json
 chef/environments/
 chef/databags/
 virtual/files/ssh/id_*
-hardware.overrides.yml
-topology.overrides.yml
--- a/virtual/Vagrantfile
+++ b/virtual/Vagrantfile
@@ -100,8 +100,8 @@ Vagrant.configure('2') do |config|
 
       transit.each do |t_iface|
         subconfig.vm.network('private_network', Util.privnet_args(
-		t_iface['neighbor']['name'],
-		macaddr: t_iface['mac'].delete(':') ))
+            t_iface['neighbor']['name'],
+            macaddr: t_iface['mac'].delete(':')))
       end
 
       subconfig.vm.provider 'virtualbox' do |vb|
--- a/virtual/Vagrantfile
+++ b/virtual/Vagrantfile
@@ -27,6 +27,15 @@ def mount_apt_cache(config)
   config.vm.synced_folder cache_dir, apt_cache_dir, create: true, owner: '_apt'
 end
 
+# get virtualbox private network suffix to use
+# * set to environment variable 'VBOX_NET_SUFFIX'
+# * default to login name (eg: _login)
+# * eg: vbox_net_suffix = ENV['VBOX_NET_SUFFIX'] || '_'+ENV['USER']
+vbox_net_suffix = '_' + ENV['USER']
+if ENV.has_key? 'VBOX_NET_SUFFIX'
+  vbox_net_suffix = ENV['VBOX_NET_SUFFIX']
+end
+
 # load the vm topology profile
 topology = './topology/topology.yml'
 topology_overrides = './topology/topology.overrides.yml'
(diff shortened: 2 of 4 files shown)


virtual/Vagrantfile
```diff
@@ -27,6 +27,15 @@ def mount_apt_cache(config)
   config.vm.synced_folder cache_dir, apt_cache_dir, create: true, owner: '_apt'
 end
 
+# get virtualbox private network suffix to use
```
justinjpacheco: i think this should be turned into a function that is stored in a shared lib used by both the network Vangrantfile and vm VangrantFIle. it should take a single argument (the name of whatever your trying to add a suffix to) and depending on the existence of the environment variable VBOX_SUFFIX, it will append the desired suffix to the argument or just return the parameter passed to the function. I suggest we put this function `vbox_name_builder(param`)` (for example)  in `virtual/lib/utils.rb` and have that imported where needed. The overall goal being to reduce duplication and ensure that nothing changes that if this environment variable doesn't exist

comay:
I want to throw out a suggestion that rather than supplying a suffix via an environment variable, instead the environment variable be considered a boolean - if set, then a suffix be generated by hashing the base directory of the virtual environment. I don't know what limits there are on the length of VirtualBox network names but I'm hoping there's some reasonable hash that would work here.

mihalis68:
I like @comay 's idea a lot. This would enable me to do multiple full builds myself, as well as enable more than one user to have a build per build host.

virtual/Vagrantfile
```diff
@@ -20,11 +20,12 @@ base_box_version = '201812.27.0'
...
 require 'yaml'
 
-def mount_apt_cache(config)
-  user_data_path = Vagrant.user_data_path.to_s
-  cache_dir = File.join(user_data_path, 'cache', 'apt', config.vm.box)
-  apt_cache_dir = '/var/cache/apt/archives'
-  config.vm.synced_folder cache_dir, apt_cache_dir, create: true, owner: '_apt'
+# load util module
+util_path = './lib/util.rb'
+if File.file?(File.expand_path(util_path))
```
justinjpacheco: won't think already raise an exception if the library isn't found?

.gitignore
```diff
@@ -6,7 +6,6 @@
 *.swo
 assets/
 ansible/inventory.yml
-ansible/group_vars/all/overrides.yml
```
justinjpacheco: these were added specifically to allow for untracked overrides to the default config

virtual/lib/util.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+    suffix
+  end
+
+  def self.privnet_args(name, nic_type: '82543GC', macaddr: nil)
+    vbox_args = {}
+    vbox_args[:virtualbox__intnet] = name + VBOX_NET_SUFFIX
+    vbox_args[:nic_type] = nic_type
+    vbox_args[:auto_config] = false
+    unless macaddr.nil?
```
justinjpacheco: vbox already checks that the mac address conforms to what it considers a valid mac addresss. it will throw an exception

.gitignore
peter-at: this caused me few hours of debugging and multiple builds, because git status hide them and made me think there was no changes
to me that worth it to remove them from .gitignore

virtual/lib/util.rb
```diff
@@ -0,0 +1,37 @@
+module Util
+  # initialize constants
+  # ## virtualbox private network suffix to use
+  # * set to environment variable 'VBOX_NET_SUFFIX'
+  # * default to a static hash
+  # * eg: VBOX_NET_SUFFIX = ENV['VBOX_NET_SUFFIX'] || '_'+hash(__dir__)
+  VBOX_NET_SUFFIX = begin
```
justinjpacheco: this should not build any kind of suffix if one is not desired which is why i suggested that this be a method with the sole purpose of either returning the name or the name+suffix.

virtual/Vagrantfile
```diff
@@ -98,18 +94,16 @@ Vagrant.configure('2') do |config|
...
       transit.each do |t_iface|
-        args = {}.tap do |a|
-          a[:virtualbox__intnet] = t_iface['neighbor']['name']
-          a[:nic_type] = '82543GC'
-          a[:auto_config] = false
-          a[:mac] = t_iface['mac'].delete(':')
-        end
-
-        subconfig.vm.network('private_network', args)
+        subconfig.vm.network('private_network',
```
justinjpacheco: is this really serving a purpose if it's just taking all the same arguments?

virtual/lib/util.rb
```diff
@@ -0,0 +1,32 @@
+module Util
+  # initialize constants
+  # ## virtualbox private network suffix to use
+  # * control by environment variable 'ENABLE_VBOX_SUFFIX'
+  # * set and non-empty -> set to a static hash base off full path of this file
+  # * not set or empty -> set to empty (as if there's no suffix)
+  VBOX_SUFFIX = begin
```
justinjpacheco:
```suggestion
def self.vbox_name(name:) 
  # return name if suffix not enabled
  unless ENV.key?('ENABLE_VBOX_SUFFIX')
    return name
  end

  # return name + hashed __dir__
  require 'digest/sha1'
  hash = Digest::SHA1.hexdigest(__dir__)[0,7]
  return name + '_' + hash
end
```

PR #1639, opened 2020-02-11
Ignore download if exists key defined for an asset

srsakhamuri:
When dynamically generated asset we don't have a download path. Make sure we don't attempt to download those assets

Specify the file that need to be skipped download
```
  - name: archive
    filename: archive.tar.gz
    exists: true
```
**Testing performed**
Tested on vanilla builds with this, noticed that the download assets code skips this file


justinjpacheco:
i think it would be better to have a key that makes it obvious what operation is being skipped: eg skip-download: true etc

PR #1643, opened 2020-03-12
Add additional apache2 configuration

tj90241:
These commits:
- Explicitly configure `mpm_event` as the MPM module, rather than relying on the packaging scripts to do the same.
- Allow for `mpm_event` configuration to be tuned via the Chef environment.
- Disable `mod_status` by default in order to strengthen endpoint security.
- Disable HTTP persistent connections by default to improve OpenStack API reliability during log rotation/graceful restarts of `apache2`.


peter-at:
lgtm
nice! like the detail comment in the recipe.

PR #1644, opened 2020-03-20
log bird log to it's own log file

bzhang257:


justinjpacheco:
could we see some output from a fresh build using this patch please?

tj90241:
Instead of touching the log file, I would suggest this:

```
$ cat /etc/systemd/system/bird.service.d/override.conf 
[Service]
User=bird
Group=bird
$ cat /etc/systemd/system/bird6.service.d/override.conf 
[Service]
User=bird
Group=bird
$ systemctl daemon-reload
```

This will cause `bird` to parse the config and open log files as itself, rather than implementing that from the orchestration.

tj90241:
By the way, does log rotation work?
```
$ cat /etc/logrotate.d/bird
/var/log/bird/bird.log {
  daily
  compress
  missingok
  notifempty
}
```

Not sure if this would flap BGP state when `logrotate` runs or not - wouldn't think so...

bzhang257:
> By the way, does log rotation work?
> 
> ```
> $ cat /etc/logrotate.d/bird
> /var/log/bird/bird.log {
>   daily
>   compress
>   missingok
>   notifempty
> }
> ```
> 
> Not sure if this would flap BGP state when `logrotate` runs or not - wouldn't think so...

I haven't tested log rotation yet.

ansible/playbooks/roles/common/templates/systemd/bird/override.conf
```diff
@@ -0,0 +1,6 @@
+[Service]
+ExecStart=
+ExecStart=/usr/sbin/bird -f $BIRD_ARGS $BIRD4_ARGS
+User=bird
+Group=bird
+
```
comay: Can you please remove this extraneous blank line?

ansible/playbooks/roles/common/tasks/configure-bgp.yml
```diff
@@ -2,18 +2,42 @@
...
+    path: /etc/systemd/system/bird.service.d
+    state: directory
+    group: root
+    owner: root
+
+- name: service bird conf override
+  template:
+    src: systemd/bird/override.conf
+    dest: /etc/systemd/system/bird.service.d/override.conf
+  register: bird_conf_override
```
comay: Nit - suggest `bird_service_override` here.

ansible/playbooks/roles/common/tasks/configure-bgp.yml
```diff
@@ -2,18 +2,42 @@
...
 
-- name: restart bird
+- name: reload bird
   service:
     name: bird
-    state: restarted
-    enabled: yes
-  when: bird.changed
+    state: reloaded
+  when: bird_conf.changed or bird_conf_override.changed
```
comay: Likewise.

ansible/playbooks/roles/common/tasks/configure-bgp.yml
```diff
@@ -2,18 +2,42 @@
   apt:
     name: bird
 
+- name: create /var/log/bird directory with correct permission
```
tj90241: @justinjpacheco had a good suggestion on this: this: it can be moved to the systemd override.  `ExecStartPre=` should give you an opportunity to `mkdir` and `chown`.

ansible/playbooks/roles/common/templates/systemd/bird/override.conf
bzhang257: Empty line at end of file makes every line look same to parser, without it last line ends with EOF instead of \n.

ansible/playbooks/roles/common/tasks/configure-bgp.yml
bzhang257: mkdir would fail if the directory is already there, I'm not sure how systemd handles the failure.

ansible/playbooks/roles/common/templates/systemd/bird/override.conf
comay: Really? The other `custom.conf` files we have been creating don't have a blank line like in line 6.

bzhang257: If we know systemd parser works I can remove the empty line, in general it's safer to have a empty line in the end.

ansible/playbooks/roles/common/tasks/configure-bgp.yml
```diff
@@ -2,18 +2,35 @@
...
+    dest: /etc/systemd/system/bird.service.d/override.conf
+
 - name: configure bird
   template:
     src: bird/bird.conf.j2
     dest: /etc/bird/bird.conf
-  register: bird
+  register: bird_conf
+
+- name: start bird
```
justinjpacheco: is bird not started when it's insalled?

comay: Also, doesn't the `daemon_reload` belong under the `reload bird` action?

Since we want to avoid restarting bird and *if* reloading is sufficient if `bird.conf` changes, then it seems either adding the override or configuring bird.conf should cause a daemon reload to take place.

bzhang257: > Also, doesn't the `daemon_reload` belong under the `reload bird` action?
> 
> Since we want to avoid restarting bird and _if_ reloading is sufficient if `bird.conf` changes, then it seems either adding the override or configuring bird.conf should cause a daemon reload to take place.

Daemon-reload is needed when the override changes, but I didn't find an Ansible module that only does daemon-reload, so I put it in as part of the start bird section. The alternative is using Ansible command module and call /bin/systemctl daemon-relaod explicitly and set override file change as condition. But it's more tedious that way. Having it in start bird section is more concise.

ansible/playbooks/roles/common/tasks/configure-bgp.yml
```diff
@@ -2,18 +2,35 @@
   apt:
     name: bird
 
+- name: create /etc/systemd/system/bird.service.d directory with correct permission
```
justinjpacheco:
```suggestion
- name: create /etc/systemd/system/bird.service.d
```

ansible/playbooks/roles/common/tasks/configure-bgp.yml
```diff
@@ -2,18 +2,35 @@
...
+  file:
+    path: /etc/systemd/system/bird.service.d
+    state: directory
+    group: root
+    owner: root
+
+- name: service bird conf override
```
justinjpacheco:
```suggestion
- name: install bird service override.conf
```

ansible/playbooks/roles/common/tasks/configure-bgp.yml
```diff
@@ -2,18 +2,41 @@
...
+    owner: root
+
+- name: install bird service override.conf
+  template:
+    src: systemd/bird/override.conf
+    dest: /etc/systemd/system/bird.service.d/override.conf
+  register: bird_service_override
+
+- name: reload bird serivce override.conf
+  sudo: yes
```
justinjpacheco: i don't think you need this, this task is already running as root

ansible/playbooks/roles/common/tasks/configure-bgp.yml
```diff
@@ -2,18 +2,41 @@
...
+  command: /bin/systemctl daemon-reload
```
comay: Shouldn't need to specify the full path name here (none of our other playbooks seem to do this either).

ansible/playbooks/roles/common/tasks/configure-bgp.yml
```diff
@@ -2,18 +2,41 @@
...
+  when: bird_service_override.changed
+
+- name: restart bird
+  service:
+    name: bird
+    state: restarted
+  when: bird_service_override.changed
```
comay: I guess I don't understand why this is here rather than another condition for `reload bird`?

ansible/playbooks/roles/common/templates/systemd/bird/override.conf
```diff
@@ -0,0 +1,7 @@
+[Service]
+ExecStartPre=/bin/mkdir -p /var/log/bird
+ExecStartPre=/bin/chown bird: /var/log/bird
+ExecStart=
+ExecStart=/usr/sbin/bird -f $BIRD_ARGS $BIRD4_ARGS
+User=bird
+Group=bird
```
comay: Still curious about this - I see an indication that there isn't a trailing newline. That should be there - earlier I was complaining about an *extra* blank line after this. I'm curious if it has something to do with the editor you're using to edit the files?

ansible/playbooks/roles/common/tasks/configure-bgp.yml
```diff
@@ -2,18 +2,41 @@
...
+    group: root
+    owner: root
+
+- name: install bird service override.conf
+  template:
+    src: systemd/bird/override.conf
+    dest: /etc/systemd/system/bird.service.d/override.conf
+  register: bird_service_override
+
+- name: reload bird serivce override.conf
```
justinjpacheco:
```suggestion
- name: systemctl daemon-reload
```

ansible/playbooks/roles/common/tasks/configure-bgp.yml
justinjpacheco: - name: systemd to reread configs
  systemd:
    daemon_reload: yes

ansible/playbooks/roles/common/tasks/configure-bgp.yml
bzhang257: In the service override we added two new ExecStartPre and modified bird launch arguments, so bird needs to be restarted. Reload bird execute "birdc configure".

ansible/playbooks/roles/common/templates/systemd/bird/override.conf
bzhang257: I guess when "\n" is the last character before EOF, it always shows the new line.

ansible/playbooks/roles/common/templates/systemd/bird/override.conf
bzhang257: Looks like it was the editor. When vi has "\n" at the end of the file, it adds 2 "\n". emacs only add 1.

bzhang257:
Review please

ansible/playbooks/roles/common/tasks/configure-bgp.yml
comay: Ah, thanks for reminding me on that.

comay:
I think this looks good but the `logrotate` changes are still coming, yes?

bzhang257:
> I think this looks good but the `logrotate` changes are still coming, yes?

Yes I'll do a separate PR

comay:
I think it makes sense for this to be part of this PR.

ansible/playbooks/roles/common/tasks/configure-bgp.yml
bzhang257: This is updated to use Ansible systemd module.

bzhang257:
> I think it makes sense for this to be part of this PR.

Added Logrotate change to thes PR

justinjpacheco:
so it just occurred to me that if we make these systemctl changes then we have to restart the bird daemon which is much more disruptive than what we were originally planning for.

justinjpacheco:
i don't think this disruption is worth having bird run the service over root

comay:
True but it's a one-time thing and presumably the `configure-bgp` task can be staged when appropriate? Or am I misunderstanding?

peter-at:
since this is to support one time thing to debug bird can we make the playbook safer?
such that multiple hosts won't be run by accident.
like running via extra-var, like in this [so post](https://stackoverflow.com/a/18195217)

comay:
The intent though is to do this across all of the nodes in the cluster permanently, rather than as a one-time debugging aid.

bzhang257:
> Bin, this looks ready to approve but can you please verify that the log file created during a log rotation has the correct owner/group and is being written to after it's been rotated?

Just tested, log rotate works fine. It creates new bird.log with correct owner/group.

PR #1645, opened 2020-03-24
Pushing change to assert deterministic NVME device naming.

desrod:
This addresses the changes in JIRA RDTIBCC-2037

**Describe your changes**
Asserting deterministic NVME device naming 

**Testing performed**
With this  configuration option enabled, the devices will enumerate with the same device and  node identifier at each boot, cold or warm. Without this, devices are enumerated as they appear on the bus, which is not always in  numerical or bus ordered.


chef/cookbooks/bcpc/recipes/system.rb
```diff
@@ -56,6 +56,10 @@
   source 'grub/default.erb'
 
   cmdline = []
+  # This addresses the issue with controller vs. disk/device naming issues,
```
tj90241: I don't think you want this here; the value is already being inserted a few lines below:
```
  unless node['bcpc']['grub']['cmdline_linux'].empty?
    cmdline += node['bcpc']['grub']['cmdline_linux']
  end
```

You won't see this change currently take effect on BVE, as the BVE prop has an override which redefines the `cmdline_linux` list.

comay: I know David D, Justin, and I went through this a few months back and I thought the consensus was to make it generally set (although unless you're doing the device passthrough, it won't be seen or necessary in the virtual case) in a similar way to the `elevator=deadline` setting. In that way, the properties defined in branches like `bve` supplement rather than replace.

tj90241: That's agreeable - though in that case, the *other* reference to the setting should be dropped from `cmdlinx_linux`.  Otherwise, it's liable to appear twice if the prop does not provide an override for `cmdline_linux`.

tj90241:
Unrelated build issues are being tackled in #1647.

chef/cookbooks/bcpc/recipes/system.rb
tj90241: Made a slight change that drags the setting out of the `grub` list and makes `nvme_multipath` a first-class Chef hardware attribute, much like `io_scheduler` as @comay mentioned above.

The default is _unset_ as the introduction of NVMe multipath support is what causes the inconsistent labeling.

tj90241:
Folded into https://github.com/bloomberg/chef-bcpc/pull/1695

PR #1646, opened 2020-03-24
Add support for specifying raw disk devices for pass-through into nodes

comay:

This set of changes allows one to specify in the `hardware` YAML file a list of exclusive disk device names to export into a node. This can be used, for example, to deploy Ceph OSDs directly to devices from the host system rather than creating virtual disk devices from the host file system.

### DIFF
--- a/virtual/Vagrantfile
+++ b/virtual/Vagrantfile
@@ -115,34 +115,50 @@ Vagrant.configure('2') do |config|
 
         # add additional hard drives
         if hw_profile.key?('ext_disks')
-
           size_gb = hw_profile['ext_disks']['size_gb']
-          count = hw_profile['ext_disks']['count']
-          drive_letters = ('a'..'z').to_a[1..count]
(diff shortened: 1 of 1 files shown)


virtual/Vagrantfile
```diff
@@ -115,34 +115,50 @@ Vagrant.configure('2') do |config|
```
justinjpacheco: wouldn't this be better served and understood by having two distinctive code blocks (one for ext_disks method and another for the raw_disks method)?

i don't think we should be merging the two different disk styles into the same code context

comay: Understood although I was trying to avoid code duplication. I can certainly though create larger code blocks under `hw_profile.key?('ext_disks')` and `hw_profile.key?('raw_disks')`.

justinjpacheco: yah, exactly:

if ext_disks:

if raw_disks:

i think it would go a long way towards readability

PR #1650, opened 2020-04-13
Add a new task with a tag to configure chrony

srsakhamuri:
Add a new task with a tag to configure chrony
  - remove un-necessary comment from chrony

Signed-off-by: Srinivas Sakhamuri <ssakhamuri@bloomberg.net>

Add a new sub task to setup chrony

Tested on plain build and make sure it is working and chrony conf is updated as expected

### DIFF
--- a/ansible/playbooks/roles/common/tasks/main.yml
+++ b/ansible/playbooks/roles/common/tasks/main.yml
@@ -14,3 +14,7 @@
 - import_tasks: configure-networking.yml
   become: true
   tags: [never,configure-networking]
+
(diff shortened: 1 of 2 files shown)


comay:
When this was introduced, it was purposely added to the `configure-networking` task so that it was done before `configure-sshd`. Is there a reason to break it out like this or is it so it can be executed by itself?

srsakhamuri:
@comay yeah it is to isolate the changes chrony alone. This is to help day 2 operations.

PR #1653, opened 2020-04-20
update bgp protocol name in bird.conf 

bzhang257:
change bgp protocol name in bird.conf to 
[tor name]:[transit interface name]

### DIFF
--- a/ansible/playbooks/roles/common/templates/bird/bird.conf.j2
+++ b/ansible/playbooks/roles/common/templates/bird/bird.conf.j2
@@ -45,7 +45,7 @@ protocol device {
 }
 
 {% for transit in transit_interfaces %}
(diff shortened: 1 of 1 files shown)


bzhang257:
Below is updated bgp protocol section in bird config file.

```
protocol bgp 'management1:eth1' {
  multihop;
  next hop self;
  graceful restart;
  local 10.121.84.3 as 4200858801;
  neighbor 10.121.84.1 as 4200858701;
  import all;
  export filter to_tor;
}

protocol bgp 'management4:eth2' {
  multihop;
  next hop self;
  graceful restart;
  local 10.121.88.3 as 4200858801;
  neighbor 10.121.88.1 as 4200858702;
  import all;
  export filter to_tor;
}
```

PR #1659, opened 2020-04-29
add support to source rabbitmq from distribution

justinjpacheco:


chef/cookbooks/bcpc/attributes/rabbitmq.rb
```diff
@@ -0,0 +1,21 @@
+###############################################################################
+# rabbitmq
+###############################################################################
+
+# source rabbitmq from 3rd party apt repository
+default['bcpc']['rabbitmq']['source']['repo']['enabled'] = false
+default['bcpc']['rabbitmq']['source']['repo']['url'] = 'http://dl.bintray.com/rabbitmq/debian'
+
+# source rabbitmq from a different distribution
```
desrod, line 9: Not picking nits, but it's from a different repository pocket (eg: backports vs. security vs. updates pocket), not a different distribution. 

Something like: "source rabbitmq from Ubuntu backports pocket" would be more representative of where these packages are coming from. The 3rd-party bintray repo url is the one coming from a different distribution altogether. 

This link gives a good explanation of those terms and their usage: 

https://docs.ubuntu.com/landscape/en/repositories

chef/cookbooks/bcpc/templates/default/rabbitmq/apt-preferences.erb
```diff
@@ -0,0 +1,7 @@
+Package: rabbitmq-server
+Pin: release a=<%= @release %>
+Pin-Priority: 500
```
comay, line 3: Isn't the default value here already 500? Alternatively, why is this file actually necessary?

justinjpacheco: the default priority reported from `apt-cache policy rabbitmq-server` reports packages from bionic-backports as 100, which gives them less priority over the non-backport packages. this change puts them on equal footing but preferred because of their higher package version.

PR #1662, opened 2020-05-04
Update BGP daemon from bird (1.x) to bird2.

tj90241:
Internal testing has demonstrated that bird2 is better
suited for BGP peering with fast timers.  This commit
introduces a change to the Ansible workflows that will
provision/upgrade to bird2 from bionic-backports.

### DIFF
--- a/ansible/playbooks/roles/common/defaults/main/base.yml
+++ b/ansible/playbooks/roles/common/defaults/main/base.yml
@@ -220,7 +220,6 @@ networking:
   mtu: 1500
 
 use_primary_transit_interface: true
-configure_bgp: true
 configure_service_ip: true
 configure_network_interfaces: true
 
--- a/ansible/playbooks/roles/common/defaults/main/base.yml
+++ b/ansible/playbooks/roles/common/defaults/main/base.yml
@@ -244,3 +244,10 @@ ubuntu_apt:
       url: http://security.ubuntu.com/ubuntu
     - dist: bionic-updates
       url: http://archive.ubuntu.com/ubuntu
+
+###############################################################################
+# bird
+###############################################################################
+bird:
+  distribution:
+    name: bionic-backports
--- a/ansible/playbooks/roles/common/defaults/main/base.yml
+++ b/ansible/playbooks/roles/common/defaults/main/base.yml
@@ -219,7 +219,6 @@ host_aggregates: []
 networking:
   mtu: 1500
 
-use_primary_transit_interface: true
 configure_service_ip: true
 configure_network_interfaces: true
 
--- a/ansible/playbooks/roles/common/tasks/configure-networking.yml
+++ b/ansible/playbooks/roles/common/tasks/configure-networking.yml
@@ -33,6 +33,9 @@
 - import_tasks: configure-ubuntu-apt-sources.yml
   tags: [configure-ubuntu-apt-sources]
 
+- import_tasks: upgrade-bird.yml
+  when: configure_bgp | default(true)
+
 - import_tasks: configure-bgp.yml
   when: configure_bgp | default(true)
 
--- a/ansible/playbooks/roles/common/tasks/configure-networking.yml
+++ b/ansible/playbooks/roles/common/tasks/configure-networking.yml
@@ -1,16 +1,6 @@
-- name: gather host facts
+- name: gather network and hardware/vm facts
   setup:
-    gather_subset: all
-
-- name: define ansible_host
-  set_fact:
-    ansible_host: "{{ interfaces['transit'] | primary_ip(hostvars[inventory_hostname]) }}"
-  when: use_primary_transit_interface
-
-- name: define transit_interfaces
-  set_fact:
-    transit_interfaces: >
-      {{ interfaces['transit'] | transit_interfaces(ansible_facts) }}
+    gather_subset: "!all,!min,hardware,network,virtual"
 
 - name: define libvirt_kvm_guest
   set_fact:
@@ -18,9 +8,6 @@
       {{ ansible_virtualization_type == 'kvm'
           and ansible_virtualization_role == 'guest'
           and ansible_system_vendor == 'QEMU' }}
-
-- import_tasks: configure-etc-hosts.yml
-  tags: [configure-etc-hosts]
 
 - import_tasks: configure-root-user-ssh.yml
   tags: [configure-root-user-ssh]
@@ -33,17 +20,14 @@
 - import_tasks: configure-ubuntu-apt-sources.yml
   tags: [configure-ubuntu-apt-sources]
 
-- import_tasks: upgrade-bird.yml
-  when: configure_bgp | default(true)
-
-- import_tasks: configure-bgp.yml
-  when: configure_bgp | default(true)
+- import_tasks: configure-network-interfaces.yml
+  when: configure_network_interfaces | default(true)
 
 - import_tasks: configure-service-ip.yml
   when: configure_service_ip | default(true)
 
-- import_tasks: configure-network-interfaces.yml
-  when: configure_network_interfaces | default(true)
+- import_tasks: configure-bgp.yml
+  tags: [configure-bgp]
 
 - import_tasks: configure-chrony.yml
   tags: [configure-chrony]
--- a/ansible/playbooks/roles/common/tasks/define-transits.yml
+++ b/ansible/playbooks/roles/common/tasks/define-transits.yml
@@ -1,0 +1,8 @@
+- name: gather host facts for interfaces
+  setup:
+    gather_subset: "!all,!min,network"
+
+- name: define transit_interfaces
+  set_fact:
+    transit_interfaces: >
+      {{ interfaces['transit'] | transit_interfaces(ansible_facts) }}
--- a/ansible/playbooks/roles/common/tasks/upgrade-bird.yml
+++ b/ansible/playbooks/roles/common/tasks/upgrade-bird.yml
@@ -1,0 +1,62 @@
+---
+- name: query_for_bird_installation
+  command: dpkg-query -s bird
+  register: dpkg_query_bird
+  failed_when: dpkg_query_bird.rc != 0
+  ignore_errors: true
+  changed_when: false
+
+# Install new bird2 package
+- name: Upgrade bird to bird2
+  block:
+    - name: Check if existing /etc/bird/bird.conf exists
+      stat:
+        path: /etc/bird/bird.conf
+      register: existing_bird_conf
+
+    - name: Backup existing bird config file
+      copy:
+        remote_src: yes
+        src: /etc/bird/bird.conf
+        dest: /tmp/bird.conf.old
+      when: existing_bird_conf.stat.exists
+
+    - name: Generate apt preferences for bird2 and dependencies
+      template:
+        src: bird/apt-preferences.j2
+        dest: /etc/apt/preferences.d/98bird
+
+    - name: Drop new bird2 config file
+      template:
+        src: bird/bird.conf.j2
+        dest: /tmp/bird.conf.new
+        owner: root
+        group: root
+        mode: 0640
+
+    - name: Download bird2 to apt's package cache
+      command:
+        cmd: /usr/bin/apt-get install -d -y bird2
+        warn: false
+
+    - name: Run upgrade commmands
+      shell:
+        cmd: |
+          systemctl stop bird bird6; \
+          apt-get purge -y bird; \
+          apt-get install -y bird2; \
+          mv /tmp/bird.conf.new /etc/bird/bird.conf; \
+          mkdir -p /var/log/bird; \
+          touch /var/log/bird/bird.log; \
+          chown bird:bird /etc/bird/bird.conf /var/log/bird /var/log/bird/bird.log; \
+          systemctl restart bird
+        warn: false
+      async: 600
+      poll: 10
+
+    - name: Remove the backup bird config file
+      file:
+        path: /tmp/bird.conf.old
+        state: absent
+
+  when: dpkg_query_bird is successful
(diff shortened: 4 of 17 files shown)


ansible/playbooks/roles/common/tasks/upgrade-bird.yml
```diff
@@ -0,0 +1,52 @@
...
+  command:
+    cmd: /usr/bin/apt-get install -t bionic-backports -d -y bird2
+    warn: false
+
+# Run the lights-out script that perform the update
+- name: Run upgrade commmands
+  shell: |
+    systemctl stop bird bird6; \
+    apt-get purge -y bird; \
+    apt-get install -t bionic-backports -y bird2; \
```
justinjpacheco, line 40: shouldn't we tell apt via /etc/apt/preferences.d that we prefer bird2 from bionic-backports? apt-get update/upgrade commands that are executed outside of this context wouldn't install these packages with the same specificity.

ansible/playbooks/roles/common/tasks/configure-networking.yml
```diff
@@ -7,6 +7,12 @@
     ansible_host: "{{ interfaces['transit'] | primary_ip(hostvars[inventory_hostname]) }}"
   when: use_primary_transit_interface
 
+- name: query_for_bird2_installation
+  command: dpkg-query -s bird2
+  failed_when: no
```
justinjpacheco, line 12: failed_when: no? we generally check the rc value of the registered variable for determining the failure status

tj90241: `dpkg-query -s bird2` returns rc = 1 when `bird2` is not installed, which inadvertently stops the play execution unless `failed_when: no` is specified.  The `when ...` clause in the `upgrade-bird.yml` import task uses the registered variable here.

justinjpacheco: is it possible to move this check into update-bird.yml and used in a block/when statement to keep all the upgrade bird2 stuff centralized?

justinjpacheco: `failed_when: dpkg_query_bird2.rc == 1` and `ignore_errors: true` similar to: https://github.com/bloomberg/chef-bcpc/blob/master/ansible/playbooks/roles/chef-server/tasks/chef-server.yml#L24

ansible/playbooks/roles/common/tasks/configure-networking.yml
```diff
@@ -33,6 +39,9 @@
 - import_tasks: configure-ubuntu-apt-sources.yml
   tags: [configure-ubuntu-apt-sources]
 
+- import_tasks: upgrade-bird.yml
+  when: (dpkg_query_bird2.rc == 1) and (configure_bgp | default(true))
```
justinjpacheco, line 43: the configure_bgp conditions should be removed and you can update the when to just check your dpkg_query_bird2 variable is succeeded (when: dpkg_query_bird2 is succeeded) https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#the-when-statement

ansible/playbooks/roles/common/tasks/configure-networking.yml
tj90241: yes, it's possible.  I had opted not to do that because it would result in most stanzas needing a `when: ...` clause in `upgrade-bird.yml`, which felt less optimal than hoisting it up in this manner and putting it next to the conditional `configure_bgp` variable --  I'm not opposed to changing it, though.

ansible/playbooks/roles/common/tasks/upgrade-bird.yml
tj90241: `bird2` only lives in `bionic-backports/universe`... so specifying `-t` either here or in `/etc/apt/preferences.d` is redundant.  I've removed `-t ...` from the this task and rebuild BVE, and it still works OK if you'd like me to push that change.
Specifying it in `/etc/apt/preferences.d` may create more hazards during release upgrades and things of the sort - I would think creating a file there is suboptimal for that reason alone.

ansible/playbooks/roles/common/tasks/configure-networking.yml
tj90241: Why should it be removed?  It's the same condition used by `configure-bgp`, which installs `bird2`.
```
- import_tasks: upgrade-bird.yml
  when: (dpkg_query_bird2.rc == 1) and (configure_bgp | default(true))

- import_tasks: configure-bgp.yml
  when: configure_bgp | default(true)
```

justinjpacheco: they should be removed everywhere is what i meant

ansible/playbooks/roles/common/tasks/upgrade-bird.yml
```diff
@@ -0,0 +1,52 @@
...
+    mv /tmp/bird.conf.new /etc/bird/bird.conf; \
+    mkdir -p /var/log/bird; \
```
comay, line 42: How does these next three lines interact with the recently added lines in `configure-bgp.yml` that also set up the logging. I realize it uses `-p` with `mkdir(1)` so if the directory is there, it doesn't do any harm but in which cases is this code necessary? Is it because the latter hasn't yet been deployed?

ansible/playbooks/roles/common/tasks/upgrade-bird.yml
comay: FWIW, it seems to me we should be consistent here if we can with other things from `bionic-backports` such as RabbitMQ. As such, I thought this was going to look similar to Justin's recent commit although I understand there may be differences due to the issues around doing an in-place upgrade.

Also, is the date we set for `bionic-backports` sufficient for this already? I thought it included the new RabbitMQ but didn't yet have `bird2` in it?

tj90241: aye, I will add it to `/etc/apt/preferences.d` then as there seems to be consensus for that.

The `bionic-backports` date is 05-02-2020.  I put up a PR for our BVE prop -- the upstream builds without a prop will get it organically from the current backports.

ansible/playbooks/roles/common/tasks/upgrade-bird.yml
tj90241: This basically mirrors much of what `configure-bgp.yml` does, as it starts `bird2` with the same config file that it provisions.  The re-used config file specifies `/var/log/bird/bird.log` as the logging location, and the service fails to start if the directory does not exist.
If it provides solace, this was meant to be more of a temporary task until we cutover everything to `bird2`, at which point it can be jettisoned if favor of everything being built new with `bird2` from the get-go (organically via `configure-bgp.yml`).

ansible/playbooks/roles/common/tasks/upgrade-bird.yml
tj90241: @justinjpacheco @comay - added an `/etc/apt/preferences.d` entry - please let me know what you think.  Just an FYI: I need to do some more functional testing on this change still.

ansible/playbooks/roles/common/tasks/configure-networking.yml
tj90241: Changed to match the style of the other code.

tj90241:
Small problem here - `calico-network` depends on the old `bird`:
```
$ apt-cache show calico-compute
Package: calico-compute
Priority: optional
...
Source: networking-calico
Version: 1:3.6.0~bionic
Depends: bird, calico-felix (>= 1.4), networking-calico (= 1:3.6.0~bionic), neutron-dhcp-agent, iptables-persistent
...
```

ansible/playbooks/roles/common/tasks/configure-networking.yml
tj90241: I believe we should leave `configure_bgp | default(true)` as the condition for both at present - if we are to not use `configure_bgp`, it is a separate change from this PR.  I have removed the `dpkg_query_bird2` from this section, though.

ansible/playbooks/roles/common/tasks/configure-networking.yml
```diff
@@ -1,6 +1,7 @@
 - name: gather host facts
   setup:
     gather_subset: all
+  tags: [configure-bgp]
```
justinjpacheco, line 4: why is this tag added here?

ansible/playbooks/roles/common/tasks/configure-networking.yml
```diff
@@ -33,8 +35,13 @@
 - import_tasks: configure-ubuntu-apt-sources.yml
   tags: [configure-ubuntu-apt-sources]
 
+- import_tasks: upgrade-bird.yml
+  when: configure_bgp | default(true)
```
justinjpacheco, line 39: is this still required?

justinjpacheco: also, i think this import would be better served inside the `configure-bgp` task instead of outside in the main task flow for configure-networking

tj90241: Resolved both in a new cleanup commit.

ansible/playbooks/roles/common/tasks/configure-networking.yml
tj90241: Created a new task for use with `configure-bgp.yml` so tag does not have to be added here.

ansible/playbooks/roles/common/tasks/define-transits.yml
```diff
@@ -0,0 +1,8 @@
+- name: gather host facts for interfaces
+  setup:
+    gather_subset: interfaces
+
+- name: define transit_interfaces
+  set_fact:
+    transit_interfaces: >
+      {{ interfaces['transit'] | transit_interfaces(ansible_facts) }}
```
tj90241, line 8: Can probably change the arg to be `ansible_facts['interfaces']` here now.

ansible/playbooks/roles/common/tasks/configure-networking.yml
tj90241: Removed in a later commit.

tj90241:
Very much not working yet, I'll remove the tag when this is more ready-ish.

ansible/playbooks/roles/common/tasks/upgrade-bird.yml
```diff
@@ -0,0 +1,62 @@
+---
+- name: query_for_bird_installation
+  command: dpkg-query -s bird
+  register: dpkg_query_bird
+  failed_when: dpkg_query_bird.rc != 0
+  ignore_errors: true
+  changed_when: false
+
+# Prepare old bird (in case we need to backout)
```
justinjpacheco, line 9: is this section needed outside of the "upgrade bird to bird2" block? if "query_for_bird_installation" finds that it doesn't need to do anything, do we still want to backup bird.conf?

ansible/playbooks/roles/common/defaults/main/base.yml
```diff
@@ -219,8 +219,6 @@ host_aggregates: []
 networking:
   mtu: 1500
 
-use_primary_transit_interface: true
```
justinjpacheco, line 222: bold move, but i think this is safe to remove now

ansible/playbooks/roles/common/tasks/upgrade-bird.yml
tj90241: ahhh, wasn't thinking - right, let me move it in.

ansible/playbooks/roles/common/defaults/main/base.yml
tj90241: 👍 - was debating it, feels out of scope for this PR, but it's code that's not really tested anymore and I'm not sure we can build over transits anymore without a few line changes elsewhere.

ansible/playbooks/roles/common/tasks/upgrade-bird.yml
```diff
@@ -0,0 +1,61 @@
...
+          mkdir -p /var/log/bird; \
+          touch /var/log/bird/bird.log; \
+          chown bird:bird /etc/bird/bird.conf /var/log/bird /var/log/bird/bird.log; \
+          systemctl restart bird
+        warn: false
+      async: 600
+      poll: 10
+  when: dpkg_query_bird is successful
+
+- name: Remove the backup bird config file
```
justinjpacheco, line 58: should this be moved inside the block?

tj90241: I was going to leave it outside as it's idempotent when the backup file is missing, but can pull it in.

tj90241:
I believe this is now ready, sans the two comments above.

ansible/playbooks/roles/common/tasks/configure-networking.yml
```diff
@@ -33,15 +9,15 @@
...
   tags: [configure-ubuntu-apt-sources]
 
+- import_tasks: configure-network-interfaces.yml
+  when: configure_network_interfaces | default(true)
+
 - import_tasks: configure-bgp.yml
-  when: configure_bgp | default(true)
+  tags: [configure-bgp]
 
 - import_tasks: configure-service-ip.yml
```
justinjpacheco, line 18: shouldn't this task be moved either above configure-bgp or included in configure-network-interfaces? prior to the re-org, it was above configure-network-interfaces? seems to make sense that the service0 interface is configure before bgp starts so that it has an ip to bind to (the service ip)

tj90241: I had thought `bird` created `service0`, but it looks like that isn't the case on closer inspection, so yes... it should.

tj90241:
Folded into https://github.com/bloomberg/chef-bcpc/pull/1695.

PR #1666, opened 2020-05-06
configure host aggregate task fixes

justinjpacheco:
- add any amount of hosts to existing aggregates if they are not ready
    a member of one

### DIFF
--- a/ansible/playbooks/roles/headnode/tasks/configure-host-aggregates.yml
+++ b/ansible/playbooks/roles/headnode/tasks/configure-host-aggregates.yml
@@ -1,15 +1,25 @@
 - name: configuring host aggregates
   shell: |
+    set -o pipefail
+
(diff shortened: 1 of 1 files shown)


ansible/playbooks/roles/headnode/tasks/configure-host-aggregates.yml
```diff
@@ -4,12 +4,19 @@
     hostname={{ item }}
     zone={{ hostvars[item]['zone'] }}
 
+    # compute service list json output
```
comay, line 7: Nit but shouldn't this be moved right before we echo and parse it? I realize it's not likely to change but it seems it belongs "closer" to lines 17-19?

justinjpacheco: yep

PR #1668, opened 2020-05-07
change default ceph pool pg[p]-num

justinjpacheco:
systems building for the first time with multiple ceph pools (due to
zone partitioning) will be unable to create those pools with the
previous pg[p]-num setting of 64. lowering this value to 8 gives us a bit
more runway

### DIFF
--- a/ansible/playbooks/roles/common/defaults/main/chef.yml
+++ b/ansible/playbooks/roles/common/defaults/main/chef.yml
@@ -9,6 +9,9 @@ chef_environment:
   chef_type: environment
   override_attributes:
     bcpc:
+      ceph:
+        pg_num: 32
+        pgp_num: 32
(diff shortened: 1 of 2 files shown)


ansible/playbooks/roles/common/defaults/main/chef.yml
```diff
@@ -9,6 +9,9 @@ chef_environment:
   chef_type: environment
   override_attributes:
     bcpc:
+      ceph:
+        pg_num: 64
+        pgp_num: 64
```
comay, line 14: As mentioned on Slack, while I wish we could use 128 here so both `1h1h` & `3h3w` come up healthy at the start, it looks like it's too large. So I would suggest 32 instead which is the minimum for the `1h1w` case.

PR #1669, opened 2020-05-07
Bump up default VM topology/hardware specs.

tj90241:
We've noticed an increased amount of false-positive build
failures recently due to haproxy timeouts and other items
that suggest the default topology/hardware.yml is too
conservative for recent changes to `chef-bcpc`.
    
This PR bumps up the amount of vCPUs/RAM allocated to
each VM (by default) to avoid false-positive build failures
and to reduce the amount of swapping that occurs during
builds.
    
With these changes, 1h1w builds ought to still easily work
on a host with quad-core CPU and 32GB of RAM, which is more
obtainable today in 2020 than it was a few years ago when
these settings were first instated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,12 +22,18 @@ deploy the chef-bcpc on hardware.
 ### Prerequisites
 
 * OS X or Linux
-* CPU that supports VT-x virtualization extensions
-* 16 GB of memory
+* Quad-core CPU that supports VT-x virtualization extensions
+* 32 GB of memory
 * 100 GB of free disk space
 * Vagrant 2.1+
 * VirtualBox 5.2+
 * git, curl, rsync, ssh, jq, make, ansible
+
+**NOTE**: It is likely possible to build an environment with 16GB of RAM or less
+if one is willing to make slight modifications to the virtual topology and/or
+change some of the build settings and overrides.  However, we've opted to spec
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -22,13 +22,19 @@ deploy the chef-bcpc on hardware.
...
 * Vagrant 2.1+
 * VirtualBox 5.2+
 * git, curl, rsync, ssh, jq, make, ansible
 
+**NOTE**: It is likely possible to build an environment with 16GB of RAM or less
+if one is willing to make slight modifications to the virtual topology and/or
+change some of the build settings and overrides.  However, we've opted to spec
+the minimum requirements slightly more aggressively and target hosts with 32GB
+RAM or more to provide the best out-of-the-box experience.
+
```
peter-at, line 37: LGTM
just nip picks: since notes probably for newbies I would link to the actual hardware.yml and topology.yml files

comay: We should also probably tell users that they can leave `hardware.yml` and `topology.yml` alone and instead create/copy into the override files. That's a better way IMO of carrying a local change to these parts of the configuration.

mihalis68: please include AMD-V when mentioning VT-x since AMD cpus are now a decent alternative choice for a build of this project

tj90241:
I think I've accounted for everyone's suggestions - please let me know if not!

PR #1670, opened 2020-05-07
rabbitmq: Use N/2+1 queue mirrors instead of N.

tj90241:
Performance testing has demonstrated that using N/2+1
queue mirrors instead of N improves maximum throughput
by ~40% and RPC call latency by 60% in a 3-node setup.
It is expected that such benefits would continue to
pile on when provisioning additional RMQ nodes.

Using slightly fewer queue mirrors should be safe as
we currently use a pause_minority partition/recovery
strategy, meaning that the cluster should pause before
or when all replicas of a queue are lost.

The only downside to this is that when a cluster member
drops, there may be some intermittent pausing as mirrors
are synchronized to account for the change in cluster
membership. This itself can be countered by tuning the
ha-sync-batch-size setting.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/rabbitmq.rb
+++ b/chef/cookbooks/bcpc/recipes/rabbitmq.rb
@@ -138,9 +138,17 @@ execute 'set rabbitmq user password' do
   command "rabbitmqctl change_password #{username} #{password}"
 end
 
+# Use n/2+1 queue mirrors as long as we have at least three headnodes.
+# If we have fewer than three headnodes, fallback to ha-all.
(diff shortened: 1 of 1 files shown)


tj90241:
Still need to do some testing to better understand what exactly happens when of the N/2+1 queue mirrors drops.  With `ha-all` it would be a NOP, but re-synchronization (and thus pausing) may occur when using fewer than N replicas and the effects of that need to be understood before this is made a production candidate.

tj90241:
Tested provision/failure scenario workflows -- all checks out!

PR #1672, opened 2020-05-13
per zone allocation

peter-at:
allow per-network-zone overrides to be effective for cpu/ram allocation ratio

**Testing performed**
Built multi-zone cluster for the test.
Re-Cheffing resulted with expected allocation ratios in nova.conf.
Allocate VMs to exercise the cpu allocation limit have expected results.


justinjpacheco:
very nice

PR #1674, opened 2020-05-14
Fix build failures due to PR #1672

peter-at:
fix PR #1672 - as template nova.conf.erb is used both in headnode and worknode recipe, there needs to be a guard for accessing nova-compute settings.

Signed-off-by: Peter Lee <plee38@bloomberg.net>

**Testing performed**
Rebuild using make destory; make create all;


comay:
Is there a reason these settings aren't part of `chef/cookbooks/bcpc/templates/default/nova/nova-compute.conf.erb` rather than `nova.conf`?

PR #1675, opened 2020-05-14
Upgrade to Calico 3.14, etcd3gw 0.2.5

comay:

Upgrade to Calico 3.14, etcd3gw 0.2.5

This set of changes upgrades Calico to version 3.14 and its dependency, etcd3gw, to 0.2.5.

This has been tested in a normal `1h1w` upstream configuration as well as multiple builds of various configurations including `3h3w` and `3h3w3s` using the larger Bloomberg virtual environment. It has also been tested on a single rack of equipment comprising of a `3h12w5s` cluster. Additional testing has been also done internally with other stakeholders. Finally, both multiple full builds as well as upgrades from the current Calico 3.6 environment have been done.


justinjpacheco:
do you want to make a pr for 3h3w3s?

chef/cookbooks/bcpc/files/default/calico/99-networking-calico.conf
```diff
@@ -0,0 +1,2 @@
+net.ipv4.ip_forward=1
+net.ipv6.conf.all.forwarding=1
```
tj90241, line 2: Our current production builds have `net.ipv6.conf.all.forwarding=0`.  I'm curious as to this one - is there something specific about IPv6 in these new Calico packages?

tj90241: ah nevermind, was looking at a headnode!

comay: This actually isn't new - these settings use to be appended to `/etc/sysctl.conf` through the `calico-compute` package `postinst` script. One of the issues I filed with them was that every time the package was rev'ed, another copy of these lines got appended. (!) In 3.14, this was changed at our request to just drop a `/etc/sysctl.d` fragment instead.

In this particular case since we need to make sure that `calico-compute` isn't installed in order to be able to switch the BIRD package versions, I introduced that fragment into our own automation since the package delivering it isn't being installed anymore. The IPv6 forwarding line could be removed but since we've been running with it forever, I'd rather leave that for another PR.

comay:
Yes, adding a `3h3w3s` example would be good. I'll work up a PR shortly on that.

chef/cookbooks/bcpc/files/default/calico/99-networking-calico.conf
tj90241: Thanks for the very through explanation - appreciate it!
I agree that it is a separate change and outside of this PR.

chef/cookbooks/bcpc/recipes/calico-work.rb
```diff
@@ -18,12 +18,24 @@
...
-  calico-dhcp-agent
-).each do |pkg|
-  package pkg
+package 'calico-compute' do
+  action :remove
+end
+
+package 'neutron-dhcp-agent'
+
+package %w(calico-felix networking-calico calico-dhcp-agent) do
```
tj90241, line 27: I believe we need to add `netfilter-persistent` and `iptables-persistent` here now that we are removing `calico-compute`.
It seems that we've been reliant on those packages, and they were formerly being pulled in by `calico-compute`.

comay: It’s actually by design they were left of as those packages are no longer necessary as IIRC, are not pulled in anymore by calico-compute

tj90241: Yes, you are right -- worked it out to a host in a bad state before the canary upgrade proceeded.

comay:
Closing this as it's been pulled into the 8.1-canary branch.

PR #1679, opened 2020-05-19
Enable openstack watcher in chef-bcpc

srsakhamuri:
Installs and enables the watcher using the config given in the control data

This change installs openstack watcher which aims to optimize the infrastructure usage on openstack installations. https://wiki.openstack.org/wiki/Watcher

Installed on plain environment and tested the basic commands


chef/cookbooks/bcpc/templates/default/watcher/haproxy.cfg.erb
```diff
@@ -0,0 +1,14 @@
+###############################################################################
```
justinjpacheco, line 1: does watcher have a healthcheck option?

srsakhamuri: Doesn't appear to have an healthcheck option

srsakhamuri: it doesn't appear watcher have an api-paste.ini in their repos, similar to [glance](https://github.com/openstack/glance/blob/master/etc/glance-api-paste.ini) or [heat](https://github.com/openstack/heat/blob/master/etc/heat/api-paste.ini). Also looks like the config need to be updated in order to generate one. There isn't one being developed at this point, looking at the [gerrit for watcher](https://review.opendev.org/#/q/status:open+project:openstack/watcher).

Probably we will just have to depend on plain httpchk until that time

chef/cookbooks/bcpc/templates/default/watcher/watcher.conf.erb
```diff
@@ -0,0 +1,38 @@
...
+
+[watcher_clients_auth]
+auth_url = <%= "https://#{@node['bcpc']['cloud']['fqdn']}:35357/" %>
+project_name = service
+project_domain_name = Default
+user_domain_name = Default
+username = <%= @config['watcher']['creds']['os']['username'] %>
+password = <%= @config['watcher']['creds']['os']['password'] %>
+auth_uri = <%= "https://#{@node['bcpc']['cloud']['fqdn']}:5000/" %>
+memcached_servers = <%= @headnodes.map{ |n| "#{n['service_ip']}:11211" }.join(',') %>
```
tj90241, line 36: Guessing the other `memcached_*` options are desirable here:
https://github.com/bloomberg/chef-bcpc/blob/master/chef/cookbooks/bcpc/templates/default/keystone/keystone.conf.erb#L16

They prevent timeouts from being too long when one of the headnodes is unavailable.

PR #1680, opened 2020-05-29
Make memcached memory limit a prop tunable.

npustchi:

**Describe your changes**
change to raise the memcached memory limit from 64M to 1G was made. This is now out of sync with automation. Made the memory limit a prop tunable that gets applied as part of automation.

**Testing performed**
I ran the chef receipe and checked the limit_maxbytes in memcached server.

**Additional context**
1- ssh to the head node
2- using find the port and IP address
ps -ef | grep memcached
3- use this https://lzone.de/cheat-sheet/memcached
To connect and get the stats 
telnet 10.0.0.2 11211
stats
Look for limit_maxbytes to get the limit bytes

### DIFF
--- a/chef/cookbooks/bcpc/attributes/memcached.rb
+++ b/chef/cookbooks/bcpc/attributes/memcached.rb
@@ -1,0 +1,6 @@
+###############################################################################
+# apache2
+###############################################################################
+
+# ells memcached how much RAM to use for item storage (in megabytes). Note carefully that this isn't a global memory limit
+default['bcpc']['memcached']['memory'] = 1024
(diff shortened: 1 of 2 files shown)


chef/cookbooks/bcpc/attributes/memcached.rb
```diff
@@ -0,0 +1,6 @@
+###############################################################################
+# apache2
```
justinjpacheco, line 2: apache2?

chef/cookbooks/bcpc/attributes/memcached.rb
```diff
@@ -0,0 +1,6 @@
...
+###############################################################################
+
+# ells memcached how much RAM to use for item storage (in megabytes). Note carefully that this isn't a global memory limit
```
justinjpacheco, line 5: s/ells/tells

PR #1687, opened 2020-06-26
Calico 3.15, other related changes/cleanup

comay:


comay:
Good point about the `system_ipv6_enable` tunable. I'll make that a separate commit after testing things out in `bve`.

PR #1690, opened 2020-06-30
Fix a potential build issue.

tj90241:
If pool names are created in the following order:
```
super-stuff
stuff
...
```

The 'stuff' pool is not created in the cinder and nova-head
recipes because the grep 'stuff' for the pool_name check
matches 'super-stuff' in the above example.  This results
in subsequent automation failing.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/cinder.rb
+++ b/chef/cookbooks/bcpc/recipes/cinder.rb
@@ -160,7 +160,7 @@ cinder_config.ceph_pools.each do |pool|
       ceph osd pool application enable #{pool_name} rbd
     DOC
 
-    not_if "ceph osd pool ls | grep -w #{pool_name}"
+    not_if "ceph osd pool ls | grep -w ^#{pool_name}$"
   end
 
   execute 'set ceph pool size' do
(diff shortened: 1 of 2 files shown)


tj90241:
There may be other instances of this idiom not limited to this, e.g.:
```
./nova-compute.rb:  only_if 'virsh net-list | grep -i default'
./nova-head.rb:  only_if 'nova-manage cell_v2 list_cells | grep cell1'
```

A network with the name of `default` in it, `cell11` existing... though I think in many cases it may be context-sensitive as to whether or not its impactful.
`grep not_if recipes/* | grep grep` lists several other possibly related occurrences in addition to the above that we might want to apply a similar more-specific regex match to.

tj90241:
I think there's a couple more we might want to tackle similarly - kind of unsure as to how aggressively to apply the pattern?
```
recipes/etcd-member.rb:  not_if "etcdctl --user root:#{root['password']} user list | grep -w root"
recipes/etcd-member.rb:  not_if "etcdctl --user root:#{root['password']} role list | grep -w rw"
recipes/etcd-member.rb:  not_if "etcdctl --user root:#{root['password']} role list | grep -w ro"
recipes/etcd-member.rb:  not_if "etcdctl --user root:#{root['password']} user list | grep -w server"
```

comay:
To be clear, I think those can be separate PRs but it's good for us to work through these.

tj90241:
+1, I'll file a JIRA now so it doesn't get lost.  This keeps things rolling for now though.

PR #1691, opened 2020-07-02
configure unbound on all nodes

justinjpacheco:


ansible/playbooks/roles/common/tasks/configure-systemd-resolved.yml
```diff
@@ -1,19 +1,11 @@
...
-- name: resolv.conf symlink
-  file:
-    src: /run/systemd/resolve/stub-resolv.conf
-    dest: /etc/resolv.conf
-    state: link
-    force: true
-  register: resolved
-
-- name: restart systemd-resolved
+- name: stop, disable and mask systemd-resolved
```
tj90241, line 1: good riddance!

ansible/playbooks/roles/common/tasks/configure-unbound.yml
```diff
@@ -0,0 +1,36 @@
+- name: sysctl allow non-local ip bind
+  sysctl:
```
tj90241, line 2: I think this will drop the sysctl in `/etc/sysctl.conf`- should we put it somewhere in `/etc/sysctl.d` instead?
Could set it here without making it persistent for the purposes of bootstrapping and set it in `/etc/sysctl.d/70-bcpc.conf` or elsewhere...

ansible/playbooks/roles/common/tasks/configure-unbound.yml
```diff
@@ -0,0 +1,36 @@
...
+  template:
+    src: unbound/default.j2
+    dest: /etc/default/unbound
+  register: unbound_default_conf
+
+- name: install unbound server config
+  template:
+    src: unbound/server.conf.j2
+    dest: /etc/unbound/unbound.conf.d/server.conf
+  register: unbound_conf
```
tj90241, line 30: Does this registration override the one defined above in `remove default unbound conf files`?
If this file does not change, but the `root-auto-trust-anchor-file.conf` is deleted, it may be possible that `unbound` is not reloaded?

ansible/playbooks/roles/common/tasks/configure-unbound.yml
justinjpacheco: yah, i could move it to a .d folder, i'd probably want to move 70-bcpc.conf (00-bcc.conf?) to an earlier point in the automation as both would define the same param (ip bind)

ansible/playbooks/roles/common/tasks/configure-unbound.yml
justinjpacheco: it's possible, i'd i have to test that scenario

ansible/playbooks/roles/common/templates/unbound/server.conf.j2
```diff
@@ -0,0 +1,37 @@
...
+{% if 'headnodes' is in group_names %}
+    forward-zone:
+      name: consul
+      forward-addr: {{ hostvars[inventory_hostname]['interfaces']['service']['ip'] }}@8600
+    forward-zone:
+      name: {{ cloud_domain }}
+      forward-addr: {{ hostvars[inventory_hostname]['interfaces']['service']['ip'] }}@5300
+    forward-zone:
+      name: .
+{% for ns in dns['servers'] %}
```
tj90241, line 30: Is declaring a forward-zone here possibly redundant for headnodes, now that `resolv.conf` _also_ has the DNS servers?  My initial thinking of the headnode name resolution is:
* `/etc/hosts` is checked, then `/etc/resolv.conf`/DNS per `nsswitch`
*`unbound` is checked via `127.0.0.1`, which should either return a Consul-resolved record or `SERVFAIL`
* The remaining DNS servers in `/etc/resolv.conf` are consulted in sequential order

So for headnodes and worknodes which proxy to these headnodes for name resolution, putting the DNS servers both here and in `/etc/resolv.conf` seems like it can result in DNS servers getting queried twice if resolution results in `SERVFAIL`

That being said, putting nameservers here and not in `/etc/resolv.conf` is interesting and maybe even ideal as `unbound` can cache the records accordingly, whereas the `libc` resolver will not?

virtual/provisioners/download-debs.sh
```diff
@@ -0,0 +1,29 @@
...
+set -e
+
+http_proxy=${1}
+https_proxy=${2}
+
+main() {
+    export http_proxy=${http_proxy}
+    export https_proxy=${https_proxy}
+    sudo -E apt-get update
+    sudo -E apt-get install --download-only -y bird tinyproxy unbound
```
tj90241, line 26: `bird` -> `bird2` once this gets rebased -- FYI (which itself will also need `init-system-helpers` from backports).

comay: Does `init-system-helpers` need to be specified explicitly though? I thought you determined that it wasn't actually necessary or did I misunderstand?

ansible/playbooks/roles/common/templates/unbound/server.conf.j2
justinjpacheco: my thinking here is that unbound becomes the primary source for dns resolution which affords it the opportunity to cache every result. the remaining dns servers listed in `/etc/resolv.conf` are there as last ditch efforts when unbound isn't available (eg: when the cluster is coming online for the first time and consul/pdns/vip aren't available yet)

virtual/provisioners/download-debs.sh
justinjpacheco: this will be updated to `bird2` when 8.1 is released

tj90241: @comay I'd have to check again just to confirm as it's a sticky issue - for `configure-system.yml` I believe I did find that unless we had an `apt-preferences` which explicitly said to get `init-system-helpers` from backports, the `apt-get` would fail.

PR #1694, opened 2020-07-10
Added authentication for apache status module

ajaytikoo:

**Describe your changes**
I have made changes that are related to using Apache Status module. This module is disabled in the attributes file chef-bcpc/chef/cookbooks/bcpc/attributes/apache2.rb, and that stays same. So, the module will stay disabled, by default, even after my changes. What I have changed specifically is as described below:
- Changed the file ansible/playbooks/roles/common/defaults/main/chef.yml to include a default username and password for the user that will be allowed access to the server_status page, if the module is enabled.
- Changed the file chef/cookbooks/bcpc/recipes/apache2.rb so that it creates an htpasswd file that will be used by the status module for user authentication.
- Added a new template file chef/cookbooks/bcpc/templates/default/apache2/status.conf.erb, which configures the status module to use authentication based on the htpasswd file created in above step.

**Testing performed**
I have tested the changes successfully in my virtual environment.

**Additional context**
None.


chef/cookbooks/bcpc/recipes/apache2.rb
```diff
@@ -85,6 +85,37 @@
...
+# apache status module
+if node['bcpc']['apache2']['status']['enabled']
+  # create htpasswd file for apache status
+  apache_status_username = config['apache']['status']['username']
+  apache_status_password = config['apache']['status']['password']
+  execute 'set password for apache_status user' do
+    command "htpasswd -cb /etc/apache2/server_status_htpasswd #{apache_status_username} #{apache_status_password}"
+    creates '/etc/apache2/server_status_htpasswd'
+    sensitive true
+    notifies :restart, 'service[apache2]', :delayed
```
tj90241, line 97: Can you add an `only_if htpasswd -bv /etc/apache2/server_status_htpasswd #{apache_status_username} #{apache_status_password}` (or similar) here?

Making this state idempotent will prevent otherwise unnecessary `apache2` restarts during a full Chef client run.  This is desirable as it will reduce the amount of API disturbance noticed when the password is not changing.

tj90241: err, in this case it looks idempotent... but if the password changes - it's not updated.  I'm not sure if this is due to the `creates`?

chef/cookbooks/bcpc/recipes/apache2.rb
```diff
@@ -85,6 +85,37 @@
...
+  end
+
+  # secure the htpasswd file for apache_status
+  file '/etc/apache2/server_status_htpasswd' do
+    mode '0600'
```
tj90241, line 102: 0640/root/www-data?

chef/cookbooks/bcpc/templates/default/apache2/status.conf.erb
```diff
@@ -0,0 +1,26 @@
+<IfModule mod_status.c>
+    # Allow server status reports generated by mod_status,
+    # with the URL of http://servername/server-status
+
+    <Location /server-status>
```
tj90241, line 5: Perhaps add a href conditionally here?
`chef/cookbooks/bcpc/templates/default/apache2/index.html.erb`

That way it is advertised as one of the dashboards alongside RMQ/HAproxy/etc.

chef/cookbooks/bcpc/recipes/apache2.rb
ajaytikoo: `creates` will result in file not getting overwritten if it exists. So we do have an issue with updating the password here. I guess that using `only_if`, as you suggested, will be better here.

chef/cookbooks/bcpc/templates/default/apache2/status.conf.erb
```diff
@@ -0,0 +1,26 @@
+<IfModule mod_status.c>
```
justinjpacheco, line 1: this file already exists under /etc/apache2/mods-available. why aren't we just enabling the configuration that is already there?

justinjpacheco: also, don't we need to enable the mod_status module for this to work? i thought this was disabled in an earlier commit

chef/cookbooks/bcpc/recipes/apache2.rb
justinjpacheco: using bcpc.apache2.status.enabled as the boolean to determine if we should enable the status module AND put it behind authentication seems to be a bit overloaded

chef/cookbooks/bcpc/templates/default/apache2/status.conf.erb
ajaytikoo: enabling or disabling of the status module is already taken care of here:
https://github.com/bloomberg/chef-bcpc/blob/83567d5e804fd0dfd1349bde5dfc667c039fb897/chef/cookbooks/bcpc/recipes/apache2.rb#L53

justinjpacheco: yah, i had to look at the recipe to remind myself how this works

chef/cookbooks/bcpc/recipes/apache2.rb
ajaytikoo: @justinjpacheco, apache status module will be enabled ONLY if authentication is enabled for it. That is what I was given go ahead for. Does that answer your question, or do we still want to have a separate flag for authentication?

chef/cookbooks/bcpc/recipes/apache2.rb
ajaytikoo: yes, that makes sense, though at this time `www-data` user is the only member in the group `www-data`. I will make that change.

ansible/playbooks/roles/common/defaults/main/chef.yml
```diff
@@ -182,6 +182,10 @@ chef_databags:
       username: guest
       password: Rv1GjTlGmfHPiSRt5bqK
       cookie: Y7SAgyFKPcxlzLdT5lJV
+    apache:
```
justinjpacheco, line 185: could you move this up so that it appears in the correct lexical order?

chef/cookbooks/bcpc/templates/default/apache2/status.conf.erb
ajaytikoo: will do

justinjpacheco: 👍

ansible/playbooks/roles/common/defaults/main/chef.yml
ajaytikoo: leaving it as is because this file is not in lexical order even before my changes.

chef/cookbooks/bcpc/recipes/apache2.rb
tj90241: It's probably not too big of a deal and can stay as-is -- just commenting as least-priv principle suggests that `apache2` (`www-data`) only needs to read the password, and `root` should be able to write it.

ajaytikoo: done

ajaytikoo:
I will be approaching this a bit differently, so there will be less changes needed to this repo

PR #1698, opened 2020-07-30
Use the multiple package support (see https://docs.chef.io/resources/…

mihalis68:
…package/)

I am not requesting this be merged, just sharing to encourage widespread review and testing. I have not even tested yet but it seems to save 23% of build time!!!


mihalis68:
Replaced by https://github.com/bloomberg/chef-bcpc/pull/1699

PR #1702, opened 2020-08-07
virtual environment provisioner refactoring

justinjpacheco:
- refactored vagrant provisioners into a single provisoner script
    called finish.sh

  - added ability to pre-load debs during provisioning

  - added support for custom apt url and key via the BCC_APT_URL and BCC_APT_KEY_URL
    environment variables

  - added support for operations user creation at provision time
    including ssh auth key

  - removed deprecated use of password auth during the configure-operator
    ansible task in favor of ssh key auth with the operations user

  - removed unused 01-netcfg.yml file


virtual/provisioners/finish.sh
```diff
@@ -0,0 +1,169 @@
...
+        fi
+
+        if ! grep "${swap_file}" /etc/fstab; then
+            echo "${swap_file}  none  swap  sw 0  0" >> /etc/fstab
+        fi
+    fi
+}
+
+function configure_linux_kernel {
+    apt-get install -y linux-generic-hwe-18.04-edge
```
tj90241, line 166: The kernel selection is a variable in Ansible - we might want to either hard code the kernel selection and not make it optional, or have this respect the variable somehow.  I cannot currently think of a way to do the latter though.  Thoughts?

justinjpacheco: > we might want to either hard code the kernel selection and not make it optional
im not sure i follow. it is both hardcoded and not optional

i had a similar thought regarding the variable in ansible. i don't know of a way to extract that information from ansible in a reliable way so this might be something that needs to be kept in sync for the time being. the goal here was to just have the vms come back up with a newer kernel than what we got from the baseline vbox, if they're slightly out of sync, i don't see that as a huge problem, the system still reboots with a 5.4.x kernel

tj90241: Oh, I meant not allowing it to be user-specified at the project level.  I guess it would still be a variable somewhere though.

I agree with you on the second point -- this is certainly a lot better than what we have today, and it's always something we could revisit at a later time if need be.

comay: Could the name of the kernel (eg, linux-generic-hwe-18.04-edge) be turned at last into a variable that is read by the script? Something like a new file `virtual/kernel-version` that contains the string `LINUX_KERNEL=linux-generic-hwe-18.04-edge` and than `finish.sh` could simply source this file and use that variable.

virtual/lib/util.rb
```diff
@@ -13,6 +13,9 @@ def self.vbox_name(name)
   end
 
   def self.mount_apt_cache(config)
+    if ENV.key?('BCC_DISABLE_APT_CACHE')
```
comay, line 16: Nit - we haven't prefixed environment variables used by `Vagrantfile` or various functions until now. I can understand doing so but is it worth being consistent? Either we always prefixed with `BCC_` or with don't and instead choose unique enough names?

virtual/provisioners/finish.sh
```diff
@@ -0,0 +1,169 @@
...
+    --name "$(basename "$0")" \
+    --options "" \
+    -- "$@"
+)
+
+eval set --"$opts"
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --apt-url)
```
comay, line 50: Nit - could you please sort these case blocks?

virtual/provisioners/finish.sh
```diff
@@ -0,0 +1,169 @@
...
+deb ${apt_url} bionic-updates main restricted universe multiverse
+EOF
+    fi
+
+    apt-get update
+}
+
+function download_debs {
+    apt-get install --download-only -y -t bionic-backports \
+        bird2 init-system-helpers
```
comay, line 141: Does `init-system-helpers` actually need to be spelled out here? I thought Tyler determined with the 8.1 upgrade playbook that it wasn't actually necessary?

justinjpacheco: i tried multiple permutations of this, it needs to be referenced directly against the bionic-backports pool in order to download the correct dependences. i believe this is why init-system-helpers is pinned in the bird ansible playbook

virtual/lib/util.rb
justinjpacheco: it made sense to me to have external sources of information (in this case, environment variables) easily identifiable with a known prefix (BCC_). the only env var that doesn't have this is the network vm env var, which im happy to change but i thought that would be too disruptive

virtual/provisioners/finish.sh
comay: Thanks for the explanation. For consistency with the pinning done for `bird2`, would it instead make sense to drop in the same `apt-preferences` done by Ansible?

virtual/lib/util.rb
comay: There's also `LIBVIRT_DRIVER` & `ENABLE_VBOX_SUFFIX`. I'm fine leaving this as-is but it seemed strange to have this mixture of environment variable names. For the record, I like the idea of having the `BCC_` prefix for the reasons you stated.

justinjpacheco: i'll make the update to have them all fall inline

virtual/provisioners/finish.sh
justinjpacheco: ok

virtual/provisioners/finish.sh
justinjpacheco: sure thing, that's a nice idea

virtual/provisioners/finish.sh
justinjpacheco: i thought about dropping the pref file but i figured that because automation was going to do the same thing, i didn't want to run into a situation where we were dropping files not tracked by automation. to put that another way: if the file name changed in automation, then the file dropped by the provisioner wouldn't be cleaned up unless explicitly stated in the automation. in general, i'd prefer not to do too much of what bcc automation is going to do, just bootstrap it enough to get it off the ground

virtual/provisioners/finish.sh
comay: Although it's perhaps overkill, I just noticed that we actually install both `linux-{{ kernel_variant }}` & `linux-tools-{{ kernel_variant }}` so perhaps the variable should only include the version proper (eg, `generic-hwe-18.04-edge`) and we instead install both of these packages.

virtual/provisioners/finish.sh
comay: Makes sense.

virtual/provisioners/finish.sh
justinjpacheco: i think it's ok for the additional kernel packages to come in during normal automation, im not sure if tools or any other package has any impact on the running kernel

comay: Fair enough although I do suggest having the new `kernel-version` file strictly have only the version (eg, `generic-hwe-18.04-edge`) in case we need to later add any additional tagged packages.

So the install of the kernel would be `apt-get install -y linux-"${LINUX_KERNEL}"`

And perhaps the name LINUX_KERNEL should be changed to KERNEL_VERSION?

justinjpacheco: done

PR #1703, opened 2020-08-11
added support for generating chef databag secrets

justinjpacheco:
- added make target (generate-chef-databags) that calls a python
    script that will generate all chef databag values

  - updated ansible template generation to include file mode setting
    to account for ansibles switch from the default 666 filemode to 600

  - updated README.md with some additional build notes


ansible/playbooks/roles/common/tasks/configure-service-ip.yml
```diff
@@ -2,12 +2,14 @@
   template:
     src: systemd/network/10-service0.netdev
     dest: /etc/systemd/network/10-service0.netdev
+    mode: 0666
```
comay, line 5: This seems too permissive - perhaps 0644 and if necessary, owned by a different user?

ansible/playbooks/roles/common/tasks/configure-service-ip.yml
```diff
@@ -2,12 +2,14 @@
...
   register: networkd
 
 - name: service ip network config
   template:
     src: systemd/network/20-service0.network
     dest: /etc/systemd/network/20-service0.network
+    mode: 0666
```
comay, line 12: This seems too permissive - perhaps 0644 and if necessary, owned by a different user?

ansible/playbooks/roles/common/tasks/configure-service-ip.yml
justinjpacheco: i was keeping in line with what we currently have 🤷‍♂️

comay: Are you sure they're 0666? I see 0644 in use on one of the real BCC clusters.

README.md
```diff
@@ -51,9 +51,18 @@ changes to them instead.
 * If additional CA certificates are required (e.g. for a proxy), set the variables TBD
 * From the root of the chef-bcpc git repository run the following command:
 
+Create a Python virtual environment (virtualenv) and activate it
+
+```shell
+python3 -mvenv venv
+source venv/bin/activiate
```
mihalis68, line 58: typo, should be activate

README.md
```diff
@@ -51,9 +51,18 @@ changes to them instead.
...
+pip install PyYaml ansible netaddr pyOpenSSL pycryptodome
+```
+
 To create a virtualbox build (the default):
 
 ```shell
+make generate-chef-databags
```
mihalis68, line 65: where do the resulting files go? I did this but I don't see any files added at least according to git status

justinjpacheco: the generated file is ignored by git via .gitignore (that change is in this pr. see .gitignore). the file is put here: ansible/group_vars/all/chef_databags.yml

mihalis68: thanks!

PR #1704, opened 2020-08-12
Changes to better align `create-virtual-hosts` step with metal nodes

comay:
Update Bento box version for ubuntu-18.04 to 202005.21.0
Treat BCC_DEPLOY_NETWORK_VM the same as other boolean environment variables
Do the Ansible equivalent of a apt(8) dist upgrade as part of provisioning

### DIFF
--- a/virtual/provisioners/finish.sh
+++ b/virtual/provisioners/finish.sh
@@ -92,6 +92,7 @@ function main {
     create_operations_user
     configure_vagrant_user
     configure_apt
+    upgrade_system
     download_debs
     configure_swap
     configure_linux_kernel
@@ -136,6 +137,15 @@ EOF
     apt-get update
 }
 
+# Taken from Ansible's dist upgrade logic for apt(8)
+function upgrade_system {
+    env DEBIAN_FRONTEND='noninteractive' DEBIAN_PRIORITY='critical' \
+        apt-get -y \
+            -o 'Dpkg::Options::=--force-confdef' \
+            -o 'Dpkg::Options::=--force-confold' \
+        dist-upgrade
+}
+
 function download_debs {
     apt-get install --download-only -y -t bionic-backports \
         bird2 init-system-helpers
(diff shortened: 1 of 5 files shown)


virtual/provisioners/finish.sh
```diff
@@ -165,10 +176,20 @@ function configure_swap {
...
     KERNEL_VERSION_FILE=/vagrant/kernel-version
     if test -f "${KERNEL_VERSION_FILE}"; then
-        # shellcheck disable=SC1090,SC1091
+        # shellcheck disable=SC1090
         source "${KERNEL_VERSION_FILE}"
         apt-get install -y "linux-${KERNEL_VERSION}"
     fi
 }
 
+function enable_ipv6 {
```
justinjpacheco, line 185:
```suggestion
function disable_ipv6 {
```

virtual/provisioners/finish.sh
```diff
@@ -136,6 +138,15 @@ EOF
     apt-get update
 }
 
+# Taken from Ansible's safe upgrade logic for apt(8)
```
justinjpacheco, line 141: do we want to `dist-upgrade` instead?

virtual/provisioners/finish.sh
```diff
@@ -92,9 +92,11 @@ function main {
     create_operations_user
     configure_vagrant_user
     configure_apt
+    upgrade_apt
```
justinjpacheco, line 95: should we combine these two apt tasks together?

virtual/provisioners/finish.sh
comay: LOL, I went back and forth on this. I tried to align the name of the function with the environment variable so if the latter is undefined, then we don't enable (well, proactively then disable) IPv6. If it is set, then we enable (well, leave it enabled) IPv6.

If you think it really is clearer the other way, I can change it but that was my rationale.

virtual/provisioners/finish.sh
comay: Currently the 8.1 upgrade playbook does the equivalent of a `apt-get upgrade` with the possible addition of new packages. Looking at the Ansible sources, I mimicked what that does.

PR #1707, opened 2020-08-25
changed permissions for socket files for ceph and haproxy

ajaytikoo:
Changed the default permissions for the socket files that ceph and haproxy create. The group has the write permission. This is needed so that Telegraf process has permissions  for these socket files even after these services are restarted.

**Testing performed**
Tested this in my lab

### DIFF
--- a/chef/cookbooks/bcpc/templates/default/ceph/ceph.conf.erb
+++ b/chef/cookbooks/bcpc/templates/default/ceph/ceph.conf.erb
@@ -9,6 +9,7 @@ mon host = <%= @headnodes.map{ |n| n['service_ip'] }.join(',') %>
 max open files = <%= node['bcpc']['ceph']['max_open_files'] %>
 rbd default features = <%= node['bcpc']['ceph']['rbd_default_features'] %>
 osd crush initial weight = <%= node['bcpc']['ceph']['osd_crush_initial_weight'] %>
+admin socket mode = 0775
 
 [mon]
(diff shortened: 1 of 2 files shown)


chef/cookbooks/bcpc/templates/default/ceph/ceph.conf.erb
```diff
@@ -9,6 +9,7 @@ mon host = <%= @headnodes.map{ |n| n['service_ip'] }.join(',') %>
 max open files = <%= node['bcpc']['ceph']['max_open_files'] %>
 rbd default features = <%= node['bcpc']['ceph']['rbd_default_features'] %>
 osd crush initial weight = <%= node['bcpc']['ceph']['osd_crush_initial_weight'] %>
+admin socket mode = 0775
```
justinjpacheco, line 12: does the group require wx and others x?

chef/cookbooks/bcpc/templates/default/haproxy/haproxy.cfg.erb
```diff
@@ -7,7 +7,7 @@ global
   user haproxy
   group haproxy
   pidfile /var/run/haproxy.pid
-  stats socket /var/run/haproxy/haproxy.asok level admin
+  stats socket /var/run/haproxy/haproxy.asok user root group haproxy mode 775 level admin
```
justinjpacheco, line 10: same question as above

PR #1708, opened 2020-08-27
added functionality for scheduling nova db archive-deleted-rows

ajaytikoo:
**Describe your changes**
Added functionality to archive rows for deleted instances from instance tables in nova database to the shadow tables. Unless someone specifically enables this functionality, it is not enabled by default. So, no changes are applied to the system unless one explicitly enables this functionality in chef.yml file.

**Testing performed**
I have tested these changes in my lab setup (bve), both with this functionality enabled and disabled, though it is left disabled in this commit.

**Additional context**
As instances are deleted, their records are left unnecessarily in the instance tables. Over the period of time, this makes these tables large resulting in performance overheads. Archiving these rows into shadow tables periodically

### DIFF
--- a/chef/cookbooks/bcpc/recipes/nova-head.rb
+++ b/chef/cookbooks/bcpc/recipes/nova-head.rb
@@ -405,3 +405,19 @@ execute 'wait for nova to come online' do
   retries 30
   command 'openstack compute service list'
 end
+
+cron 'nova-manage db archive' do
+  action  :create
+  minute  node['bcpc']['nova']['db-archive']['cron_minute']
+  hour    node['bcpc']['nova']['db-archive']['cron_hour']
+  weekday node['bcpc']['nova']['db-archive']['cron_weekday']
+  day     node['bcpc']['nova']['db-archive']['cron_day']
+  month   node['bcpc']['nova']['db-archive']['cron_month']
+  user    'root'
+  command <<-DOC
+    /usr/local/bcpc/bin/if_leader \
+    nova-manage db archive_deleted_rows --until-complete --verbose 2>&1 \
+    | logger -t nova-db-archive-deleted-rows
+  DOC
+  only_if { node['bcpc']['nova']['db-archive']['enabled'] }
+end
--- a/chef/cookbooks/bcpc/attributes/nova.rb
+++ b/chef/cookbooks/bcpc/attributes/nova.rb
@@ -96,3 +96,16 @@ default['bcpc']['nova']['quota']['project']['admin']['gigabytes'] = -1
 #
 default['bcpc']['nova']['vendordata']['name'] = nil
 default['bcpc']['nova']['vendordata']['port'] = 8444
+
+# nova db archive deleted rows
+#
+
+# is nova db archive enabled
+default['bcpc']['nova']['db-archive']['enabled'] = false
+
+# if enabled, what is the schedule to run
+default['bcpc']['nova']['db-archive']['cron_month'] = '*'
+default['bcpc']['nova']['db-archive']['cron_day'] = '*'
+default['bcpc']['nova']['db-archive']['cron_weekday'] = '6'
+default['bcpc']['nova']['db-archive']['cron_hour'] = '4'
+default['bcpc']['nova']['db-archive']['cron_minute'] = '0'


ansible/playbooks/roles/common/defaults/main/chef.yml
```diff
@@ -30,6 +30,14 @@ chef_environment:
             - lahf_lm
             - popcnt
             - sse4a
+        db-archive-deleted-rows:
```
justinjpacheco, line 33: i would move this to a file in chef attributes as the global defaults. if they need to changed a per cluster basis, we can add them to the prop files

chef/cookbooks/bcpc/recipes/nova-head.rb
```diff
@@ -405,3 +405,15 @@
...
+cron 'nova-manage db archive' do
+  action  :create
+  minute  node['bcpc']['nova']['db-archive-deleted-rows']['schedule']['cron_minute']
+  hour    node['bcpc']['nova']['db-archive-deleted-rows']['schedule']['cron_hour']
+  weekday node['bcpc']['nova']['db-archive-deleted-rows']['schedule']['cron_weekday']
+  day     node['bcpc']['nova']['db-archive-deleted-rows']['schedule']['cron_day']
+  month   node['bcpc']['nova']['db-archive-deleted-rows']['schedule']['cron_month']
+  user    'root'
+  command '/usr/local/bcpc/bin/if_leader nova-manage db archive_deleted_rows --until-complete --verbose 2>&1 | logger -t nova-db-archive-deleted-rows'
+  only_if { node['bcpc']['nova']['db-archive-deleted-rows']['enabled'] rescue false }
```
justinjpacheco, line 418: is there a reason you're using rescue false here?

chef/cookbooks/bcpc/recipes/nova-head.rb
```diff
@@ -405,3 +405,15 @@
   retries 30
   command 'openstack compute service list'
 end
+
+cron 'nova-manage db archive' do
```
justinjpacheco, line 409: if you wrap this block around an if statement you can achieve a few things:
1. create a ref to node['bcpc']['nova']['db-archive-deleted-rows'] so that you're not repeating yourself so often
2. avoid the only_if clause

chef/cookbooks/bcpc/recipes/nova-head.rb
```diff
@@ -405,3 +405,15 @@
...
+  action  :create
+  minute  node['bcpc']['nova']['db-archive-deleted-rows']['schedule']['cron_minute']
```
justinjpacheco, line 411: could 'cron' over 'schedule' be a better key? then you can just use 'minute' vs 'cron_minute'

chef/cookbooks/bcpc/recipes/nova-head.rb
justinjpacheco, line 418:
```suggestion
  only_if { node['bcpc']['nova'].fetch('db-archive-deleted-rows',{}).fetch('enabled',false) }
```

chef/cookbooks/bcpc/recipes/nova-head.rb
```diff
@@ -405,3 +405,15 @@
...
+  hour    node['bcpc']['nova']['db-archive-deleted-rows']['schedule']['cron_hour']
+  weekday node['bcpc']['nova']['db-archive-deleted-rows']['schedule']['cron_weekday']
+  day     node['bcpc']['nova']['db-archive-deleted-rows']['schedule']['cron_day']
+  month   node['bcpc']['nova']['db-archive-deleted-rows']['schedule']['cron_month']
```
comay, line 415: Along with Justin's suggestion on moving the values into Chef attributes, I'd suggest removing the `['schedule']` element from the name of the attribute for consistency with other recipes that have the `cron_*` names under the top-level name, eg

node['bcpc']['nova']['db-archive-deleted-rows']['cron_minute']

justinjpacheco:
overall i think this is a great feature, nice work

chef/cookbooks/bcpc/recipes/nova-head.rb
justinjpacheco: i would use the above over rescue false

ansible/playbooks/roles/common/defaults/main/chef.yml
justinjpacheco: i would put this under the `nova` key and call it `db-archive`

chef/cookbooks/bcpc/recipes/nova-head.rb
justinjpacheco, line 411: i'd rename the top level from `db-archive-deleted-rows` to just `db-archive`

chef/cookbooks/bcpc/recipes/nova-head.rb
```diff
@@ -405,3 +405,15 @@
...
+
+cron 'nova-manage db archive' do
+  action  :create
+  minute  node['bcpc']['nova']['db-archive']['cron_minute']
+  hour    node['bcpc']['nova']['db-archive']['cron_hour']
+  weekday node['bcpc']['nova']['db-archive']['cron_weekday']
+  day     node['bcpc']['nova']['db-archive']['cron_day']
+  month   node['bcpc']['nova']['db-archive']['cron_month']
+  user    'root'
+  command '/usr/local/bcpc/bin/if_leader nova-manage db archive_deleted_rows --until-complete --verbose 2>&1 | logger -t nova-db-archive-deleted-rows'
```
justinjpacheco, line 417: could we break this statement up with newlines to make this easier to read?

chef/cookbooks/bcpc/recipes/nova-head.rb
```diff
@@ -405,3 +405,15 @@
...
+  only_if { node['bcpc']['nova'].fetch('db-archive', {}).fetch('enabled', false) }
```
justinjpacheco, line 418:
```suggestion
  only_if { node['bcpc']['nova']['db-archive']['enabled'] }
```

now that db-archive is an attribute default, we can drop the check for it and look directly at enabled

justinjpacheco:
it just occurred to me that if this is ever enabled then disabled there's nothing here that will remove the cron. is this a concern @comay @ajaytikoo ?

chef/cookbooks/bcpc/recipes/nova-head.rb
```diff
@@ -405,3 +405,19 @@
...
+
+cron 'nova-manage db archive' do
+  action  :create
+  minute  node['bcpc']['nova']['db-archive']['cron_minute']
+  hour    node['bcpc']['nova']['db-archive']['cron_hour']
+  weekday node['bcpc']['nova']['db-archive']['cron_weekday']
+  day     node['bcpc']['nova']['db-archive']['cron_day']
+  month   node['bcpc']['nova']['db-archive']['cron_month']
+  user    'root'
+  command db_archive_command
```
justinjpacheco, line 421:
```suggestion
  command <<-DOC
    /usr/local/bcpc/bin/if_leader \
      nova-manage db archive_deleted_rows \
        --until-complete --verbose 2>&1 \
      | logger -t nova-db-archive-deleted-rows
  DOC
```

PR #1709, opened 2020-08-28
Work-around change put in for TTA-2019-002 that prevents VXLAN use by VMs

comay:
Although neither VXLAN or IP in IP encapsulation is used here by Calico,
the change in https://github.com/projectcalico/felix/pull/2063 to address
a security issue (see Tigera Technical Advisory TTA-2019-002) prevents
instances from using VXLAN or IP in IP encapsulation themselves. As a
work-around for the VXLAN case, an alternate port can be set for Calico
to "use" which allows the use of the standard VXLAN port by instances
within OpenStack.


comay:
Actually, I didn't even associate port 9 with Wake-on-LAN! I figured anything single digits would be highly unlikely to be used by any real application since those were reserved for the most legacy of apps like `discard`. I suspect any even safer thing would be a well-known port that is only used for TCP (such as 80) but that might cause some heartburn if someone say a DROP rule for port 80 (but UDP!)

PR #1711, opened 2020-08-31
Add additional packages/configuration items

tj90241:
Adding in the following components:
  * `lm-sensors`
  * `build-essential`
  * `lvm2`


chef/cookbooks/bcpc/recipes/common-packages.rb
```diff
@@ -48,7 +49,13 @@
   vim
   ksh
   bash-completion
-)
+) do
+  :options --no-install-recommends
+end
+
+cookbook_file '/etc/sensors.d/power-acpi-monitor.conf' do
```
comay, line 56: Is the original `/etc/sensors3.conf` file still left on the file-system?

tj90241: `/etc/sensors3.conf` is technically owned by `libsensors4` -- it's a part of the packaging and could be overwritten in a future upgrade!
The file already has other contents and we should try to avoid writing to it if possible, hence the change in direction here.

comay: Ah, great find then. So if it's not meant to be "edited", then I agree with your assessment although perhaps we should devise some ad-hoc Ansible to deploy reverting the packaged version of the file after this gets pushed & deployed.

PR #1716, opened 2020-09-03
configure-system: Add CPU microcode loader logic

tj90241:
Conditionally enable or disable CPU microcode loading
before the system boots.


tj90241:
Will post testing results shortly.

ansible/playbooks/roles/common/defaults/main/base.yml
```diff
@@ -270,6 +270,10 @@ hardware_io_scheduler: deadline
 # providing consistent, deterministic device naming conventions.
 hardware_nvme_multipath: false
 
+# Select whether or not to load newer CPU microcode mitigations at runtime
+# when possible (and when it is newer than what the BIOS firmware provides).
+hardware_cpu_load_microcode: True
```
comay, line 275: Nit - s/True/true/ as lower-cased names are used elsewhere.

But more generally, rather than a boolean should this be just the type such as `auto`, `no`, `early` so someone can customize it? Even if it is a boolean, is there a reason to set it below to `early` as it seems `auto` is the default?

tj90241: doh, I thought I had fixed that `True` already!

Two considerations why I don't think we should allow for a string here:
- Using a string with dynamic implications that may be hard to support for other microarchitectures going forward.
- Non-early microcode loading has subtle knock-on effects that a user might not know about if they assume this only has `yes` or `no`.  `early` was likely introduced as a late option for [this PR](https://lwn.net/Articles/530346/), and I don't think there is any reason why you would not want to do it...

tj90241: An additional point - the package/code default is `early`, suggesting that `early` is a very usable default that is likely not known to break any systems.

tj90241:
Verification performed:
- `amd64-microcode` is configured only on AMD systems, `intel-microcode` is configured only on Intel systems
- On an Intel platform, I verified that the microcode is updated according to the direction of the flag.  I could not find an AMD system with microcode was sufficiently out of date to provably demonstrate that it works, but the config is simple and similar enough that I think it's OK.
```
# flag left at default, configure-system run, and system rebooted:
$ sudo dmesg -T | grep -i micro
[Thu Sep  3 15:36:56 2020] microcode: microcode updated early to revision 0x2006906, date = 2020-04-24
[Thu Sep  3 15:37:01 2020] microcode: sig=0x50654, pf=0x80, revision=0x2006906
[Thu Sep  3 15:37:01 2020] microcode: Microcode Update Driver: v2.2.

$ grep microcode /proc/cpuinfo  | head -n1
microcode	: 0x2006906
```

```
# flag set to disable, configure-system run, and system rebooted:
$ sudo dmesg -T | grep -i micro
[Thu Sep  3 15:55:06 2020] microcode: sig=0x50654, pf=0x80, revision=0x2000060
[Thu Sep  3 15:55:06 2020] microcode: Microcode Update Driver: v2.2.

$ grep microcode /proc/cpuinfo  | head -n1
microcode	: 0x2000060
```

ansible/playbooks/roles/common/defaults/main/base.yml
comay: Okay, if the package/codedefault is `early` than I'm fine with having that set. And also OK with leaving this as a boolean given your point although it's not clear to me how stable this sort of thing is namespace-wise with the Linux kernel.

ansible/playbooks/roles/common/templates/amd64-microcode/default.j2
```diff
@@ -0,0 +1,4 @@
+###############################################################################
+#                         generated by ansible
+###############################################################################
+AMD64UCODE_INITRAMFS={{ 'early' if hardware_cpu_load_microcode else 'no' }}
```
comay, line 4: For readability, can you please include the file as-is from the `amd64-microcode` package here, templated as needed with the "generated" comment at the top?

ansible/playbooks/roles/common/templates/intel-microcode/default.j2
```diff
@@ -0,0 +1,6 @@
+###############################################################################
+#                         generated by ansible
+###############################################################################
+IUCODE_TOOL_INITRAMFS={{ 'early' if hardware_cpu_load_microcode else 'no' }}
+IUCODE_TOOL_SCANCPUS=yes
+#IUCODE_TOOL_EXTRA_OPTIONS=""
```
comay, line 6: Ditto but from the `intel-microcode` package?

ansible/playbooks/roles/common/templates/amd64-microcode/default.j2
tj90241: yep, will do that for this and the AMD file in a sec.

comay (approved):

PR #1718, opened 2020-09-08
- Remove packages installed only because of apt recommends by using

mihalis68:
--no-install-recommends everywhere
- explicitly install python-setuptools instead of getting via recommends
- remove kexec recipe from the core chef "node" role
- remove kexec.rb chef recipe

- move kexec configuration to ansible too

**Describe your changes**

**Testing performed**

**Additional context**


comay (commented):
Chris, is chef/cookbooks/bcpc/files/default/kexec/default being removed? I didn't see it marked explicitly removed or it seems "git mv"ed to ansible/playbooks/roles/common/templates/kexec/default.j2 but perhaps that's not properly recorded by GitHub.

mihalis68:
> Chris, is chef/cookbooks/bcpc/files/default/kexec/default being removed? I didn't see it marked explicitly removed or it seems "git mv"ed to ansible/playbooks/roles/common/templates/kexec/default.j2 but perhaps that's not properly recorded by GitHub.

Good point. I've removed it!

mihalis68:
The impact of removing the extra packages obtained from recommended is shown in this gist which has before, after and a diff https://gist.github.com/mihalis68/0cfaebef3633b383a584cbd46be1ef9c

comay (commented):
There are a set of packages that aren't covered here - namely those installed via Ansible and ad-hoc `apt-get` installs in scripts.

ansible/playbooks/roles/chef-node/tasks/chef-node.yml                   chef
ansible/playbooks/roles/chef-server/tasks/chef-server.yml               chef-server-core
ansible/playbooks/roles/chef-workstation/tasks/chef-workstation.yml     chef
ansible/playbooks/roles/common/tasks/configure-bgp.yml                  bird2
ansible/playbooks/roles/common/tasks/configure-chrony.yml               chrony
ansible/playbooks/roles/common/tasks/configure-local-proxy.yml          tinyproxy
ansible/playbooks/roles/common/tasks/configure-resolver.yml             unbound
ansible/playbooks/roles/common/tasks/configure-system.yml               linux-*
ansible/playbooks/roles/common/tasks/configure-system.yml               kexec-tools
ansible/playbooks/roles/common/tasks/configure-system.yml               amd64-microcode
ansible/playbooks/roles/common/tasks/configure-system.yml               intel-microcode
virtual/network/provisioner.sh                                          lldpd
virtual/network/provisioner.sh                                          traceroute
virtual/network/provisioner.sh                                          bird2
virtual/network/provisioner.sh                                          iptables-persistent
virtual/provisioners/finish.sh                                          linux-*

Perhaps a more straightforward solution here would be to make this the default by setting

APT::Get::Install-Recommends "false";

in /etc/apt/apt.conf (thanks to @desrod for the line)?

tj90241:
@comay 👏 that is a MUCH better solution!  Will give it a spin.

The `python-setuptools` addition is because we now are not getting it for free as a recommended-package, and it's needed to run `setup.py` for `etcd3gw`.  We can possibly install `setuptools` in the `venv` there (and elsewhere?) if that's desirable.  (this is unfortunately another problem that would be solved by consuming this package directly from Canonical).

comay:
Looks like it's the latter name, at least according to

```
$ `apt-config dump APT::Install-Recommends`
APT::Install-Recommends "1";
```

so setting it to "0" should do the trick.

comay:
> The `python-setuptools` addition is because we now are not getting it for free as a recommended-package, and it's needed to run `setup.py` for `etcd3gw`. We can possibly install `setuptools` in the `venv` there (and elsewhere?) if that's desirable. (this is unfortunately another problem that would be solved by consuming this package directly from Canonical).

👍🏻

mihalis68:
I think futurist is something we need that we don't get without installing the recommended packages i.e. a hidden dependency. However I think th suggestion of setting apt:;install-recommends is much cleaner. Perhaps we should nuke this PR and start again clean? I have no problem with that. I can take it on again, but it seems @tj90241 may already be doing so.

comay:
It's not `python-futurist` but rather the addition of `python-setuptools` but Tyler explained why that was necessary.

tj90241:
I haven't started yet, @mihalis68 , if you are interested.

mihalis68:
yep I'll take this one again. can you stop the race condition I showed you?

tj90241:
yeah, I'll take a crack at that in a second.

mihalis68:
Replaced by the superior (much smaller) https://github.com/bloomberg/chef-bcpc/pull/1841

PR #1720, opened 2020-10-06
Add a backend optimization var to cinder.conf

tj90241:
This PR allows for us to control the value of `cinder`'s
`backend_native_threads_pool_size` via Chef attributes.

### DIFF
--- a/chef/cookbooks/bcpc/attributes/cinder.rb
+++ b/chef/cookbooks/bcpc/attributes/cinder.rb
@@ -7,6 +7,7 @@ default['bcpc']['cinder']['db']['dbname'] = 'cinder'
 default['bcpc']['cinder']['debug'] = false
 default['bcpc']['cinder']['workers'] = nil
 default['bcpc']['cinder']['allow_az_fallback'] = true
+default['bcpc']['cinder']['backend_native_threads_pool_size'] = nil
 default['bcpc']['cinder']['rbd_flatten_volume_from_snapshot'] = true
 default['bcpc']['cinder']['rbd_max_clone_depth'] = 5
 default['bcpc']['cinder']['database']['max_overflow'] = 10
--- a/chef/cookbooks/bcpc/recipes/cinder.rb
+++ b/chef/cookbooks/bcpc/recipes/cinder.rb
@@ -347,8 +347,9 @@ cinder_config.backends.each do |backend|
   end
 end
 
-service 'cinder-volume' do
-  action :start
+execute 'make sure cinder-volume comes up' do # ~FC004
+  action :nothing
   retries 30
+  command 'systemctl start cinder-volume'
   not_if 'systemctl status cinder-volume'
 end
--- a/chef/cookbooks/bcpc/templates/default/cinder/cinder.conf.erb
+++ b/chef/cookbooks/bcpc/templates/default/cinder/cinder.conf.erb
@@ -22,6 +22,11 @@ enable_v3_api = true
 scheduler_default_filters = <%= @scheduler_default_filters.join(',') %>
 <% end %>
 
+<% if !node['bcpc']['cinder']['backend_native_threads_pool_size'].nil? %>
+[backend_defaults]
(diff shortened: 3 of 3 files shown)


tj90241:
Testing to see if the resource in `ab12357` is still necessary or not.

chef/cookbooks/bcpc/attributes/cinder.rb
```diff
@@ -7,6 +7,7 @@
 default['bcpc']['cinder']['debug'] = false
 default['bcpc']['cinder']['workers'] = nil
 default['bcpc']['cinder']['allow_az_fallback'] = true
+default['bcpc']['cinder']['backend_native_threads_pool_size'] = 20
```
justinjpacheco, line 10: should we make the default `nil`?

chef/cookbooks/bcpc/templates/default/cinder/cinder.conf.erb
```diff
@@ -22,6 +22,9 @@ enable_v3_api = true
 scheduler_default_filters = <%= @scheduler_default_filters.join(',') %>
 <% end %>
 
+[backend_defaults]
```
justinjpacheco, line 25: if we agree with making the default value `nil` then should we make this section conditional based on that value?

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -347,8 +347,9 @@
   end
 end
 
-service 'cinder-volume' do
-  action :start
+execute "make sure cinder-volume comes up" do
```
comay, line 350: Can you explain the change here please?

tj90241: See the commit comments:
https://github.com/bloomberg/chef-bcpc/pull/1720/commits/ab123578779dfe23df1a1baec0d1203ea6c169df

We could also likely change this to a service resource and drop the `not_if` - just happened to go with the former.

chef/cookbooks/bcpc/templates/default/cinder/cinder.conf.erb
tj90241: I was thinking about doing that when writing this -- I had at first opted not to, as I thought the pattern we had typically used is to copy the current default value into the attributes (instead of eliding the configuration, and letting OpenStack maintain the default).

e.g., for `memcached` we specify `-m 64` even though that's a daemon-default.

I am happy to make this change though.

chef/cookbooks/bcpc/attributes/cinder.rb
tj90241: see below

chef/cookbooks/bcpc/templates/default/cinder/cinder.conf.erb
comay: I too have wondered about this in the past - whatever we do, we should be consistent and perhaps write down how such parameters be handled and open tickets to update the places that don't conform to whatever convention we use.

justinjpacheco: for memcached we did do this, that's true, but [here](https://github.com/bloomberg/chef-bcpc/pull/1720/files#diff-93a0bda3f49972303b9a45ad672b5e77R8) you can see that we have it set to `nil` by default for the service workers.

chef/cookbooks/bcpc/recipes/cinder.rb
comay: I guess it seems there's either a Chef `service` resource bug here (not honoring the `notifies :restart, 'service[cinder-volume]', :immediately`) or if it's indeed being blocked by the `not_if`, then it seems removing that makes more sense rather than emulating the `service` resource.

Does it matter at all with Chef that the service definition appears at the end of the file rather than being define prior to first reference (such as in the `notifies :restart, 'service[cinder-volume]', :immediately`?)

comay (approved):

justinjpacheco (approved):

justinjpacheco (approved):

PR #1726, opened 2020-10-16
Refactor/clean-up some libvirt build tooling

tj90241:
`libvirt`-based builds have drifted from `virtualbox` builds.
This PR aims to bring the `libvirt` builds up to parity.


virtual/network/Vagrantfile.libvirt
```diff
@@ -21,7 +21,7 @@ base_box_version = '202005.21.0'
 def setup_proxy(node)
   http_proxy = ENV['http_proxy'] || ''
   https_proxy = ENV['https_proxy'] || ''
-  return if http_proxy == '' and https_proxy == ''
+  return unless http_proxy != ''
```
tj90241, line 24: Making various random changes to keep behavior of `Vagrantfile.libvirt` consistent with whatever `Vagrantfile` is doing.

virtual/network/Vagrantfile.libvirt
```diff
@@ -60,13 +61,18 @@ Vagrant.configure(2) do |config|
...
       lv.cpu_mode = 'host-passthrough'
       config.vm.synced_folder './', '/vagrant', type: '9p'
+      if ENV['BCC_LIBVIRT_KVM_HUGEPAGES'] == 'true'
+        lv.memorybacking :hugepages
+      end
       #mount_apt_cache(config)
     end
     lv.cpus = 1
-    lv.memory = 512
+    lv.memory = 1024
```
tj90241, line 70: Need 1024mb to support hugepages under all scenarios.

virtual/Vagrantfile.libvirt
```diff
@@ -112,16 +131,18 @@ Vagrant.configure('2') do |config|
         args = {}.tap do |a|
           a[:ip] = t_iface['ip'].split('/')[0]
           a[:mac] = t_iface['mac'].delete(':')
-          a[:libvirt__network_name] = t_iface['neighbor']['name']
+          a[:libvirt__network_name] = Util.vbox_name(t_iface['neighbor']['name'])
           a[:libvirt__dhcp_enabled] = false
-          a[:libvirt__forward_mode] = 'none'
+          a[:libvirt__forward_mode] = 'veryisolated'
```
tj90241, line 136: If `forward_mode` is not `veryisolated`, libvirt with bark about:
`Address x.x.x.x does not match with network name`

... when standing up VMs.  This topic was contentiously debated in: https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1053

but regardless, changing it makes it have the same behavior that Virtualbox does.

tj90241:
@justinjpacheco:
- This PR includes the deterministic hash in the domain name when `BCC_ENABLE_LIBVIRT_PREFIX` is `true` so multiple libvirt-based builds can be done on the same host.  I hate making more envvars, but `BCC_ENABLE_VBOX_SUFFIX` is a no-go as it (a) does not work for `vagrant-libvirt`, which only allows the _prefix_ of the domain to be customized and (b) does not make sense given the `VBOX` clearly in the envvar name.  We could change Virtualbox to prefix the hash on the VM name to match libvirt's style, and rename the envvar to BCC_ENABLE_VM_PREFIX?  It would necessitate another envvar name change though... thoughts?

There are still two features libvirt builds lack, hopefully TBD in later PRs:
- The `network/Vagrantfile.libvirt` is still missing the bits to standup the full set of network ToR/spine VMs and only builds the network VM.  I will fix this in a future PR as an additional feature of sorts as this PR already has enough in it as it is.
- The apt cache is not mounted inside the VMs.  I have to investigate if this is possible in libvirt/if this requires Virtualbox-specific features.

Net-new items added in this PR:
- Very slight documentation touch up in README.md
- Add envvar `BCC_LIBVIRT_KVM_HUGEPAGES` and use hugepages for VM backing when enabled
- Bump libvirt's network VM to 1G RAM so 1G hugepages can be leveraged on x86(_64).

Testing performed:
- Build 2x libvirt-based 3h3w with hugepages on the same host using `make destroy create all`.

justinjpacheco (approved):

PR #1729, opened 2020-11-04
Add a RabbitMQ node role

tj90241:
Add a 'rmqnode' role, thereby enabling operators to build
"headnodes" which only run RabbitMQ, and no other headnode
services.

### DIFF
--- a/ansible/playbooks/roles/common/defaults/main/chef.yml
+++ b/ansible/playbooks/roles/common/defaults/main/chef.yml
@@ -61,7 +61,6 @@ chef_roles:
       - role[node]
       - recipe[bcpc::haproxy]
       - recipe[bcpc::etcd-member]
-      - recipe[bcpc::rabbitmq]
       - recipe[bcpc::memcached]
       - recipe[bcpc::unbound]
       - recipe[bcpc::consul]
@@ -80,6 +79,14 @@ chef_roles:
       - recipe[bcpc::os-quota]
       - recipe[bcpc::flavors]
       - recipe[bcpc::watcher]
+
+  - name: rmqnode
+    description: cloud infrastructure MQ services
+    json_class: Chef::Role
+    chef_type: role
+    run_list:
+      - role[node]
+      - recipe[bcpc::rabbitmq]
 
   - name: node
     description: common role for all bcpc nodes
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ export playbooks = ansible/playbooks
 export ANSIBLE_CONFIG = ansible/ansible.cfg
 
 headnodes = $$(ansible headnodes -i ${inventory} --list | tail -n +2 | wc -l)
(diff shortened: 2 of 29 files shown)


ansible/playbooks/roles/common/defaults/main/chef.yml
```diff
@@ -84,6 +83,15 @@ chef_roles:
       - recipe[bcpc::flavors]
       - recipe[bcpc::watcher]
 
+  - name: rmqnode
+    description: cloud infrastructure MQ services
+    json_class: Chef::Role
+    chef_type: role
+    run_list:
+      - role[node]
+      - recipe[bcpc::calico-felix]
```
comay, line 92: It seems that `recipe[bcpc::etcd-proxy]` needs to be added here prior to including the recipe for `calico-felix`. That said, I don't know how that might interfere with the inclusion of `recipe[bcpc::etcd-member]` by headnodes.

comay (commented):
Just starting to look at this - it mostly makes sense although it will take a bit longer to digest the changes under `virtual/network`.

One nit - although I don't expect we would ever use some other AMQP implementation, it seems the prefix `rmq` might be better written as `mq`. That said, I'm fine with leaving that as-is given how unlikely it is we switch from RabbitMQ.

tj90241:
The `rmqnode` => `mqnode` should be easy enough to change... I agree it is better to remain neutral to these details from within the inventory.

mihalis68:
Not worth it. Even openstack is specifically built around rabbit. I doubt there are working bindings to anything else. Let's not get caught in the trap of false/premature generality

ansible/playbooks/roles/common/defaults/main/chef.yml
tj90241: Very good catching... working on a general solution to this problem.

justinjpacheco: i think this is similar to the issue with nodes that are worknodes + storagenodes (ie: in bve 1h1w). you might require some logic to return early based on the situation.

tj90241: General solution: https://github.com/bloomberg/chef-bcpc/pull/1739

It's a bit different than the direction we normally go, and I found the best way to get out of an a-b-a problem (`calico-felix` from `rmqnode` requires `etcd`, but many of the services from `headnode` require `rabbitmq`) was to use a 'bootstrap' playbook that goes around and solves some prereqs.  It should also help build time by again lifting more things up to `role[node]`, which is now applied in parallel.

tj90241:
Rebased this on top of #1739 for now; once that's reviewed/merged, I'll force-push this branch less those changes.

tj90241:
Tested old/new topologies (with dedicated rabbit + 3 rack, without dedicated rabbit + 1 rack)... no regression observed.

comay (approved):

comay (approved):

PR #1730, opened 2020-11-05
Add props to configure neutron's DB pool size

mboruta1:
Added properties to allow users to configure neutron's backend DB pool
size. The default values of these properties have also been increased
from 5 (pool size) and 50 (overflow) to 50 and 150, respectively.

**Testing performed**
Tested on my local 3h3w BVE set up; tempest tests no longer fail due to the neutron db backend running out of connections.

### DIFF
--- a/chef/cookbooks/bcpc/attributes/neutron.rb
+++ b/chef/cookbooks/bcpc/attributes/neutron.rb
@@ -35,3 +35,7 @@ default['bcpc']['neutron']['quota']['default']['quota_floatingip'] = 0
 
 # per-project quota settings
 default['bcpc']['neutron']['quota']['project']['admin']['rbac-policies'] = -1
+
+# database connection pool settings
+default['bcpc']['neutron']['db']['max_pool_size'] = 64
+default['bcpc']['neutron']['db']['max_overflow'] = 128
--- a/chef/cookbooks/bcpc/templates/default/neutron/neutron.conf.erb
+++ b/chef/cookbooks/bcpc/templates/default/neutron/neutron.conf.erb
@@ -28,6 +28,8 @@ etcd_key_file = <%= node['bcpc']['etcd']['client-rw']['key']['filepath'] %>
 
 [database]
 connection = mysql+pymysql://<%= "#{@db['username']}:#{@db['password']}@#{node['bcpc']['mysql']['host']}/#{@db['dbname']}" %>
+max_pool_size = <%= node['bcpc']['neutron']['db']['max_pool_size'] %>
+max_overflow = <%= node['bcpc']['neutron']['db']['max_overflow'] %>
 
 [keystone_authtoken]
 auth_uri = <%= "https://#{node['bcpc']['cloud']['fqdn']}:5000" %>


chef/cookbooks/bcpc/attributes/neutron.rb
```diff
@@ -35,3 +35,7 @@
 
 # per-project quota settings
 default['bcpc']['neutron']['quota']['project']['admin']['rbac-policies'] = -1
+
```
justinjpacheco, line 38: the attributes file for neutron already has a `db` section (see line 5 of this file). could we move these parameters under `default['bcpc']['neutron']['db']` please?

chef/cookbooks/bcpc/attributes/neutron.rb
```diff
@@ -35,3 +35,7 @@
...
+# database connection pool settings
+default['bcpc']['neutron']['database']['max_pool_size'] = 50
+default['bcpc']['neutron']['database']['max_overflow'] = 150
```
comay, line 41: When we set "default" values for the upstream project, we have a number of choices. One is to use the default "compiled-in" values from the upstream/vendor. Another is to choose something different if 1) it generally makes sense and 2) does no harm with the virtual builds.

With that in mind, where are these values coming from? They're not the upstream defaults AFAIK (5 & 50, respectively) but are they reasonable for a typical virtual build.

Also, what values would you consider for real clusters?

chef/cookbooks/bcpc/attributes/neutron.rb
mboruta1: Ahh, sorry, missed that. Made the changes and pushed out new commit.

chef/cookbooks/bcpc/attributes/neutron.rb
mboruta1: The default values (5/50) are not sufficient in order to run tempest tests on a 1h1w BVE node, and the neutron services on both pw and ob have run out of backend db connections over the past 30 days (see [this humio query](https://humio.prod.bloomberg.com/bcc/search?query=%22http%3A%2F%2Fsqlalche.me%2Fe%2F3o7r%22%0A%7C%20%40source%20%3D%20%22%2Fvar%2Flog%2Fneutron%2Fneutron-server.log%22&live=false&start=30d&fullscreen=false)). Since these additional connections are used only on an as-needed basis this change should have no adverse effects on either BVE environments or on real clusters. On real clusters db connection usage peaks at around 4k out of 8k, so even if these extra connections were all used simultaneously on all 3 head nodes we would just end up in the same position as we are now, neutron service cannot open new connections while other services (i.e. nova) continue to connect as usual. 

As for where the values are coming from, I thought that a 4x increase in the total number of connections should give us plenty of breathing room while still keeping us below the 8k limit (which we can easily raise later if needed). It also resolved the error when running tempest tests. I think these values should be fine to use on real clusters as well.

More details can be found in the [issue](https://jira.prod.bloomberg.com/browse/RDTIBCC-2718) that spawned this change, as well as a related [issue](https://jira.prod.bloomberg.com/browse/RDTIBCC-2737) I posted regarding a near-cluster wide issue where many services maxed out their db connection pool.

comay (approved):

justinjpacheco (approved):

PR #1739, opened 2020-11-12
chef: Add etcd/calico-felix node's to run_list.

tj90241:
Every BCPC node runs `calico-felix` for policy enforcement,
which itself requires an etcd connection for policy to be
obtained.  Headnodes obtain an etcd connection by virtue
that they are the etcd members.  All othernodes require an
`etcd-proxy` service.
    
Add a recipe which installs `etcd-proxy` on all non-
headnodes, and amend the cluster bootstrapping process
so that prereqs for both `etcd-proxy` and `calico-felix`
come up very early.  Finally, add both `etcd-proxy` and
`calico-felix` to `role[node]`'s run_list.

### DIFF
--- a/ansible/playbooks/configure-common-node.yml
+++ b/ansible/playbooks/configure-common-node.yml
@@ -1,0 +1,31 @@
+# When we bring-up a new cluster, there are some circular dependencies that
+# present themselves (depending on how roles are broken out).  This playbook
+# attempts to resolve those dependencies before chef is used to converge
+# each node so that things go smoothly.
+#
+# The list of dependencies we resolve in configure-web-server [1] and here [2]:
+#   * [1] etcd requires bcpc::web-server (to obtain packages)
+#   * [2] etcd-proxy (non-headnodes) requires etcd (headnodes)
+#   * [2] calico-felix requires etcd-proxy or etcd-member
+#
+# In the future, we should move Consul, etcd, etc. out of Chef and into
+# Ansible so these dependencies can be resolved more elegantly.
+- hosts: headnodes
+  gather_facts: no
+  become: yes
+  order: inventory
+  tasks:
+    - name: provision the first etcd server
+      become: true
+      command: chef-client -o bcpc::etcd-member
+      run_once: true
+      changed_when: false
+
+- hosts: cloud
+  gather_facts: no
+  become: yes
+  tasks:
+    - name: run chef-client with role[node]'s run_list
+      become: true
+      command: chef-client -o role[node]
+      changed_when: false
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ all : \
 	configure-chef-workstation \
 	configure-chef-nodes \
 	configure-web-server \
+	configure-common-node \
 	run-chef-client \
 	reweight-ceph-osds \
 	add-cloud-images \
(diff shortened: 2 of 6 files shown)


chef/cookbooks/bcpc/recipes/guarded-etcd-proxy.rb
```diff
@@ -0,0 +1,24 @@
+# Cookbook:: bcpc
```
justinjpacheco, line 1: why not just have etcd-proxy return early if on a headnode? ie:

return if headnode?

tj90241: Chef documentation seemed to suggest that `return` would stop execution of the _entire_ client run, though I admittedly never tried it after reading that.  I'll give it a shot.

comay (commented):
Is there a reason not to just have a single "etcd" recipe that has conditionals on whether it's a member or a proxy? That would make more sense to me as in all node types, you are running an `etcd` daemon.

Can you also add some comments as to why you're pulling out the provisioning of the first `etcd` server into the Ansible task? It's not clear to me why you're doing that but I suspect there's a reason concerning bootstraping things.

chef/cookbooks/bcpc/recipes/guarded-etcd-proxy.rb
tj90241: Worked as mentioned!  thanks

tj90241:
There is not too much similarity between the `etcd` recipes, and we decided to keep `bcpc::etcd-member`out of the `node[role]`'s picture in particular as it's a clustered service that's specific to the `headnodes`.

Added comments to the playbook to explain why we're doing things in this order.

For documentation here, we've been somewhat reliant on the order in which run lists are ordered in Chef (note for the bootstrap, `bcpc::web-server` was _before_ `bcpc::etcd-proxy` and `bcpc::calico-felix`.  This dependencies of this form up to Ansible playbook and allows these dependencies to be resolved more explicitly via targeted Chef runs.  In the future, as we step more towards Ansible, we can resolve those dependencies more naturally with fine-grained Ansible roles.

I tested bringing up various topologies up in a virtual environment and didn't observe any regressions with this.

tj90241:
The build failure can be ignored and will be fixed with a rebase when this is approved.

comay (approved):

comay (approved):

ansible/playbooks/configure-common-node.yml
```diff
@@ -0,0 +1,31 @@
...
+#   * [1] etcd requires bcpc::web-server (to obtain packages)
+#   * [2] etcd-proxy (non-headnodes) requires etcd (headnodes)
+#   * [2] calico-felix requires etcd-proxy or etcd-member
+#
+# In the future, we should move Consul, etcd, etc. out of Chef and into
+# Ansible so these dependencies can be resolved more elegantly.
+- hosts: headnodes
+  gather_facts: no
+  become: yes
+  order: inventory
```
justinjpacheco, line 16: isn't this the default?

tj90241: `order: inventory`, you mean?  Yes, I think so... I had just put it there to provide an additional hint alongside `run_once` to hint that it's ref'ing the first `headnode` that appears in the inventory order.  Should it be removed?

justinjpacheco: just confirming that this is the default behavior. if you feel it helps with readability then you should leave it in

justinjpacheco (approved):

comay (dismissed):

justinjpacheco (approved):

PR #1740, opened 2020-11-16
DoS protection: Phase 1

mboruta1:
**Describe your changes**
Operators of BCPC may now rate limit users based on:
- number of open TCP connections to the cluster leader
- TCP connection rate over a specified period of time
- API request rate over a specified period of time
- API Error rate over a specified period of time

Users exceeding these limits will receive either a 429 HTTP status code or their TCP connection will be automatically closed by the server.

Traffic originating from nodes within the cluster are exempt. Operators may also exempt an arbitrary number of CIDRs from rate limiting.

All of these changes occur in the haproxy component. When modifying the haproxy configuration (i.e. adding additional CIDRs to exempt from rate limiting, turning off rate limiting altogether) one only needs to reload the service to apply the changes. End users see no service disruption.

**Testing performed**
Tested locally on my bve by running Apache Benchmark (ab).


chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -117,7 +117,7 @@
     headnodes: headnodes(all: true),
     vip: node['bcpc']['cloud']['vip']
   )
-  notifies :restart, 'service[haproxy-cinder]', :immediately
+  notifies :reload, 'service[haproxy-cinder]', :immediately
```
mboruta1, line 120: Only need to use reload here since its an haproxy config change. Reload results in virtually no service disruption and should be used when we just need some config changes to be applied. See https://cbonte.github.io/haproxy-dconv/2.0/management.html#4 for more info.

chef/cookbooks/bcpc/templates/default/haproxy/qos-exempt.acl.erb
```diff
@@ -0,0 +1,9 @@
+<%= node['bcpc']['cloud']['vip'] %>
+<%= node['bcpc']['networking']['networks']['primary']['aggregate-cidr'] %>
+<% if node['bcpc'].attribute?('qos_exemptions') %>
```
justinjpacheco, line 3: you can reduce this by one block if you use `node['bcpc'].fetch('qos_exemptions',{})`

chef/cookbooks/bcpc/attributes/haproxy.rb
```diff
@@ -0,0 +1,29 @@
+###############################################################################
```
justinjpacheco, line 1: attributes/default.rb contains an haproxy section which i think should be removed and placed into this file

chef/cookbooks/bcpc/templates/default/haproxy/qos-exempt.acl.erb
justinjpacheco: is this key defined anywhere in this pr?

chef/cookbooks/bcpc/templates/default/heat/haproxy.cfg.erb
```diff
@@ -25,3 +27,5 @@ listen heat-api-cfn
 <% @headnodes.each do |n| -%>
   server <%= n['hostname'] %> <%= n['service_ip'] %>:8000 check inter 5s rise 1 fall 1
 <% end -%>
+
```
justinjpacheco, line 30: is there a reason why this is defined twice in the same file?

chef/cookbooks/bcpc/templates/default/haproxy/qos-exempt.acl.erb
mboruta1: No, not in this pr. It is designed to be completely optional, and is specified in my pr to our private repo.

chef/cookbooks/bcpc/templates/default/haproxy/haproxy.cfg.erb
```diff
@@ -20,7 +20,7 @@ defaults
   option dontlognull
   option redispatch
   retries 3
-  timeout http-request 10s
+  timeout http-request <%= node['bcpc']['haproxy']['qos']['http_request_timeout'] %>
```
peter-at, line 23: is 10s -> 5s (default) a big change?

chef/cookbooks/bcpc/templates/default/heat/haproxy.cfg.erb
mboruta1: Yes, the QoS template needs to be copied to each listen section. It was either this or creating a single frontend and defining how QoS is done once.

chef/cookbooks/bcpc/templates/default/haproxy/qos-exempt.acl.erb
justinjpacheco: and `qos_exemptions` is meant to be an array?

chef/cookbooks/bcpc/templates/default/haproxy/haproxy.cfg.erb
mboruta1: No its not but it was recommended on one of HAProxy's blogs: https://www.haproxy.com/blog/application-layer-ddos-attack-protection-with-haproxy/.

chef/cookbooks/bcpc/templates/default/haproxy/qos-exempt.acl.erb
mboruta1: Yes, containing an arbitrary number (including 0) of CIDRs and/or IPs.

justinjpacheco: i think this should read as
node['bcpc']['haproxy'].fetch('qos_exemptions',[]).each do |qe|
  qe
end

notice that qos_exemptions is under haproxy because this is an haproxy related config option. also, because this is an array, i don't see much value in checking if its empty or not, the for loop should just skip it (unless we're concerned about actual null values)

mboruta1: Ok, makes sense, will do!

chef/cookbooks/bcpc/templates/default/heat/haproxy.cfg.erb
justinjpacheco: ahhh, ok. the indentation on qos template is important then. ok, i understand

peter-at: instead of duplicate them for each listen section, is it possible to put in a default section?

mboruta1: @peter-at No, it is not, unfortunately. It was either this or a single frontend. I haven't looked too much into lua scripts so *maybe* they could help us here, but chef-wise it would still be the same amount of code. Something to look into in the future though.

chef/cookbooks/bcpc/attributes/haproxy.rb
```diff
@@ -0,0 +1,29 @@
...
+default['bcpc']['haproxy']['qos']['conn_rate_period'] = '120s'
+# The period across which the HTTP request rate of clients is measured
+default['bcpc']['haproxy']['qos']['http_req_rate_period'] = '120s'
+# The period across which the HTTP error response rate of clients is measured
+default['bcpc']['haproxy']['qos']['http_err_rate_period'] = '30s'
+
+# Maximum number of concurrent connections allowed per client
+default['bcpc']['haproxy']['qos']['conn_limit'] = 50
+# The number of TCP connections allowed per client within conn_rate_period
+default['bcpc']['haproxy']['qos']['conn_rate'] = 1200
```
peter-at, line 24: are these one size fit all?
keystones are in range of 25 reqps and neutron 100 reqps
will both be okay with these settings for the 120s period?

mboruta1: What we are doing here is setting a system-wide limit of 600 HTTP requests over a 120 second period per user. We have a stick table that keeps track of this data for each client based on source IP. The stick table needs to be updated in each listen section that routes requests to openstack services, and the logic that rate limits the given request needs to be present in each listen section as well.

One potential improvement is to apply different limits per endpoint (i.e. less for keystone, more for neutron), followed by the holy grail of per API type (i.e. does this API call mutate resources? Does it just retrieve data? Is it a known expensive operation? etc).

justinjpacheco:
it just occurred to me that this qos feature should be something that can be enabled and disabled much like our other optional features. perhaps something like node['bcpc']['haproxy']['qos']['enabled'] = true | false

mboruta1:
Ok makes sense. I added an option to enable/disable QoS.

chef/cookbooks/bcpc/attributes/haproxy.rb
```diff
@@ -0,0 +1,42 @@
...
+###############################################################################
+
+# Installation-related configuration options
+default['bcpc']['haproxy']['apt']['enabled'] = false
+default['bcpc']['haproxy']['apt']['url'] = 'http://ppa.launchpad.net/vbernat/haproxy-1.8/ubuntu'
+
+# QoS protection-related configuration options
+
+# Whether QoS is enabled or not.
+default['bcpc']['haproxy']['qos']['enabled'] = true
```
justinjpacheco, line 12: we usually leave optional features disabled by default and enable them when we are ready to support them

justinjpacheco (dismissed):

chef/cookbooks/bcpc/attributes/haproxy.rb
```diff
@@ -0,0 +1,42 @@
...
+# QoS protection-related configuration options
+
+# Whether QoS is enabled or not.
+default['bcpc']['haproxy']['qos']['enabled'] = false
+
+# The amount of time to wait for HTTP headers to be sent
+default['bcpc']['haproxy']['qos']['http_request_timeout'] = '5s'
+
+# The maximum number of entries in the stick table
+default['bcpc']['haproxy']['qos']['max_entries'] = '1m'
```
comay, line 18: This all looks more or less in good shape but for readability, can you please separate each of these commented attributes with a blank line?

mboruta1: Ok, wanted them grouped according to the type of attributes they are but I can see why it may be a bit hard to read. Made the change and pushed it out.

justinjpacheco (approved):

PR #1741, opened 2020-11-16
chef: Raise `oslo`'s database connection thresholds.

tj90241:
We recently raised `neutron`'s `oslo.db` tunable,
`max_overflow`, from 10 to 128.  The default in upstream
from a pre-Rocky era was 10, and it's now 50.  Thus, we're
likely way too conservative on some of these settings --
bump them up to 128 to be slightly more aggressive than
upstream is so we can avoid overrides.
    
Also raise `max_pool_size` along with `max_overflow`, in
the same manner that it was raised for `neutron`.  Note
that this was not a default setting change upstream;  just
a change that we are making locally.


tj90241:
ref: https://docs.openstack.org/oslo.db/rocky/reference/opts.html

justinjpacheco:
is there any reason to not bump these up to something that won't require an override?

tj90241:
@justinjpacheco raised them all to 128 to match the recent `neutron` PR.

justinjpacheco (approved):

PR #1745, opened 2020-12-02
Optionally support a multi-fabric, leaf/spine/super-spine virtual network

comay:
Introduce optional support for multiple pods & "Super Spine" routers
Cleanup existing BIRD configuration files
Include support for up to seven host BGP peerings in each TOR
Change management networks to a /27 to support additional hosts

This has been regression tested with `1h1w` & `3h6w` configurations and with a multiple pod configuration in a `1h5w` configuration with the following virtual hosts:

Pod 1
=====
r1n0 (bootstrap)
r2n1 (headnode)
r3n2 (worknode)

Pod 2
=====
r5n2 (worknode)
r6n2 (worknode)
r7n2 (worknode)

The most visible change even with a single pod configuration are the hostnames of the TOR and spine routers.


virtual/network/Vagrantfile
```diff
@@ -27,9 +27,34 @@ def setup_proxy(node)
...
+FABRIC_ID = "bcpc-"
+
+if ENV.key?('BCC_POD_COUNT')
+  PODS = ENV['BCC_POD_COUNT'].to_i
+else
+  PODS = 1
+end
+
+# We assume that the number of planes & spines per pod are the same.
+PLANES = 2
```
tj90241, line 39: A future enhancement, perhaps - almost looks like we could use a `network.yaml` at some point, similar to what we have today a la `hardware.yaml` and `topology.yaml` for configuring the parameters of the VM and virtual rack/etc. layout, respectively.  Doing so would abolish the use of the env vars and allow developers to have multiple sizes of network builds on the same host without swizzling env vars.

comay: Indeed - what I've suggested is something along the lines of `datacenter.yaml` that defines the network, the racks, the hosts, etc - the whole kit and caboodle. From that, we would render all of the various data files such as the ones for `bird`, `netplan`, etc.

virtual/network/Vagrantfile
```diff
@@ -50,67 +75,110 @@ Vagrant.configure(2) do |config|
         virtualbox__intnet: Util.vbox_name('management3'),
         auto_config: false)
       setup_proxy(node)
-      node.vm.provision 'shell', path: 'provisioner.sh', args: 'network'
+      node.vm.provision 'shell', path: 'provisioner.sh', args: '-E network'
```
tj90241, line 78: Can you please reflect this change in `Vagrantfile.libvirt` as well?
https://github.com/bloomberg/chef-bcpc/blob/master/virtual/network/Vagrantfile.libvirt#L90

tj90241 (approved):

virtual/network/Vagrantfile
comay: Done.

tj90241 (dismissed):

justinjpacheco (approved):

tj90241 (approved):

PR #1749, opened 2020-12-04
Define a stub node role and stubnodes group.

tj90241:
Define a role/group for "stub" nodes, which are nodes
that are to receive basic system, network, and policy
configuration, but no additional cloud infrastructure
services.


tj90241:
A stub node can be defined like so:
(note: `hardware_profile` denotes the characteristics of the VM, so ignore the fact that it is `headnode` here, as it is not actually describing a `headnode` in the sense of its software composition).
```
  - host: r1n3
    group: stubnodes
    hardware_profile: headnode
    host_vars:
      bgp:
        asn: 4200858801
      interfaces:
        service:
          ip: 10.65.0.4
        transit:
          - ip: 10.121.84.5/28
            mac: '08:00:27:00:00:16'
            neighbor:
              ip: 10.121.84.1
              asn: 4200858701
              name: management1
          - ip: 10.121.88.5/28
            mac: '08:00:27:00:00:17'
            neighbor:
              ip: 10.121.88.1
              asn: 4200858702
              name: management4
      run_list:
        - role[stub]
```

comay (dismissed):

tj90241:
Changed the nomenclature of the Chef role from `stub` to `stubnode`, as the rest of our Chef roles are suffixed with `node`.  Without this being the case for `stubnode`, it looked very out of place in other portions of the code.  Maybe something to discuss before we add too many more roles?

mboruta1 (approved):
Looks good to me.

comay (approved):

PR #1756, opened 2020-12-10
use unbound over systemd-resolved

justinjpacheco:
- configure unbound to be the primary dns resolver
  - disable/mask systemd-resolved


ansible/playbooks/roles/common/tasks/configure-networking.yml
```diff
@@ -12,27 +12,32 @@
...
+- import_tasks: configure-unbound.yml
+  tags: [configure-unbound]
+
+- import_tasks: configure-local-proxy.yml
+  tags: [configure-local-proxy]
+
 - import_tasks: configure-chrony.yml
   tags: [configure-chrony]
 
+- import_tasks: configure-ubuntu-apt-sources.yml
```
comay, line 36: I know we discussed this reordering a few months back but should the setting of the apt sources occur earlier in the task list? I ask just in case one of the earlier tasks require installation from one of these sources? Perhaps before at least Chrony since I'm guessing Tinyproxy may be needed to do the `apt-get update` in the apt source task.

ansible/playbooks/roles/common/tasks/configure-unbound.yml
```diff
@@ -0,0 +1,36 @@
+- name: sysctl allow non-local ip bind
+  sysctl:
+    name: net.ipv4.ip_nonlocal_bind
```
comay, line 3: Is this necessary to bind to the VIP address unconditionally?

ansible/playbooks/roles/common/templates/systemd/resolved.conf.j2
```diff
@@ -3,9 +3,9 @@
 ###############################################################################
```
comay, line 3: As this file is no longer a template but I know we need to redeliver it for upgrade purposes, should it be moved to `ansible/playbooks/roles/common/files/systemd/resolved.conf`? Also, the "compiled-in" default of `Cache` is "yes", so you should change that. Here's the actual default version of this file if you want to just redeliver that

<cut here>
#  This file is part of systemd.
#
#  systemd is free software; you can redistribute it and/or modify it
#  under the terms of the GNU Lesser General Public License as published by
#  the Free Software Foundation; either version 2.1 of the License, or
#  (at your option) any later version.
#
# Entries in this file show the compile time defaults.
# You can change settings by editing this file.
# Defaults can be restored by simply deleting this file.
#
# See resolved.conf(5) for details

[Resolve]
#DNS=
#FallbackDNS=
#Domains=
#LLMNR=no
#MulticastDNS=no
#DNSSEC=no
#Cache=yes
#DNSStubListener=yes
<cut here>

ansible/playbooks/roles/common/templates/unbound/default.j2
```diff
@@ -0,0 +1,5 @@
+###############################################################################
```
comay, line 1: Again, should this be under `ansible/playbooks/roles/common/files/unbound` and as the file is static, should it include the header?

chef/cookbooks/bcpc/templates/default/unbound/local-zones.conf.erb
```diff
@@ -0,0 +1,9 @@
+server:
+<% @local_zones.each do |zone,ns| %>
```
comay, line 2: s/,/, /

chef/cookbooks/bcpc/templates/default/unbound/local-zones.conf.erb
```diff
@@ -0,0 +1,9 @@
...
+  local-zone: <%= zone %> transparent
+<% end %>
+<% @local_zones.each do |zone,ns| %>
```
comay, line 5: Ditto.

chef/cookbooks/bcpc/recipes/unbound.rb
```diff
@@ -15,48 +15,32 @@
 # See the License for the specific language governing permissions and
```
comay, line 15: So I'm guessing I'm missing something fundamental here - it seems that Unbound is being installed & configured by both Ansible & Chef?

ansible/playbooks/roles/common/tasks/configure-unbound.yml
justinjpacheco: this just allows unbound to bind to an address that it might not have, in this case, the vip. every node in the cluster points their unbound to the vip as their primary dns resolver.

chef/cookbooks/bcpc/recipes/unbound.rb
justinjpacheco: it is only being installed by ansible, chef is simply being used to drop the .IN-ADDR configuration so that the headnodes know how to resolve reverse ips.

this .IN-ADDR configuration file requires the use of the `cidr_to_reverse_zones(IPAddress(subnet['allocation']))` function that is only defined in chef

ansible/playbooks/roles/common/templates/unbound/default.j2
justinjpacheco: i can move it under files, but shouldn't files dropped by automation have the header or are we dropping this concept?

ansible/playbooks/roles/common/tasks/configure-networking.yml
justinjpacheco: what did you have in mind?

ansible/playbooks/roles/common/templates/unbound/default.j2
comay: I thought the header was only for files that were generated by templates rather than just straight as-is file copies but perhaps I've been mistaken all this time?

chef/cookbooks/bcpc/recipes/unbound.rb
comay: Ah, so Ansible is actually responsible for both installation *and* the the majority of the configuration - Chef is just for the delivery of the IN-ADDR pieces in `local-zones.conf`. Sweet.

ansible/playbooks/roles/common/templates/unbound/default.j2
justinjpacheco: well, the way i see it, this file will have any local changes reverted on the next run of automation, so the header is meant to let operators know that the contents of this file are managed somewhere else and that their changes aren't permanent.

ansible/playbooks/roles/common/tasks/configure-system-name-resolution.yml
```diff
@@ -0,0 +1,60 @@
...
+
+- name: install /etc/resolv.conf
+  template:
+    src: etc/resolv.conf.j2
+    dest: /etc/resolv.conf
+
+###############################################################################
+# configure systemd-resolved
+#
+# the following section will stop,disable and mask the systemd-resolved service
```
comay, line 46: s/,/, /

comay (dismissed):

ansible/playbooks/roles/common/tasks/configure-system-name-resolution.yml
```diff
@@ -0,0 +1,66 @@
+###############################################################################
+# configure unbound
+#
+# the following section will configure unbound to be the primary name resolver
+###############################################################################
+
+- name: sysctl allow non-local ip bind
```
tj90241, line 7: Should be limit this `sysctl` to just `headnodes`?  Do other nodes need to bind to non-local addresses?  If not, getting "Cannot assign requested address" errors may be helpful.

justinjpacheco: good point, i'll limit to just headnodes

ansible/playbooks/roles/common/templates/unbound/server.conf.j2
```diff
@@ -0,0 +1,40 @@
+###############################################################################
+# managed by ansible
+###############################################################################
+server:
+    access-control: 0.0.0.0/0 allow
+    chroot: ""
+    directory: /etc/unbound
+    do-ip4: yes
+    do-ip6: no
```
tj90241, line 9: `no` -> `'yes' if system_enable_ipv6 else 'no'`?
https://github.com/bloomberg/chef-bcpc/blob/188460c93765ef944e1359d43baf5e588b30fbba/ansible/playbooks/roles/common/defaults/main/base.yml#L290

We haven't yet been totally consistent with use of this template var IIRC, but might not hurt.

tj90241 (dismissed):
Just two small suggestions.

tj90241 (approved):

comay (approved):

comay (approved):

PR #1757, opened 2020-12-11
Remove bird to bird2 upgrade (upgrade-bird.yml).

tj90241:
This was a temporary kludge to allow existing clusters
to be upgraded to bird2 from bird.  New clusters are
built with bird2 and this has been in the tree for a
sufficient amount of time, so it can be removed now.


tj90241:
Still testing; don't expect this to cause any issues, though.

tj90241:
Tested 3h3w without this commit + commit cherry-picked after, 3h3w build with the commit.

justinjpacheco (approved):

comay (approved):

PR #1758, opened 2020-12-21
`calico-felix` starts without TLS information in place for `etcd`

comay:
Tested by building a `3h3w` configuration and verifying that `calico-felix` on all nodes (and in particular, the second and third head nodes) was able to connect to `etcd`.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/etcd-proxy.rb
+++ b/chef/cookbooks/bcpc/recipes/etcd-proxy.rb
@@ -15,11 +15,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Headnodes have etcd-member and don't need etcd-proxy.
-return if headnode?
-
 include_recipe 'bcpc::etcd-packages'
 include_recipe 'bcpc::etcd-ssl'
+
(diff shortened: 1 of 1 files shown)


justinjpacheco (approved):

chef/cookbooks/bcpc/recipes/etcd-proxy.rb
```diff
@@ -15,12 +15,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Headnodes have etcd-member and don't need etcd-proxy.
-return if headnode?
-
 include_recipe 'bcpc::etcd-packages'
 include_recipe 'bcpc::etcd-ssl'
 
+# Headnodes have etcd-member and don't need etcd-proxy.
```
peter-at, line 21: is this comment valid? after the move of this line to after etcd* recipes?

comay: Yes, it's still valid - the rest of the recipe doesn't apply to head nodes but with the recent rework, the `etcd-proxy.rb` receipe is included on all nodes. So we still want to include the auxiliary recipes but not the rest of `etcd-proxy.rb`.

PR #1761, opened 2021-01-07
chef: Change ownership/mode of conf files

tj90241:
This commit ensures that automation renders all OpenStack
service configuration files (".conf") with mode 0640 and
sets user:group ownership as root:<roleaccount>.

In principle from a security perpsective, this is "better"
than what we are currently doing, which is to use mode 0600
and <roleaccount>:root.  Role accounts should not be writing
to configuration; they should only consume it.

This also alleviates some slight aches and pains with
in-place OpenStack upgrades, as Ubuntu packaging appears to
finagle with the modes of the configuration files.  Working
in the same direction that it finagles things results in
the headnodes being closer to convergance after the package
upgrade operations run.


tj90241:
I suspect we can also do this with `api-paste.ini` and other API-ish configuration files; testing to see how it impacts the builds and existing clusters.

comay (commented):
Tyler, can you clarify what owner/group/mode these files as delivered by Ubuntu in both Rocky and in Ussuri?

tj90241:
The ownership is not consistent (split between root:root in e.g., `keystone-common`, but `root:cinder` in `cinder-common`, as an example).  This inconsistency seems to be done to some `<service>-common` packages having a `postinst` script, while others do not:

e.g., taking a crack at `cinder-common`, which _does_ have a `postinst`:
```
apt-get download cinder-common
ar x cinder-common_2%3a13.0.9-0ubuntu1~cloud1_all.deb
tar -xf control.tar.xz
$ cat postinst
#!/bin/sh -e

CINDER_UID=64061
CINDER_GID=64061
<snip...>
    chown -R cinder:adm /var/log/cinder
    chmod 0750 /var/log/cinder
    chown -R cinder:cinder /var/lib/cinder
    chown -R root:cinder /etc/cinder
    chmod 0750 /etc/cinder
    chmod 0440 /etc/sudoers.d/cinder_sudoers
    chown -R root:root /etc/cinder/rootwrap.d
    chmod 0755 /etc/cinder/rootwrap.d
    chown root:root /etc/cinder/rootwrap.conf
<snip...>
```

There does seem to be a general consistency around mode 0640, though, as the mode appears to be derived from the `tar` file itself:
```
$ tar -tvf data.tar.xz 
<snip...>
-rw-r--r-- root/root       407 2020-05-05 17:21 ./etc/cinder/cinder.conf
```

tj90241:
Some of the inconsistencies are apparent on our builds today, e.g. ownership of: `/var/log/<dir>`:
```
vagrant@r3n1:~/keystone$ ls -l /var/log | grep keyst
drwx------  2 keystone  keystone             4096 Jan  7 06:25 keystone
vagrant@r3n1:~/keystone$ ls -l /var/log | grep cind
drwxr-x---  2 cinder    adm                  4096 Jan  7 06:25 cinder
```

tj90241:
Tests good, and makes the upgrade process smoother.  Ready for review.

comay (commented):
This looks fine but another general question - do the specific changes you have for each file match the ones supplied by Canonical for Rocky?

tj90241:
I am not sure; it would take awhile to audit each one.  Moreover, given some inconsistencies I have observed wrt the packaging, I would think it's best to not necessarily follow it just because the packaging does it.

comay (approved):

PR #1763, opened 2021-01-07
nova-compute: Avoid rendering of invalid config

tj90241:
If cpu_mode is _host-(model|passthrough)_, then cpu_model
should be disregarded; otherwise, one would have requested
a _custom_ CPU model or not specified to use the hosts'
configuration.

In fact, `nova` throws an exception during scheduling if
the `nova-compute` service specifies cpu_model when cpu_mode
is not _custom_, or when a cpu_model is not specified, but
cpu_mode is _custom_.

https://github.com/openstack/nova/blob/stable/rocky/nova/virt/libvirt/driver.py#L3983
```
Invalid: A CPU model name should not be set when a host CPU
model is requested
```

Thus, even if the chef-databag specifies this invalid
set of parameters, this commit ensures that the resulting
config will not result in scheduling errors.


tj90241:
I can also remove the note about `kvm64` in the project's README if we are willing to change the default `cpu_mode` attribute in this repository to `host-passthrough`.  With this commit, it should not break`virtualbox` builds or builds which rely on custom databag configuration, but still allow `libvirt/kvm` builds to work without the config change specified in the README.

chef/cookbooks/bcpc/templates/default/nova/nova-compute.conf.erb
```diff
@@ -18,10 +18,11 @@ live_migration_uri = qemu+ssh://nova@%s/system
 <% unless node['bcpc']['nova']['cpu_config']['cpu_mode'].nil? %>
 cpu_mode = <%= node['bcpc']['nova']['cpu_config']['cpu_mode'] %>
 <% end -%>
-<% unless node['bcpc']['nova']['cpu_config']['cpu_model'].nil? %>
+<% if node['bcpc']['nova']['cpu_config']['cpu_mode'] == "custom" %>
+<% raise "cpu_model must be specified for cpu_mode='custom'" if node['bcpc']['nova']['cpu_config']['cpu_model'].nil? %>
 cpu_model = <%= node['bcpc']['nova']['cpu_config']['cpu_model'] %>
 <% end -%>
-<% unless node['bcpc']['nova']['cpu_config']['cpu_model_extra_flags'].empty? %>
+<% unless node['bcpc']['nova']['cpu_config']['cpu_mode'].nil? or node['bcpc']['nova']['cpu_config']['cpu_model'].nil? %>
```
comay, line 25: From my reading of the documentation, setting `cpu_model_extra_flags` is allowed regardless of the value of `cpu_mode` (including the default) except for `none`.

chef/cookbooks/bcpc/templates/default/nova/nova-compute.conf.erb
```diff
@@ -18,10 +18,11 @@ live_migration_uri = qemu+ssh://nova@%s/system
 <% unless node['bcpc']['nova']['cpu_config']['cpu_mode'].nil? %>
 cpu_mode = <%= node['bcpc']['nova']['cpu_config']['cpu_mode'] %>
 <% end -%>
-<% unless node['bcpc']['nova']['cpu_config']['cpu_model'].nil? %>
+<% if node['bcpc']['nova']['cpu_config']['cpu_mode'] == "custom" %>
+<% raise "cpu_model must be specified for cpu_mode='custom'" if node['bcpc']['nova']['cpu_config']['cpu_model'].nil? %>
```
comay, line 22: Nit but I'd prefer this be written out as a nested-if because I'm not a fan of mixing conditionals and statements on the same line but perhaps that's just me.

chef/cookbooks/bcpc/templates/default/nova/nova-compute.conf.erb
tj90241: It's admittedly very muddy - I decided with the logic I did on the basis of this sentence from the documentation:
_For now, the cpu_model_extra_flags config attribute is valid only in combination with cpu_mode + cpu_model options._

However, it seems I found that in a copy of "Pike" documentation and they've reserved that call for Rocky in light of some of the Spectre and Meltdown changes:
_And specifying extra flags is valid in combination with all the three possible values for cpu_mode: custom (this also requires an explicit cpu_model to be specified)_

Good catch, I'll drop the conditional on this one.

PR #1771, opened 2021-01-21
Improve multihead build time

tj90241:
3h3w3r took ~70 minutes before this commit, ~61 minutes
afterwards.

While there are reasons to serially do passes around the
headnodes on the first build, subsequent builds can
leverage existing roles and have their client runs
performed in parallel.


tj90241:
Needs consideration for how to handle the fact that the VIP may be on something other than the last headnode, and that headnode will be subject to clustered service restarts.

PR #1772, opened 2021-01-22
Libvirt apt cache

mihalis68:
libvirt build had the apt cache commented out

**Describe your changes**
Remove unused code for apt cache. Adopt apt cache method used in virtualbox build

**Testing performed**
After making this change I was able to successfully perform make create to get the network vm and several cluster members VMs as normal.

### DIFF
--- a/virtual/Vagrantfile.libvirt
+++ b/virtual/Vagrantfile.libvirt
@@ -48,12 +48,6 @@ def load_hardware
   end
 end
 
-def mount_apt_cache(config)
-  user_data_path = Vagrant.user_data_path.to_s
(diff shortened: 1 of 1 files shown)


justinjpacheco (approved):

mihalis68:
on further testing I don't see the apt cache refill after nuking it. I think there's a problem here,

mihalis68:
Upon further testing, this results in trying to use an NFS share from the host, which while functional sometimes stops and asks for elevated privileges. Instead of harmonising withe virtualbox method (which uses some magic vbox share) I will instead replace this proposal with one that adopts the (working) share in the virtual/network/Vagrantfile.libvirt which uses a 9p share.

PR #1776, opened 2021-02-01
Create build for packer box

jingqgeng:
Added makefile and updated readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ machine for development and testing purposes.
 * 128 GB of free disk space
 * Vagrant 2.1+
 * VirtualBox 5.2+
+* Packer 1.4+
 * git, curl, rsync, ssh, jq, make, ansible
 
 **NOTE**: It is likely possible to build an environment with 16GB of RAM or less
@@ -42,6 +43,32 @@ choose from existing configurations in `virtual/topology`, or build your own.
 [topology.yml](virtual/topology/topology.yml) are used by default. To view a
 list of tested topologies and hardware configurations please see
 [virtual/README](virtual/README.md).
+* Set the variables in `virtual/vagrantbox.json`. The variable `vagrant_box` specifies the
+Vagrant box we use to build the virtual environment, and `vagrant_box_version` specifies
+the version of the Vagrant box.
+* If one would like to build a pre-provisioned custom Packer box and use it as the base box
+to create the virtual environment, the steps below should be followed:
+  * Create `virtual/packer/variables.json` and set the variables. Depends on the
+virtual machine provider, an example can be found at
+[variables.json.virtualbox.example](virtual/packer/variables.json.virtualbox.example)
+or [variables.json.libvirt.example](virtual/packer/variables.json.libvirt.example).
+This step is essential for building a Packer box that's used as a base box image for building
+the virtual environment. The variables `bcc_apt_key_url` and `bcc_apt_url` are optional,
+while others must be set. The variable `kernel_version` specifies the linux kernel version we'd
+like to have for the Packer box. While `base_box`, `base_box_version`, and `base_box_provider`
+specify an official Vagrant box we'd like to use as a baseline for the Packer box, upon which
+we make further modifications. Last but not least, the variable `output_packer_box_name` specifies
+the name we'd like to use when adding the output Packer box to Vagrant.
+  * Run make target `make create-packer-box`. This will create a Packer box and add it to Vagrant
+with the name specified by `output_packer_box_name`.
+  * Set the variables in `virtual/vagrantbox.json` accordingly. When a local custom box built by Packer
+is used, the variable `vagrant_box` needs to be set to the name of the Packer box (aka, the same as
+`output_packer_box_name` in `virtual/packer/variables.json`), and `vagrant_box_version` should be set to 0.
+  * After these steps, `make create all` would always use the Packer box, unless `virtual/vagrantbox.json`
+is specified otherwise.
+  * If the Packer box needs to be updated, we recommend first clean up the old Packer box. To clean up a
+Packer box, one must first make sure there's no VM using the Packer box by running `make destroy`, and then
+run `make destroy-packer-box` to clean up the Packer box.  
 * To make changes to the virtual topology without dirtying the tree, copy the
 [hardware.yml](virtual/topology/hardware.yml) and
 [topology.yml](virtual/topology/topology.yml) to files named
@@ -50,6 +77,15 @@ changes to them instead.
 * If a proxy server is required for internet access, set the variables TBD
 * If additional CA certificates are required (e.g. for a proxy), set the variables TBD
 * From the root of the chef-bcpc git repository run the following command:
+
+Download and install the latest version of Packer
+
+```shell
+wget https://releases.hashicorp.com/packer/1.6.6/packer_1.6.6_linux_amd64.zip -O ~/packer_1.6.6_linux.zip
+sudo apt install unzip
+sudo unzip ~/packer_1.6.6_linux.zip -d /usr/local/bin
+```
+
 
 Create a Python virtual environment (virtualenv) and activate it
 
@@ -64,20 +100,29 @@ To create a virtualbox build (the default):
 
 ```shell
 make generate-chef-databags
+make create-packer-box
 make create all
 ```
 
 To create a libvirt build:
 
 ```shell
-sudo apt-get install build-essential libvirt-dev qemu-utils
+sudo apt-get install build-essential libvirt-dev qemu-utils libguestfs-tools
 vagrant plugin install vagrant-libvirt vagrant-mutate
 vagrant box add bento/ubuntu-18.04 --box-version 202005.21.0 --provider virtualbox
 vagrant mutate bento/ubuntu-18.04 libvirt
 export VAGRANT_DEFAULT_PROVIDER=libvirt VAGRANT_VAGRANTFILE=Vagrantfile.libvirt
 make generate-chef-databags
+make create-packer-box
 make create all
 ```
+
+To clean up the build:
+```shell
+make destroy
+make destroy-packer-box
+```
+
 
 You may also want to change cpu model from `qemu64` to `kvm64` in
 `ansible/playbooks/roles/common/defaults/main/chef.yml`
@@ -143,6 +188,7 @@ chef-bcpc is built with the following open source software:
  - [HAProxy](http://haproxy.1wt.eu/)
  - [Memcached](http://memcached.org)
  - [OpenStack](http://www.openstack.org/)
+ - [Packer](https://www.packer.io/)
  - [Percona XtraDB Cluster](http://www.percona.com/software/percona-xtradb-cluster)
  - [PowerDNS](https://www.powerdns.com/)
  - [RabbitMQ](http://www.rabbitmq.com/)
--- a/virtual/Vagrantfile
+++ b/virtual/Vagrantfile
@@ -1,7 +1,7 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-# Copyright:: 2020 Bloomberg Finance L.P.
+# Copyright:: 2021 Bloomberg Finance L.P.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -17,8 +17,7 @@
 require 'yaml'
 require './lib/util'
 
-base_box = 'bento/ubuntu-18.04'
-base_box_version = '202005.21.0'
+vagrant_box, vagrant_box_version = Util.get_vagrant_box()
 
 # load the vm topology profile
 topology = './topology/topology.yml'
@@ -50,10 +49,9 @@ vb_folder = vb_folder.match(/^Default machine folder:\s+(.+)$/)[1]
 
 Vagrant.configure('2') do |config|
   config.ssh.forward_x11 = true
-  config.vm.box = base_box
-  config.vm.box_version = base_box_version
+  config.vm.box = vagrant_box
+  config.vm.box_version = vagrant_box_version
   config.vm.box_download_insecure = true
-  Util.mount_apt_cache(config)
 
   topology['nodes'].each do |node|
     vm_name = node['host']
--- a/virtual/Vagrantfile.libvirt
+++ b/virtual/Vagrantfile.libvirt
@@ -1,7 +1,7 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-# Copyright:: 2020 Bloomberg Finance L.P.
+# Copyright:: 2021 Bloomberg Finance L.P.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -17,8 +17,7 @@
 require 'yaml'
 require './lib/util'
 
-base_box = 'bento/ubuntu-18.04'
-base_box_version = '202005.21.0'
+vagrant_box, vagrant_box_version = Util.get_vagrant_box()
 
 def load_topology
   # load the vm topology profile
@@ -54,8 +53,8 @@ hardware = load_hardware
 
 Vagrant.configure('2') do |config|
   config.ssh.forward_x11 = true
-  config.vm.box = base_box
-  config.vm.box_version = base_box_version
+  config.vm.box = vagrant_box
+  config.vm.box_version = vagrant_box_version
   config.vm.box_download_insecure = true
 
   config.vm.provider :libvirt do |lv|
--- a/virtual/bin/create-virtual-environment.sh
+++ b/virtual/bin/create-virtual-environment.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2020, Bloomberg Finance L.P.
+# Copyright 2021, Bloomberg Finance L.P.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -65,11 +65,3 @@ fi
 "${virtual_dir}/bin/generate-ansible-inventory.py" \
     --ssh-config "${ssh_config_file}" \
     --topology-config "${topology_file}" > "${ansible_dir}/inventory.yml"
-
-# reboot vms to load new kernel
-if [ "${VAGRANT_DEFAULT_PROVIDER}" == "libvirt" ] ; then
-    ansible -i "${ansible_dir}/inventory.yml" cloud -b -B 1 -P 0 -m shell -a "sleep 5 && reboot"
-    ansible -i "${ansible_dir}/inventory.yml" cloud -m wait_for_connection -a "delay=15"
-else
-    ( cd "${virtual_dir}"; vagrant reload )
-fi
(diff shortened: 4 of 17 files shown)


jingqgeng (commented):
Talked with David and Justin offline about the PR. Besides the inline comments, there're other revisions we'd like to have on this PR. Including:
- Move /packer under /virtual. 
- Move the build targets in /packer/Makefile to chef-bcpc/Makefile so we don't have a Makefile under /packer (since no other sub-component under /virtual has its own Makefile).
- By default create-packer-box should always override the current box. So the user doesn't have to always call destroy-packer-box before building another box image.
- Since in Vagrantfiles we assume there is a packer box, in the make process we should check if the packer box exists. If not, print out a message explicitly about how the user should build a packer box first. 
- The document (README) can cover a bit more about how the packer make target works and be more explicit about the steps required for a build.
- Update .gitignore so files and binaries generated by packer won't show up in git diff 
- Download and install packer should be part of README. Instead of installing it in the root directory, /usr/local/bin would be a better option. 
- The output box can be better named. For example, bcc-\<os\>-\<version\>, like bcc-ubuntu-18.04.
- Config.json for packer can be parameterized, especially with the base box image (ubuntu-18.04), since it appears both in config.json and in create-packer-box.sh.
- Potentially config.json can be used just as an example like topology.example.yaml under /virtual/topology.

virtual/packer/packer_box_provision.sh
```diff
@@ -0,0 +1,86 @@
...
+	eval "$(grep ^GRUB_CMDLINE_LINUX= /etc/default/grub)"
+	NEW_CMDLINE="${GRUB_CMDLINE_LINUX} ipv6.disable=1"
+	sed -i.orig \
+            "s/^[#]*GRUB_CMDLINE_LINUX=.*/GRUB_CMDLINE_LINUX=\"${NEW_CMDLINE}\"/" \
+            /etc/default/grub
+	update-grub
+    fi
+}
+
+function cleanup_unused_kernel {
```
justinjpacheco, line 82: the command being run is generic to all unused packages, so perhaps this function should be renamed to `cleanup_unused_packages`?

virtual/Vagrantfile
```diff
@@ -16,10 +16,18 @@
...
 
+config_variable_file = File.read('./packer/variables.json')
+config_variables = JSON.parse(config_variable_file)
+base_box = config_variables['output_packer_box_name']
+base_box_dir = "~/.vagrant.d/boxes/#{base_box}/0"
+unless (File.directory?(File.expand_path(base_box_dir)))
+  raise "Packer box #{base_box} not found. Please run 'make create-packer-box' from the Makefile directory to generate #{base_box} first."
+end
+ 
+
```
justinjpacheco, line 30: could we put this block into a function that is sourced from a shared lib? identical copies are being used in multiple locations

virtual/Vagrantfile
```diff
@@ -16,10 +16,18 @@
...
 require './lib/util'
+require 'json'
 
-base_box = 'bento/ubuntu-18.04'
-base_box_version = '202005.21.0'
 
+config_variable_file = File.read('./packer/variables.json')
+config_variables = JSON.parse(config_variable_file)
+base_box = config_variables['output_packer_box_name']
+base_box_dir = "~/.vagrant.d/boxes/#{base_box}/0"
```
justinjpacheco, line 25: i don't think we should assume where the boxes are being stored. `vagrant box list` has the list of every box loaded into vagrants image store which abstracts the box's actual location. Perhaps the `Vagrant` class has access to this information but at the very least you should be able to get this by shelling out to `vagrant box list`

jingqgeng: Didn't find an attribute/function in Vagrant class to check all the boxes added. Used `vagrant box list` instead.

README.md
```diff
@@ -36,6 +37,10 @@ to provide the best out-of-the-box experience.
 
 ### Local Build
 
+* Review `virtual/packer/variables.json` and set the variables (those with default value
+'null' are required). An exmaple can be found at [variables.json.example](virtual/packer/
```
comay, line 41: s/exmaple/example/

README.md
```diff
@@ -36,6 +37,10 @@ to provide the best out-of-the-box experience.
...
+variables.json.example). This step issential for building a packer box as a base image
```
comay, line 42: s/issential/is essential/

virtual/Vagrantfile.libvirt
```diff
@@ -16,9 +16,10 @@
 # limitations under the License.
 require 'yaml'
 require './lib/util'
+require 'json'
```
comay, line 19: Remember to update the copyright year above to 2021.

virtual/bin/create-virtual-environment.sh
```diff
@@ -57,9 +57,6 @@ fi
     cd "${virtual_dir}"
     vagrant up
```
comay, line 59: Remember to update the copyright year above to 2021.

virtual/packer/config.json
```diff
@@ -0,0 +1,21 @@
+{
+    "provisioners": [
+        {
+            "type": "shell",
```
comay, line 4: Indentation level incorrect?

virtual/packer/packer_box_provision.sh
```diff
@@ -0,0 +1,86 @@
...
+deb ${apt_url} bionic-backports main restricted universe multiverse
+deb ${apt_url} bionic-security main restricted universe multiverse
+deb ${apt_url} bionic-updates main restricted universe multiverse
+EOF
+    fi
+
+    apt-get update
+}
+
+function upgrade_system {
```
comay, line 50: Could the comment from `finish.sh` be added before this function again?

virtual/packer/variables.json
```diff
@@ -0,0 +1,9 @@
+{
+    "bcc_apt_key_url":"",
```
comay, line 2: Could a space be added after each of the colons for readability?

virtual/packer/variables.json.example
```diff
@@ -0,0 +1,9 @@
+{
```
comay, line 1: Could a space be added after each of the colons (where missing) for readability?

virtual/packer/packer_box_provision.sh
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env bash
```
comay, line 1: This file should have the same copyright and license block as `finish.sh` did.

virtual/packer/packer_box_provision.sh
```diff
@@ -0,0 +1,86 @@
...
+
+apt_key_url="${BCC_APT_KEY_URL}"
+apt_url="${BCC_APT_URL}"
+kernel_version="${KERNEL_VERSION}"
+
+function main {
+    configure_vagrant_user
+    configure_apt
+    upgrade_system
+    cleanup_apt_cache
```
comay, line 14: Question for both you & @justinjpacheco - does it make sense to perform this here or later in the function sequence. I'm wondering if it makes sense to have the following order (important thing is downloading the packages to be installed later comes last)

    configure_vagrant_user
    configure_apt
    upgrade_system
    configure_linux_kernel
    cleanup_unused_packages
    cleanup_apt_cache
    download_debs

virtual/Vagrantfile
```diff
@@ -16,9 +16,9 @@
 # limitations under the License.
 require 'yaml'
 require './lib/util'
+require 'json'
```
justinjpacheco, line 19: i think this can be moved into lib/util now

virtual/packer/packer_box_provision.sh
```diff
@@ -0,0 +1,86 @@
...
+    download_debs
+    configure_linux_kernel
+    cleanup_unused_packages
+}
+
+
+function configure_vagrant_user {
```
justinjpacheco, line 21: i think this function can be kept in the finish.sh script use by vagrant directly

justinjpacheco: they will. this function adds the vagrant user to the operations group

virtual/lib/util.rb
```diff
@@ -14,6 +14,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+require 'json'
```
comay, line 17: If this file is changed, then the copyright year above should be adjusted too.

virtual/network/Vagrantfile
```diff
@@ -16,8 +16,7 @@
 # limitations under the License.
 require '../lib/util'
 
-base_box = 'bento/ubuntu-18.04'
-base_box_version = '202005.21.0'
+base_box = Util.get_base_box_name()
```
comay, line 19: Another file in need of a copyright year update.

virtual/network/Vagrantfile.libvirt
```diff
@@ -16,8 +16,7 @@
 # limitations under the License.
 require '../lib/util'
 
-base_box = 'bento/ubuntu-18.04'
-base_box_version = '202005.21.0'
+base_box = Util.get_base_box_name()
```
comay, line 19: Another file in need of a copyright year update.

virtual/packer/variables.json
```diff
@@ -0,0 +1,9 @@
+{
```
comay, line 1: Given the confusion from this morning, it seems best to not deliver this file but instead 1) document that the appropriate example file be copied into place and edited as needed and 2) the pathname `virtual/packer/variables.json` be added to `.gitignore`.

virtual/lib/util.rb
```diff
@@ -50,4 +52,19 @@ def self.mount_apt_cache(config)
     config.vm.synced_folder cache_dir, apt_cache_dir,
         create: true, owner: '_apt'
   end
+
+  def self.get_base_box_name
+    virtual_dir = File.dirname(File.dirname(__FILE__))
+    config_variable_file = File.read(virtual_dir + '/packer/variables.json')
```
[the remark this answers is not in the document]
jingqgeng: Do you mean the name of the base box we use, or the name of `variables.json`?

virtual/packer/config.json
```diff
@@ -0,0 +1,22 @@
+{
+  "provisioners" : [
+    {
+      "type" : "shell",
+      "environment_vars" : ["BCC_APT_KEY_URL={{user `bcc_apt_key_url`}}",
+			    "BCC_APT_URL={{user `bcc_apt_url`}}",
+			    "BCC_KERNEL_VERSION={{user `kernel_version`}}"],
```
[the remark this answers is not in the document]
jingqgeng: I'd prefer not to look for env vars outside of the Packer context. Although the env vars are used a lot in bve, this is the upstream project and the README doesn't mention anywhere about specifying env vars. So maybe we should apply that assumption here. 

For  `http_proxy` and `https_proxy`, I'd rather leave them out for now. They're not used in the packer box. If anyone needs proxy settings, this config file already provides an example of how it can be done.

virtual/packer/packer_box_provision.sh
```diff
@@ -0,0 +1,109 @@
...
+apt_url="${BCC_APT_URL}"
+kernel_version="${BCC_KERNEL_VERSION}"
+
+function main {
+    configure_apt
+    upgrade_system
+    configure_linux_kernel
+    cleanup_image
+    download_debs
+    sync;sync;sync;
```
comay, line 33: Only a single sync should be required. :-)

virtual/packer/config.json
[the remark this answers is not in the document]
comay: Let's discuss this one off-line.

virtual/lib/util.rb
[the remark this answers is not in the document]
comay: Let's discuss this one off-line.

README.md
```diff
@@ -42,6 +43,18 @@ choose from existing configurations in `virtual/topology`, or build your own.
...
+* Create `virtual/packer/variables.json` and set the variables. Depends on the
+virtual machine provider, an example can be found at
+[variables.json.virtualbox.example](virtual/packer/variables.json.virtualbox.example)
+or [variables.json.libvirt.example](virtual/packer/variables.json.libvirt.example).
+This step is essential for building a Packer box that's used as a base box image for building
+the virtual environment. The variables with value "" (`bcc_apt_key_url` and `bcc_apt_url`)
+are optional, while others must be set. The variable `kernel_version` specifies the
+linux kernel version we'd like to have for the Packer box. While `base_box`, `base_box_version`, 
+and `base_box_provider` specify an official Vagrant box we'd like to use as a baseline
+for the Packer box, upon which we would run provisions. Last but not least, the variable
```
tj90241, line 55: slightly awkward phrasing here:
"upon which we would run provisions"

"upon which we make further modifications", perhaps?

tj90241: (also, great job updating the docs)

virtual/lib/util.rb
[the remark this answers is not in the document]
tj90241: If we do use `variables.json`, I would love to see support for a `variables.json.overrides`, with the latter being added to `.gitignore` for the project.  With as many customizations as we have, it's easy to stub your toe when checking out different branches if these files are managed by git (and I actually did so while testing this PR - with this file!)

virtual/packer/packer_box_provision.sh
```diff
@@ -0,0 +1,109 @@
...
+            -o 'Dpkg::Options::=--force-confold' \
+        dist-upgrade
+}
+
+function configure_linux_kernel {
+    if [ -n "${kernel_version}" ]; then
+        apt-get install -y "linux-${kernel_version}"
+    fi
+
+    # Disable IPv6
```
tj90241, line 68: Seems like this should be done conditionally based on:
https://github.com/bloomberg/chef-bcpc/blob/188460c93765ef944e1359d43baf5e588b30fbba/ansible/playbooks/roles/common/defaults/main/base.yml#L290

tj90241: (though this kind of opens a can of worms; referencing that from this code isn't nice either... but given that the project's defaults are to enable IPv6, we also cannot disable IPv6 as part of the default provisioning steps here).

tj90241 (changes requested):
some nits

virtual/packer/packer_box_provision.sh
tj90241: Sorry, misread this as IPv6 is _enabled_ by default, which is not true.  So this is fine as-is.

virtual/lib/util.rb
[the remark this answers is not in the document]
jingqgeng: I actually removed `variables.json` out of the commit completely and expected people to follow the instructions in README and create `variables.json` with the right example json file for each provider. 

I think this might work better, since it forces people to think about which provider they use and which example file they should follow, while having `variables.overrides.json` and a default `variables.json` may give an impression to people that they may not need to change any of them.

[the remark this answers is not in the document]
tj90241: @jingqgeng ah okay, that solution works for me!

tj90241 (approved):

virtual/bin/create-virtual-environment.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2020, Bloomberg Finance L.P.
+# Copyright 2021, Bloomberg Finance L.P.
```
comay, line 3: Lines 68-75 which cause the VMs to be rebooted should be removed as we're already running the new kernel.

virtual/vagrantbox.json
```diff
@@ -0,0 +1,4 @@
+{
+  "vagrant_box": "",
```
comay, line 2: I think we can fill these in with the default name & version (0) of the box as specified in `variables.json` so that it's one less file one needs to edit by default.

comay (commented):
Looking very good - just three remaining comments below.

virtual/bin/create-virtual-environment.sh
jingqgeng: Good catch! I think this got pulled in when I rebased the branch.

virtual/vagrantbox.json
jingqgeng: Sure. I'll fill in the variables with the default `output_packer_box_name`  `bcc-ubuntu-18.04`, which is the same for both libvirt and vbox in variables.json

comay (dismissed):

jingqgeng:
Thank you! Will first talk to Mike and let him know about this change and then merge. Otherwise this may break Jenkins.

comay:
> Thank you! Will first talk to Mike and let him know about this change and then merge. Otherwise this may break Jenkins.

Yes, it definitely will. Mike should pull this change manually and see about uploading a box to BCS or somewhere in the Jenkins infrastructure. Alternatively, he can include building a box and then using it as part of the pipeline but I think those are two different things to test - box building and cluster building.

comay:
Jing, I believe the following changes would be useful here as well. The first is to set some of the environment variables that trigger warnings from `apt-get` when run in a non-interactive mode. The second is to do a `apt-get update` before doing the download since the clean up earlier may have removed some PPA data.

jingqgeng:
> Jing, I believe the following changes would be useful here as well. The first is to set some of the environment variables that trigger warnings from `apt-get` when run in a non-interactive mode. The second is to do a `apt-get update` before doing the download since the clean up earlier may have removed some PPA data.
> 
> ```
> diff --git a/virtual/packer/packer_box_provision.sh b/virtual/packer/packer_box_provision.sh
> index 55db83d2..5f208cbc 100644
> --- a/virtual/packer/packer_box_provision.sh
> +++ b/virtual/packer/packer_box_provision.sh
> @@ -20,6 +20,8 @@
>  set -xue
>  set -o pipefail
> 
> +export DEBIAN_FRONTEND='noninteractive' DEBIAN_PRIORITY='critical'
> +
>  apt_key_url="${BCC_APT_KEY_URL}"
>  apt_url="${BCC_APT_URL}"
>  kernel_version="${BCC_KERNEL_VERSION}"
> @@ -53,7 +55,6 @@ EOF
> 
>  # Taken from Ansible's dist upgrade logic for apt(8)
>  function upgrade_system {
> -    env DEBIAN_FRONTEND='noninteractive' DEBIAN_PRIORITY='critical' \
>          apt-get -y \
>              -o 'Dpkg::Options::=--force-confdef' \
>              -o 'Dpkg::Options::=--force-confold' \
> @@ -101,6 +102,7 @@ function cleanup_image {
>  }
> 
>  function download_debs {
> +    apt-get update
>      apt-get install --download-only -y -t bionic-backports \
>          bird2 init-system-helpers
>      apt-get install --download-only -y chrony tinyproxy unbound
> ```

Thank you for the comment! Updated the PR.

virtual/packer/packer_box_provision.sh
```diff
@@ -0,0 +1,112 @@
...
+deb ${apt_url} bionic-updates main restricted universe multiverse
+EOF
+    fi
+
+    apt-get update
+}
+
+# Taken from Ansible's dist upgrade logic for apt(8)
+function upgrade_system {
+    env DEBIAN_FRONTEND='noninteractive' DEBIAN_PRIORITY='critical' \
```
comay, line 58: Since we moving setting these environment variables to the head of the script, this line can be removed.

jingqgeng: Updated. Thank you!

comay (dismissed):
Great work, Jing. Marking this approved although it's not clear if there are still lint errors that need to be addressed - please check the link in the build report.

comay (approved):

PR #1777, opened 2021-02-08
Update QoS to help debug internal issue

mboruta1:
This commit addresses the following:
- rolls back the http timeout from 5 back to 10
- adds 127.0.0.1 to the QoS exemption list
- updates the 429.http file to contain proper line endings as per the
relevant RFCs
- print the SLO url instead of using html markup; most clients will
encounter this error when using the openstack client and thus the html
markup will not render properly

**Describe your changes**

**Testing performed**

**Additional context**


chef/cookbooks/bcpc/templates/default/haproxy/qos-exempt.acl.erb
```diff
@@ -1,3 +1,4 @@
+127.0.0.1
```
tj90241, line 1: Can localhost be used here instead of v4/v6 hardcoded?

tj90241 (dismissed):

PR #1779, opened 2021-02-09
Apply various QoS-related improvements

mboruta1:
This commit performs the following:
- rolls back the http timeout from 5 back to 10, the previous default
- adds localhost to the QoS exemption list
- updates the 429.http file to contain proper line endings as per the
relevant RFCs
- print the SLO url instead of using html markup; most clients will
encounter this error when using the openstack client and thus the html
markup will not render properly

**Describe your changes**
See the commit message above.

**Testing performed**
Tested on a BVE and production cluster.

### DIFF
--- a/chef/cookbooks/bcpc/templates/default/haproxy/429.http.erb
+++ b/chef/cookbooks/bcpc/templates/default/haproxy/429.http.erb
@@ -1,8 +1,9 @@
-HTTP/1.0 429 Too Many Requests
-Cache-Control: no-cache
-Connection: close
-Content-Type: text/html
-
+HTTP/1.0 429 Too Many Requests
+Cache-Control: no-cache
+Connection: close
+Content-Type: text/html
+
 <html><body><h1>429 Too Many Requests</h1>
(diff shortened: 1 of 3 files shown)


justinjpacheco (approved):

chef/cookbooks/bcpc/templates/default/haproxy/429.http.erb
```diff
@@ -1,8 +1,9 @@
-HTTP/1.0 429 Too Many Requests
-Cache-Control: no-cache
-Connection: close
-Content-Type: text/html
-
+HTTP/1.0 429 Too Many Requests
```
mboruta1, line 1: A note regarding the duplicate carriage returns added here. Chef currently does line ending conversion on templated files as described in [this](https://github.com/chef/chef/issues/4662) issue. Thus up until now the carriage returns in this template were actually being stripped by chef, which then resulted in an HTML file that does not conform to the relevant RFCs. 

Doubling up the carriage returns results in the desired line endings, but also results in the following warnings 'warning: encountered \r in middle of line, treated as a mere space' being printed by chef for every line these double carriage returns are in. As far as I can tell these warning cannot be suppressed.

comay (approved):

tj90241 (approved):

PR #1781, opened 2021-02-12
Explicitly manage apache2 log rotation

tj90241:
Allow for time and splay of apache2 reload/log rotation
across headnodes to be configured via chef.

### DIFF
--- a/chef/cookbooks/bcpc/attributes/apache2.rb
+++ b/chef/cookbooks/bcpc/attributes/apache2.rb
@@ -2,6 +2,20 @@
 # apache2
 ###############################################################################
 
+# apache2 WSGI daemons will be bounced when the logs reload, so for production
+# clusters, it is a good idea to stagger these bounces across the headnodes.
+# Similarly, it may not be desirable to bounce apache2 alongside the rest of
+# the cron.daily tasks.
+#
+# These attributes serve to control both aspects of reloads/log rotations.
+#   * start_hour and start_minute defines the hour and minute, respectfully,
+#     at which to start bounces/log rotations for apache2.
+#   * ...during which process, each headnode's reload will be separated by
+#     splay_minutes.
+default['bcpc']['apache2']['logrotation']['start_hour'] = 22
+default['bcpc']['apache2']['logrotation']['start_minute'] = 0
+default['bcpc']['apache2']['logrotation']['splay_minutes'] = 1
+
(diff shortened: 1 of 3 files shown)


chef/cookbooks/bcpc/recipes/apache2.rb
```diff
@@ -29,6 +29,28 @@
...
 
+# log rotation configuration
+template '/etc/cron.d/logrotate-apache2' do
+  source 'apache2/cron-logrotate.erb'
+  variables(
+    headnodes: headnodes(all: true)
+  )
+end
+
+remote_file '/etc/logrotate-apache2.conf' do
```
comay, line 40: This looks good although could we put this file under `/etc/apache2` as in something like `/etc/apache2/logrotate.conf`?

chef/cookbooks/bcpc/attributes/apache2.rb
```diff
@@ -2,6 +2,20 @@
...
+# the cron.daily tasks.
+#
+# These attributes serve to control both aspects of reloads/log rotations.
+#   * start_hour and start_minute defines the hour and minute, respectfully,
+#     at which to start bounces/log rotations for apache2.
+#   * ...during which process, each headnode's reload will be separated by
+#     splay_minutes.
+default['bcpc']['apache2']['logrotation']['start_hour'] = 22
+default['bcpc']['apache2']['logrotation']['start_minute'] = 0
+default['bcpc']['apache2']['logrotation']['splay_minutes'] = 10
```
comay, line 17: Do we have a sense on how long the whole Apache logrotation/reload dance takes for a head node? I would have thought a smaller splay would be preferable but if it really takes this long...

chef/cookbooks/bcpc/recipes/apache2.rb
tj90241: 👍 , will push a change

chef/cookbooks/bcpc/attributes/apache2.rb
tj90241: Probably takes under a minute to converge; I have no concerns dropping the splay to 1 minute.

tj90241: Dropped to 1 minute.

comay (approved):
Very nice work, Tyler.

PR #1787, opened 2021-02-26
Make mysql attributes more consistent

mboruta1:
This commit also introduces the mysql port attribute, which until now
was hard coded to 3306.

**Describe your changes**
Make references to mysql variables (url, port, dname) more consistent.

**Testing performed**
Created a 3h3w BVE and ran various rally tests.

### DIFF
--- a/chef/cookbooks/bcpc/templates/default/mysql/wsrep.cnf.erb
+++ b/chef/cookbooks/bcpc/templates/default/mysql/wsrep.cnf.erb
@@ -77,7 +77,7 @@ query_cache_type=0
 # it will have (most likely) disastrous consequences on donor node
 bind-address=<%= node['service_ip'] %>
 max_connections=<%= node['bcpc']['mysql']['max_connections'] %>
-port=3306
+port=<%= node['bcpc']['mysql']['port'] %>
 
 # Defaults that most openstack bits are happy with
 collation-server = utf8_general_ci
@@ -126,7 +126,7 @@ wsrep_cluster_name="<%= node['bcpc']['cloud']['region'] %>"
 wsrep_node_address="<%= node['service_ip'] %>:4567"
 
 # Address for incoming client connections. Autodetect by default.
-wsrep_node_incoming_address="<%= node['service_ip'] %>:4567"
+wsrep_node_incoming_address="<%= node['service_ip'] %>:<%= node['bcpc']['mysql']['port'] %>"
 
 # How many threads will process writesets from other nodes
 wsrep_slave_threads=<%= node['bcpc']['mysql']['wsrep_slave_threads'] %>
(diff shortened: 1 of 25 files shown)


chef/cookbooks/bcpc/templates/default/mysql/wsrep.cnf.erb
```diff
@@ -126,7 +126,7 @@ wsrep_cluster_name="<%= node['bcpc']['cloud']['region'] %>"
 wsrep_node_address="<%= node['service_ip'] %>:4567"
 
 # Address for incoming client connections. Autodetect by default.
-wsrep_node_incoming_address="<%= node['service_ip'] %>:4567"
```
justinjpacheco, line 129: this port needs to stay at 4567 for percona SST synchronization. perhaps we should add an `sst_port` attribute to mysql and use the same attribute on line 126

chef/cookbooks/bcpc/templates/default/mysql/wsrep.cnf.erb
```diff
@@ -126,7 +126,7 @@ wsrep_cluster_name="<%= node['bcpc']['cloud']['region'] %>"
...
+wsrep_node_incoming_address="<%= node['service_ip'] %>:<%= node['bcpc']['mysql']['port'] %>"
```
comay, line 129: Was this just an existing error where we had this set to 4567 but it should be 3306?

chef/cookbooks/bcpc/templates/default/mysql/wsrep.cnf.erb
mboruta1: I do not believe that's the case; [galera's docs](https://galeracluster.com/library/documentation/mysql-wsrep-options.html#wsrep-node-incoming-address) state that: `This parameter is used to provide the IP address and port from which the node should expect client connections.` The sst sync is done indeed via 4567, but specified by the `wsrep_cluster_address` variable, from which `wsrep_sst_receive_address` is set if not explicitly specified (which it is not in our template). 

Discovered this because Percona's proxysql-admin tool expected that it could connect to our galera cluster members via the IPs and ports specified by wsrep-node-incoming-address, which it couldn't given our previous setting of 4567, which is the galera replication port.

chef/cookbooks/bcpc/templates/default/mysql/wsrep.cnf.erb
mboruta1: Yes, please see my comment to Justin above. So far nothing was actually using this setting since, as per the galera docs, `It’s intended for integration with load balancers. For now, it’s otherwise unused by the node.`

justinjpacheco (approved):

comay (approved):

PR #1788, opened 2021-03-01
add availability zone anti-affinity.

npustchi:
add availability zone anti affinity scheduler filter

Changes

- Checks for a flag in nova attributes, if set the AZ anti-affinity filter is added.
- add the path and filter name to the nova.conf
- copy and compiles the filter.
- restart the nova-scheduler.

Testing performed
- Running BVE
   applied with enabled false that passed correctly and availability zone anti-affinity did not apply.
   applied with enabled true that availability zone anti-affinity applied, compiled and services restarted. the nova.conf was 
   changed and the services restarted.
- Fresh BVE
   I tested with a fresh build with the changes and enabled set to false. The build was successful and nova.conf changes 
   were correct and the affinity filter was not installed.
   I tested a fresh build with 3h3w and two AZs one with one host and one with two hosts. The attribute was enabled true. 
   Tested the functionality of the filter and its application. I was success.

### DIFF
--- a/chef/cookbooks/bcpc/files/default/nova/anti_affinity_availability_zone_filter.py
+++ b/chef/cookbooks/bcpc/files/default/nova/anti_affinity_availability_zone_filter.py
@@ -1,0 +1,80 @@
+# Copyright 2012, Piston Cloud Computing, Inc.
+# Copyright 2012, OpenStack Foundation
+# Copyright 2021, Bloomberg L.P.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from nova import availability_zones
+from nova import context
+from nova.scheduler import filters
+from oslo_log import log as logging
+from oslo_utils import strutils
+
+LOG = logging.getLogger(__name__)
+
+
+class _AntiAffinityAvailabilityZoneFilter(filters.BaseHostFilter):
+    """Checks the availability zone of the host if the
+    availability_zone_anti_affinity is set to true.
+    If the host is from a different zone than zones in the server group
+    it returns true, otherwise returns false."""
+
+    # a variable which is checked for all scheduler filters during a rebuild
+    # for each filter.
+    RUN_ON_REBUILD = False
+
+    def host_passes(self, host_state, spec_obj):
+        # Only invoke the filter if 'anti-affinity' and scheduler
+        # hint anti_affinity_policy = availability_zone is configured
+        instance_group = spec_obj.instance_group
+        policy = instance_group.policy if instance_group else None
+        az_hint = spec_obj.get_scheduler_hint('anti_affinity_policy', None)
+        # if the policy is not anti-affinity and anti_affinity_policy scheduler
+        # hint is not availability_zone then return true and don't apply filter
+        if self.policy_name != policy or az_hint != 'availability_zone':
+            return True
+        # Move operations like resize can check the same source compute node
+        # where the instance is. That case, AntiAffinityAvailabilityZoneFilter
+        # must not return the source as a non-possible destination.
+        if spec_obj.instance_uuid in host_state.instances.keys():
+            return True
+        # list of hosts of non-deleted instances in the server group
+        hosts_group_members = (
+            spec_obj.instance_group.hosts if spec_obj.instance_group else [])
+        # set of availability zones of the instances in the server group
+        instance_group_availability_zones = set()
+        host_availability_zone = availability_zones.get_host_availability_zone(
+            context.get_admin_context(), host_state.host)
+        for host in hosts_group_members:
+            instance_group_availability_zones.add(
+                availability_zones.get_host_availability_zone(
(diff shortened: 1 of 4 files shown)


chef/cookbooks/bcpc/files/default/nova/anti_affinity_availability_zone_filter.py
```diff
@@ -0,0 +1,81 @@
+# Copyright 2021, Bloomberg L.P.
```
comay, line 1: Nit but the Bloomberg copyright should appear after the other two.

chef/cookbooks/bcpc/recipes/nova-head.rb
```diff
@@ -339,6 +339,28 @@
...
+
+if anti_affinity['enabled']
+  available_filters.push(anti_affinity['filterPath'])
+  enabled_filters.push(anti_affinity['name'])
+
+  cookbook_file '/usr/lib/python2.7/dist-packages/nova/scheduler/filters/anti_affinity_availability_zone_filter.py' do
+    source 'nova/anti_affinity_availability_zone_filter.py'
+    mode '0644'
+    notifies :run, 'execute[compile az anti-affinity filter]', :immediately
+    notifies :restart, 'service[nova-scheduler]', :immediately
```
comay, line 355: Does the restart of `nova-scheduler` belong here or after you've compiled the filter file (assuming that is still necessary?)

chef/cookbooks/bcpc/files/default/nova/anti_affinity_availability_zone_filter.py
npustchi: will do

chef/cookbooks/bcpc/recipes/nova-head.rb
npustchi: yes this will compile the file and then restart the nova scheduler.

chef/cookbooks/bcpc/files/default/nova/anti_affinity_availability_zone_filter.py
```diff
@@ -0,0 +1,81 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class _AntiAffinityAvailabilityZoneFilter(filters.BaseHostFilter):
+    """Checks the availability zone of the host if the
+    availability_zone_anti_affinity is set to true.
+    If the host is from a different zone than zones in the server group
+    it returns true, otherwise returns false."""
+
+    # a varibale which is checked for all scheduler filters during a rebuild
```
tj90241, line 34: s/varibale/variable/

chef/cookbooks/bcpc/files/default/nova/anti_affinity_availability_zone_filter.py
```diff
@@ -0,0 +1,81 @@
...
+        if spec_obj.instance_uuid in host_state.instances.keys():
+            return True
+        # list of hosts of non-deleted instances in the server group
+        hosts_group_members = (
+            spec_obj.instance_group.hosts if spec_obj.instance_group else [])
+        # set of availability zones of the instances in the server group
+        instance_group_availability_zones = set()
+        host_availability_zone = availability_zones.get_host_availability_zone(
+            context.get_admin_context(), host_state.host)
+        for host in hosts_group_members:
```
tj90241, line 61: is there any concern about how this will impact scheduling database load when the server group is large?  It looks like `availability_zones.get_host_availability_zone` -> `nova.objects.aggregate.get_by_host`, which makes a database call -- so we'd be making N database calls 1 server.  I don't know how big we expect server groups to be, or whether or not this level of DB-querying is impactful at scale... am I being pedantic here?

tj90241: okay, if it's used in other parts of the OpenStack scheduling, it must not be too unexpected then.  Thanks!

chef/cookbooks/bcpc/files/default/nova/anti_affinity_availability_zone_filter.py
```diff
@@ -0,0 +1,81 @@
...
+    # a variable which is checked for all scheduler filters during a rebuild
+    # for each filter.
+    RUN_ON_REBUILD = False
+
+    def host_passes(self, host_state, spec_obj):
+        # Only invoke the filter if 'anti-affinity' and scheduler
+        # hint anti_affinity_policy = availability_zone is configured
+        instance_group = spec_obj.instance_group
+        policy = instance_group.policy if instance_group else None
+        az_hint = spec_obj.get_scheduler_hint('anti_affinity_policy', False)
```
comay, line 43: Since the hint is no longer a boolean, it seems the second argument should either be None or omitted altogether since the default is None. Then the check on line 46-47 can be simplified to

if self.policy_name != policy or az_hint != 'availability_zone':

chef/cookbooks/bcpc/files/default/nova/anti_affinity_availability_zone_filter.py
```diff
@@ -0,0 +1,81 @@
...
+        # if the policy is not anti-affinity and anti_affinity_policy scheduler
+        # hint is not availability_zone then return true and don't apply filter
+        if self.policy_name != policy or not az_hint or az_hint \
+            != 'availability_zone':
+            return True
+        # Move operations like resize can check the same source compute node
+        # where the instance is. That case, AntiAffinityFilter must not return
```
comay, line 50: s/AntiAffinityFilter/AntiAffinityAvailabilityZoneFilter/

justinjpacheco (changes requested):
pr looks good, could you squash the commits please?

npustchi:
> pr looks good, could you squash the commits please?

changed it.

npustchi:
> pr looks good, could you squash the commits please?

fixed the commits.

justinjpacheco (approved):

PR #1792, opened 2021-03-03
Add a stubnodes Makefile target

mboruta1:
Until now chef recipes were not run during a `make all`; this fixes that
oversight.

**Describe your changes**
See above.

**Testing performed**
Successfully ran `run-chef-client-stubnodes` on our internal test cluster.

**Additional context**
Until now chef recipes for stub nodes were not executed as part of `make all`. This fixes that by adding the `run-chef-client-stubnodes` target, which is invoked during a `make all`. So far chef recipes for stub nodes were executed via calls to ansible, bypassing the makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ headnodes = $$(ansible headnodes -i ${inventory} --list | tail -n +2 | wc -l)
 rmqnodes = $$(ansible rmqnodes -i ${inventory} --list | tail -n +2 | wc -l)
 storagenodes = \
         $$(ansible storagenodes -i ${inventory} --list | tail -n +2 | wc -l)
+stubnodes = $$(ansible stubnodes -i ${inventory} --list | tail -n +2 | wc -l)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,6 +9,8 @@ headnodes = $$(ansible headnodes -i ${inventory} --list | tail -n +2 | wc -l)
 rmqnodes = $$(ansible rmqnodes -i ${inventory} --list | tail -n +2 | wc -l)
 storagenodes = \
         $$(ansible storagenodes -i ${inventory} --list | tail -n +2 | wc -l)
+stubnodes = \
+        $$(ansible stubnodes -i ${inventory} --list | tail -n +2 | wc -l)
```
comay, line 13: I think you can fit everything here onto a single line.

mboruta1: Indeed, resolved in latest force push.

comay (approved):

PR #1793, opened 2021-03-04
README: pip: add `wheel`, `pip >= 19.1.1`

tj90241:
* `ansible` requires `wheel` to be installed, and will
    fail to install correctly (even when it is in the same
    `pip install ...` invocation as `wheel`).

  * Ensure users have `pip>=19.1.1` to avoid issues when
    installing `cryptography>3.4` as part of `pycryptodome`,
    etc.

Additionally, reflect the semantics of using `pip>=19.1.1`
but without `--upgrade` in the `rally` recipe.


README.md
```diff
@@ -53,6 +53,7 @@ Create a Python virtual environment (virtualenv) and activate it
 ```shell
 python3 -mvenv venv
 source venv/bin/activate
+pip install wheel 'cryptography<3.4'
```
comay, line 56: This is fine but was the workaround that @mboruta1 suggested with Rally of updating `pip` make more sense?

mboruta1: For reference my explanation for doing so can be found [here](https://github.com/bloomberg/chef-bcpc/pull/1784#issue-576482643). It is also the solution recommended by the package itself (see [this](https://github.com/pyca/cryptography/blob/main/CHANGELOG.rst#342---2021-02-08)).

tj90241: Tested suggestion and as expected, works great -- thanks!

mboruta1 (approved):
Looks good to me

comay (approved):

PR #1796, opened 2021-03-08
Makefile: enable parallelized virtual tasks

tj90241:
This commit addresses two items: one of correctness/safety,
and one of build.

  1) Our top-level `Makefile` is authored in such a way
     that none of the tasks should be parallelized, lest
     significant changes to the `Makefile` are done such
     that it is made declartive wrt each steps' specific
     requirements, rather than presenting an ordered list
     of steps under the `all` target.  As is, anyone who
     issues a `make -j all` is in for a bad time. To
     prevent surprises, make use of `.NOTPARALLEL:`

  2) Split out the `virtual` targets into their own
     `Makefile`, and invoke it recursively such that it
     will not be subject to `.NOTPARALLEL:`.

This allows users to run e.g. `make destroy create all -j`,
and have the destroy/creation of virtual infrastructure
done in parallel without affecting the inherently serial
parts of the `Makefile`.  This improves build time of a
1h1w by ~4.6% in my lab.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ storagenodes = \
         $$(ansible storagenodes -i ${inventory} --list | tail -n +2 | wc -l)
 stubnodes = $$(ansible stubnodes -i ${inventory} --list | tail -n +2 | wc -l)
 
+.NOTPARALLEL:
 all : \
 	sync-assets \
 	configure-operator \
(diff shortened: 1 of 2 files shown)


comay:
I understand the use of `.NOTPARALLEL` for the `all` target but I guess I don't understand how `-j` is useful unless you're referring to creating (or destroying) the virtual network and the virtual hosts at the same time. Can you please clarify?

tj90241:
So `.NOTPARALLEL` applies only to the current `make(1)` process.  Any instance of `make(1)` executed recursively will not inherent that property.  It serves two purposes in this case:
1) ensuring that `create destroy` targets are not issued to the `virtual` Makefile simultaneously, but rather 1-by-1
2) respecting the assumption that the `all` target makes

For `+$(make) -C virtual ...` we leverage the property that recursive executions of `make(1)` are parallel by default, and also pass the job servers configuration down (this is what the `+`) does.  This allows both `create` targets to be run in parallel with `-j`, and yes... both the virtual network and virtual hosts come up at the same time.

As for the virtual hosts coming up without the network -- that's possible as they bootstrap with the `vagrant` management network, which is independent of the network VMs' doings.  We're clear so long as we don't intend to standup BGP or use transits in the `virtual` environment provisioning process, and always assume that will be done by `chef-bcpc` automation.

tj90241:
Marking do-not-merge for now, this has some impact on internal builds (only) that I need to think about.

comay:
Correct, I don't think it's an issue with running `create-virtual-network` & `create-virtual-hosts` in parallel but I just wanted to understand the intent of being able to use `-j` here.

I'd like to have @justinjpacheco weigh in on this as I know he had some concerns with the Packer PR around the introduction of additional Makefiles.

PR #1797, opened 2021-03-08
create-virtual-*: Parallelize spinups/reboots

tj90241:
We previously used `vagrant reload` to turn all the VMs
one-by-one after they were provisioned with `vagrant up`.
Unfortunately, `vagrant reload` has no `--parallel` flag
as does `vagrant up`.

However, we have Ansible, and we generate an Ansible
playbook as part of the create-virtual-environment.sh
script, so just leverage that to do the reboots instead,
and do so in parallel.

Moreover, add `--parallel` to all `vagrant up` commands
to allow provisioners which respect this flag to come up
faster.

This improves build time of a 1h1w by ~1.8% in my lab.

### DIFF
--- a/virtual/bin/create-virtual-environment.sh
+++ b/virtual/bin/create-virtual-environment.sh
@@ -55,10 +55,7 @@ fi
 # bring up vagrant/virtualbox nodes
 (
     cd "${virtual_dir}"
-    vagrant up
-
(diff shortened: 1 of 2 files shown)


mihalis68 (approved):
Tested with a clean build of a 1h1w as far as completion of make create. All good.

PR #1798, opened 2021-03-08
ansible: Run common tasks in parallel during build

tj90241:
Each of the current Ansible groups imports two common
roles: `common` and `chef-node`.  By adding a playbook
for the common role itself, we can leverage this fact to
run Ansible tasks in parallel across all the groups,
whereas it was done serially in a group-by-group fashion
previously.

TBD:
  * `bootstraps` need `chef-node`
  * `headnodes` need `common`

This improves build time of a 1h1w by ~9.4% in my lab.


tj90241:
I tried deleting `common` and `chef-node` from all group-specific playbook roles as part of this commit,  but bumped into two things that prevented that.  We may want to move some things...

we would want to move this to `headnodes`:
https://github.com/bloomberg/chef-bcpc/blob/188460c93765ef944e1359d43baf5e588b30fbba/ansible/playbooks/roles/common/defaults/main/base.yml#L68

and this Chef stuff to `bootstraps`:
https://github.com/bloomberg/chef-bcpc/blob/188460c93765ef944e1359d43baf5e588b30fbba/ansible/playbooks/roles/common/defaults/main/base.yml#L130

mihalis68 (commented):
I am testing this one now.

comay:
Were you going to move `cloud_images` & `chef_extra_cookbooks`?

tj90241:
@comay yes, as well as remove `common` and `chef-node` roles from each of the group-specific playbooks.  This would be my suggestion if everyone is good with it.

mihalis68:
Tested this, it's fine, big speedup (full build of 1h1w as far as clouds ascii art in 27m)

comay:
I agree with your suggestion Tyler but as @justinjpacheco is the expert here, I'd like to see what he thinks.

PR #1800, opened 2021-03-08
chef: kill 'openstack <...> show | grep -w <...>'

tj90241:
For services/endpoints/objects that we provision with Chef,
we first do a scan using a 'openstack ... | grep ...'
check.

There are two problems with this:
  1) We are not setting pipefail, so if the OpenStack API
     mucks up during the left side of the pipe, the grep
     will still return success as if no object was found.
     We will then run the 'create' command, which will
     create a duplicate object if that command manages to
     go through.  This could cause a temporary outage
     as a result of a Chef run, that an operator may not
     at first realize.

  2) We tend to have cases where we run multiple `show`
     commands and fire up the `openstack` tool multiple
     times.  Instead, just list out all objects and do
     the filtering on the client/Chef side, as it's a
     lot faster -- especially with things like flavors.
     quotas, etc. where we have lots of them under CM.

This improves build time of a 1h1w by ~2% in my lab.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/cinder.rb
+++ b/chef/cookbooks/bcpc/recipes/cinder.rb
@@ -74,6 +74,27 @@ execute 'add openstack admin role to cinder user' do
 end
 # create cinder openstack user ends
 
+ruby_block "collect openstack service and endpoints list" do
+  block do
+    Chef::Resource::RubyBlock.send(:include, Chef::Mixin::ShellOut)
+    os_command = 'openstack service list --format json'
+    os_command_out = shell_out(os_command, :env => os_adminrc)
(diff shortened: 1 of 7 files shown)


tj90241:
If there's a good way to refactor the `ruby_block`s to be less repetitive, please let me know.  I couldn't think of a great way...

tj90241:
Linterator got me good on this one 😅

tj90241:
Oh yeah, and as an added bonus, I believe this will "fix" our automation so that it runs on OpenStack Train.

I say "fix" because it seems `openstack network show <textual_name>` appeared broken on Canonical-delievered `openstack` tools.  `openstack network list` was fine though and cuts around that problem for now.

PR #1801, opened 2021-03-08
mysql: Wait for Consul to elect a mysql primary

tj90241:
The MySQL recipe does not wait for Consul to elect a
database primary, and the next recipe to currently execute
(powerdns) tries to make use of the database host.  This
causes a race condition that has now begun to enumerate
more frequently... so wait for Consul to elect the primary
MySQL host as part of the MySQL recipe.


justinjpacheco (approved):

mihalis68:
Doing  a quick test on this now

PR #1806, opened 2021-03-11
meld multi-vm network build with libvirt bringup method

mihalis68:
**Describe your changes**
Quick hack at solving multi-vm build for libvirt

**Testing performed**

**Additional context**


virtual/network/netplan/conditionally-remove-default-route.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+ON_EDGE=@on_edge_flag@
+
+# Remove default routing from vagrant's management network when not on an edge
+# In those cases (i.e., non-spines), default route should be learned via bird
+if [ "$ON_EDGE" -eq 0 ] && [ "$IFACE"="eth0" ]; then
+	ip route del default dev $IFACE
+fi
```
comay, line 9: I'm not sure this is a better solution than actually terminating `dhclient` which should also remove the default route learned via DHCP.

virtual/network/provisioner.sh
```diff
@@ -51,17 +46,23 @@ base_config() {
...
-        sudo dhclient -x
-        sed "s|${DHCP_CONF}|${STATIC_CONF}|" \
-            "/vagrant/netplan/${1}.yaml" | sudo tee /etc/netplan/01-netcfg.yaml
-    fi
+    # TODO: use `dhcp4-overrides` instead of this bad hack
+    # TODO: once supported by future Ubuntu/netplan.io?
+    cp /vagrant/netplan/conditionally-remove-default-route.sh /etc/networkd-dispatcher/routable.d
+    sed -i "s/@on_edge_flag@/${on_edge_flag}/g" /etc/networkd-dispatcher/routable.d/conditionally-remove-default-route.sh
+    chmod +x /etc/networkd-dispatcher/routable.d/conditionally-remove-default-route.sh
+    cp /vagrant/netplan/${1}.yaml /etc/netplan/01-netcfg.yaml
```
comay, line 54: On a related note, doesn't this leave DHCP running on non-edge routers?

tj90241: These latest commits leave DHCP running on vagrant's management networks.

`dnsmasq` (which `vagrant-libvirt` uses behind the scenes) assigns dynamic leases; not static ones.  AFAIK, the leases are liable to be reclaimed if the hypervisor reboots, the virtual environment is left down, etc.  Although not a common case by any means, I don't see any particular downsides to leaving DHCP up, and it avoids the aforementioned issue.

comay: At least in the case of VirtualBox, we know what the management IP address is so I would prefer to continue making that disabled via the rendered `netplan` files. If we want to leave DHCP running in the `libvirt` configuration, then we can conditionally just copy the `netplan` file and the route deletion script.

tj90241: @comay - we now have feature parity and this works for both implementations.  Can we forego using static IPs here, keeping in mind that we are working away from Virtualbox?

comay: It seems wrong to be running DHCP and having to work-around default routes being introduced that way rather than strictly via BGP. Although this work-around works for both, it seems we're making up for a quirk in libvirt.

I'd like to take a look at this some more and see whether I can perhaps simplify things.

mihalis68:
if there are science experiments I can help with here I am more than happy to do so.

virtual/network/provisioner.sh
tj90241: We can remove the conditional default route, I think, if that's agreeable.  If so, I can go ahead and test that now.

I forget exactly why I had added it now... but generally speaking, we would not want traffic routing out the "management" interface, so to speak,  for _any_ implementation, right?

comay: At least with VirtualBox, the edge routers (so the `network` VM or the spines if there's a single pod, or the super-spines if there are multiple pods) do use the management interface for sending out traffic and so we've traditionally run DHCP there so we can obtain a default route.

It seems we should be able to use `dhcp4-overrides` to specifically set for `eth0` the following:

```
use-dns: false
use-ntp: false
use-routes: false
```

Then we can use DHCP on `eth0` for both providers.

(Note that this might be a flag-day in that earlier versions of Bionic didn't support `dhcp4-overrides` but current ones do.)

tj90241: Right, the current committed code is only (conditionally) removing the default route from non-edge networks, under the presumption that we don't want traffic egressing there.  libvirt/kvm also uses the default route from the management interface on the edge networks to provide external access by default.

I believe I had tried using `dhcp4-overrides` before this approach, but ran into the Bionic support issues.  Though I do see the LP# where the required patches were backported, so perhaps I overlooked something.

comay: I'll checkout this branch and see if I can make headway with `dhcp4-overrides`.

mihalis68:
Replaced by the related https://github.com/bloomberg/chef-bcpc/pull/1836 (the same work except for cleaning up the FIXME stuff around DHCP)

PR #1809, opened 2021-03-30
playbooks/common: Fix configure-system invocation

tj90241:
configure-system now requires the `virtual` facts to be
present in order to run.


mboruta1 (approved):
Looks good to me

PR #1810, opened 2021-03-30
ansible: Default to HWE kernel configuration

tj90241:
We now install the HWE kernel by default as part of the
packer/virtual builds, so we should make this the default
in Ansible.

Failing to do so results in kernel-utilities, such as
`bpftool`, `perf`, etc. being unavailable.  This issue has
manifested itself already as a Calico issue, as
`calico-felix` needs `bpftool` but it was not being
installed.


tj90241:
@mihalis68 this does not seem to impact my 3h3w builds (even with mutli-network vagrant stuff checked out) -- is it still breaking your builds?

comay (approved):

PR #1817, opened 2021-04-14
added patched version of vif.py

justinjpacheco:

### DIFF
--- a/chef/cookbooks/bcpc/recipes/nova-compute.rb
+++ b/chef/cookbooks/bcpc/recipes/nova-compute.rb
@@ -221,6 +221,13 @@ template '/etc/nova/nova-compute.conf' do
   notifies :restart, 'service[nova-compute]', :immediately
 end
 
+# install patched vif.py for libvirt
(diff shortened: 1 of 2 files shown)


chef/cookbooks/bcpc/recipes/nova-compute.rb
```diff
@@ -221,6 +221,12 @@
   notifies :restart, 'service[nova-compute]', :immediately
 end
 
+# install patched vif.py for libvirt
```
comay, line 224: This looks fine but can we document here why we're patching vif.py - namely to address https://bugs.launchpad.net/nova/+bug/1893263

comay (dismissed):

chef/cookbooks/bcpc/recipes/nova-compute.rb
justinjpacheco: thanks for the link. comment and git commit message updated

comay (approved):

PR #1818, opened 2021-04-14
added support for multiple primary aggregate cidrs

justinjpacheco:


mboruta1 (approved):
Looks good to me; traversed all places that could be affected by the change and think everything was covered.

comay (approved):

PR #1823, opened 2021-04-29
Clean up unused apt urls 

jingqgeng:
Signed-off-by: Jing Geng <jgeng25@bloomberg.net>

Since we moved most provisioning steps to building a packer box, we no longer need the apt and http env vars when we provision the VMs. So I'm cleaning up these unused env vars.


mboruta1 (approved):
Looks good to me.

comay (approved):

mboruta1 (approved):

PR #1824, opened 2021-05-03
DeviceManager's fill_dhcp_udp_checksums() shouldn't assume IPv6 is enabled

comay:

This provides a patched version of neutron/agent/linux/dhcp.py that addresses the following upstream bug:

https://bugs.launchpad.net/neutron/+bug/1915480

As a result of this bug, Calico's modified DHCP agent, calico-dhcp-agent, will occasionally crash when it tries to insert a special UDP checksum rule for IPv6 even though IPv6 is disabled. The actual patch is basically identical to the one upstream and is summarized by the following diff:

```
--- -   2021-05-03 23:31:38.429907469 +0000
+++ /usr/lib/python2.7/dist-packages/neutron/agent/linux/dhcp.py        2021-05-
03 23:05:24.340635209 +0000
@@ -28,6 +28,7 @@
 from oslo_log import log as logging
 from oslo_utils import excutils
 from oslo_utils import fileutils
+from oslo_utils import netutils
 from oslo_utils import uuidutils
 import six

@@ -1589,8 +1590,9 @@

     def fill_dhcp_udp_checksums(self, namespace):
         """Ensure DHCP reply packets always have correct UDP checksums."""
-        iptables_mgr = iptables_manager.IptablesManager(use_ipv6=True,
-                                                        namespace=namespace)
+        iptables_mgr = iptables_manager.IptablesManager(
+            use_ipv6=netutils.is_ipv6_enabled(), namespace=namespace
+        )
         ipv4_rule = ('-p udp -m udp --dport %d -j CHECKSUM --checksum-fill'
                      % constants.DHCP_RESPONSE_PORT)
         ipv6_rule = ('-p udp -m udp --dport %d -j CHECKSUM --checksum-fill'
```

This has been tested by building a cluster, creating some instances, and observing that exceptions were no longer be thrown by the DHCP agent.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/calico-work.rb
+++ b/chef/cookbooks/bcpc/recipes/calico-work.rb
@@ -50,6 +50,13 @@ end
   end
 end
 
+# install patched dhcp.py for calico-dhcp-agent
+# https://bugs.launchpad.net/neutron/+bug/1915480
+cookbook_file '/usr/lib/python2.7/dist-packages/neutron/agent/linux/dhcp.py' do
+  source 'neutron/dhcp.py'
+  notifies :restart, 'service[calico-dhcp-agent]', :immediately
+end
+
 template '/etc/neutron/neutron.conf' do
   source 'calico/neutron.conf.erb'
   mode '644'
--- a/chef/cookbooks/bcpc/files/default/neutron/dhcp.py
+++ b/chef/cookbooks/bcpc/files/default/neutron/dhcp.py
@@ -1,0 +1,1602 @@
+# Copyright 2012 OpenStack Foundation
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


justinjpacheco (approved):

chef/cookbooks/bcpc/recipes/calico-work.rb
```diff
@@ -50,6 +50,13 @@
   end
 end
 
+# install patched dhcp.py for calico-dhcp-agent
+# https://bugs.launchpad.net/neutron/+bug/1915480
+cookbook_file '/usr/lib/python2.7/dist-packages/neutron/agent/linux/dhcp.py' do
```
tj90241, line 55: This should _probably_ be `pycompile`'d to remain consistent with system software.

I haven't tested this and cannot confirm it for certainty, though I believe the `*.pyc` file on disk will not be updated in this case because `neutron-server` runs as `neutron`, but this `*.py` (and it's respective `*.pyc` file) is owned by `root`.  `python` should still pick up the changes, though it'll recompile the source file into bytecode each time rather than being able to load the bytecode out of the `pyc`.

tj90241 (approved):

chef/cookbooks/bcpc/recipes/calico-work.rb
comay: It's actually not necessary in this case as the pickled file is recreated because `calico-dhcp-agent` runs as `root`. 🙃

But I think it makes sense to investigate being consistent across all of `chef-bcpc` as part of another PR.

PR #1826, opened 2021-05-04
allow server resize on same host

justinjpacheco:


mboruta1 (approved):
Code looks good to me; done see any reason why this shouldnt be default true.

PR #1831, opened 2021-05-06
Move `calico-control` processing to avoid race with `neutron-server`

comay:

We should move the stanza concerning `calico-control` (which delivers the core Calico plugin for Neutron) to just before `/etc/neutron/neutron.conf` is rendered in order to avoid a race where the service definition of `neutron-server` hasn't apparently completed (plus it just make more sense to do it closer to the configuration file step).


mihalis68 (approved):
This makes sense, but we may need more. I'm testing something that builds on this now.

justinjpacheco (approved):

PR #1834, opened 2021-05-07
Patch the Neutron RBAC membership test

tj90241:
The current Neutron RBAC membership test results a database
query being issued, which itself makes use of a subquery
that is unable to make use of an index.
    
As the number of RBACs grows in a cloud deployment, the
missed opportunity for using the index becomes very apparent
with some API calls.  Adjust the SQLalchemy query
generation logic such that MySQL will make use of an index
where it did not previously.


tj90241:
I've just included @comay 's other Neutron patch for now, as I expect this PR to land after that.

chef/cookbooks/bcpc/files/default/neutron/rbac-query-performance.patch
```diff
@@ -0,0 +1,124 @@
+diff -ruw neutron/db/db_base_plugin_v2.py neutron/db/db_base_plugin_v2.py
```
tj90241, line 1: Just realizing now that this first hunk is no longer needed.

tj90241:
In BVE, you can measure impact of this patch quite simply:
```
unset OS_CLOUD
source an_rc_file.sh
# before and after:
time openstack network list
```

In my test env with a few 1k RBACs, I saw a improvement of 3.927s~4.084s => 1.474s~1.629s.  Virtually all of the improvement comes from the SQL-side, as the query time for a single query drops from ~1.2sec to fractions of a millisecond.

justinjpacheco (approved):

tj90241:
Rebased against Ussuri/8.2, just doing tests.

tj90241:
This is not quite working correctly in Ussuri.   Now the RBAC test is being applied in some cases where it doesn't make sense.

mboruta1:
Looks promising! Is there a high level summary regarding what was changed and why? This is complex code and thus difficult to evaluate its correctness; a summary would make things easier.

Also, how tested is this code? Would be nice to run it via our internal Jenkins pipeline and see whether the times for any of the operations have changed significantly (rally report can tell us).

And lastly, could we easily roll-back this change if it does not pan out?

tj90241:
Thanks -- I'll writeup some technical description of the higher-level changes to the logic here.  Some of that is in the LP ticket, but admittedly is a bit scattered:
https://bugs.launchpad.net/neutron/+bug/1918145/

As far as rollback goes, the changes from upstream should be re-plastered on (and `neutron-server` bounced) once the Chef changes been rolled back and synced with the bootstrap.  This can be done via:
```shell
ansible -I inventory.yml headnodes -m shell -ba 'apt-get install --reinstall -y python3-neutron python3-neutron-lib; chef-client -o bcpc::neutron-head'- f1`
```

For performance testing, see the comment of the `openstackclient` runtimes above.

For correctness testing, I tried comparing all permutations of `openstackclient network ... , port ...` etc. commands that I could think of, in both admin and non-admin contexts with and without `--share` and `--no-share` flags where applicable.  Similarly, I checked Horizon and whatnot and ensured that all the filters in the GUI worked.

I also mean to run unit-tests/integration-tests from `neutron-lib` against these changes and ensure that they still pass, but I admittedly haven't done that yet.

mboruta1 (approved):

tj90241:
Repushed to tickle the CI/CD checks...

mboruta1 (approved):

PR #1841, opened 2021-05-14
Do not install recommended packages

mihalis68:

**Describe your changes**
- tell apt never install apt recommended packages right in the packer box
- add python-setuptools to common packages (needed for etcd3gtw recipe

**Testing performed**
Builds clean most of the time. Some large builds require chefing the first head node be restarted due to an apparent race condition where neutron-server restarts in chef/cookbooks/bcpc/recipes/neutron-head.rb sometimes fail. That will be the subject of a separate PR

**Additional context**
Part of an effort to strip down BCC cluster member nodes to their mimimal essential packages to reduce the "surface area"to look at when reviewing CVEs and similar.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/etcd3gw.rb
+++ b/chef/cookbooks/bcpc/recipes/etcd3gw.rb
@@ -1,7 +1,7 @@
 # Cookbook:: bcpc
 # Recipe:: etcd3gw
 #
-# Copyright:: 2019 Bloomberg Finance L.P.
+# Copyright:: 2021 Bloomberg Finance L.P.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


chef/cookbooks/bcpc/recipes/common-packages.rb
```diff
@@ -36,6 +36,7 @@
   python-pip
   python-memcache
   python-mysqldb
+  python-setuptools
```
comay, line 39: Chris, in the interest of minimization it appears only `etcd3gw` requires this so I added it after `python-futurist` in the latter's recipe.

mihalis68: Ok i'll change this

comay (commented):
Code changes look good. What's left is to recommit using `git commit -s` and then squash all of the commits.

mihalis68:
I expect this to be more reliable now that we have merged https://github.com/bloomberg/chef-bcpc/pull/1844

justinjpacheco (approved):

comay (approved):

PR #1847, opened 2021-05-20
Introduce `storageheadnode` role

comay:

These changes introduce a new `storageheadnode` role that can be assigned to hosts that run Ceph control plane services including `ceph-mon` and `ceph-mgr`. For compatibility, existing headnodes will need to have this new role added to them.

The implementation borrows heavily from the existing worknode/storagenode and headnode/rmqnode split done earlier. The primary work is making sure the appropriate `/etc/ceph/ceph.conf` is rendered on each node type.

There are also several slightly orthogonal fixes to issues I encountered doing this work:

* The PowerDNS recipe requires `python-dnspython` & `python-jinja2` that were previously pulled into headnodes.

* Both the anti-affinity Nova scheduler filter & patched `nova-helper.py` for Watcher should have delayed service restarts

This has been built and tested by creating instances & volumes with both single and multiple racks scenarios, with coverged headnode/storageheadnode as well as separate roles.


chef/cookbooks/bcpc/recipes/powerdns.rb
```diff
@@ -178,6 +178,11 @@
   end
 end
 
+package %w(
```
tj90241, line 181: eh?

Makefile
```diff
@@ -169,25 +187,25 @@ run-chef-client-stubnodes :
 			-t chef-client --limit stubnodes; \
 	fi
 
-reweight-ceph-osds:
+reweight-ceph-osds :
 
 	ansible-playbook -v \
 		-i ${inventory} ${playbooks}/site.yml \
-		-t reweight-ceph-osds --limit headnodes
+		-t reweight-ceph-osds --limit storageheadnodes
```
tj90241, line 194: `storagenodes` should refer to the Ansible group, right?  In the oldschool "converged" nodes (e.g., a 1h1w build), I don't see a `storagenode` group being defined... what is it that is making it still work in those cases?

Makefile
```diff
@@ -256,13 +273,13 @@ generate-chef-environment :
 		-i ${inventory} ${playbooks}/site.yml \
 		-t generate-chef-environment --limit bootstraps
 
-adjust-ceph-pool-pgs:
+adjust-ceph-pool-pgs :
 
 	ansible-playbook -v \
 		-i ${inventory} ${playbooks}/site.yml \
-		-t adjust-ceph-pool-pgs --limit headnodes
+		-t adjust-ceph-pool-pgs --limit storageheadnodes
```
tj90241, line 280: ditto to the above, in case I've missed something.

chef/cookbooks/bcpc/recipes/nova-head.rb
```diff
@@ -211,12 +209,13 @@
 
 # nova package installation and service definition
 package %w(
+  ceph
```
tj90241, line 212: Perhaps just `python-rbd` and `python-rados` instead?

chef/cookbooks/bcpc/recipes/powerdns.rb
tj90241: aha, missed the notice in the commit log!

chef/cookbooks/bcpc/recipes/glance.rb
```diff
@@ -133,45 +133,41 @@
...
-  command "ceph osd pool set #{pool} size #{size}"
-  not_if "ceph osd pool get #{pool} size | grep -w 'size: #{size}'"
-end
+  rbd_users = []
+  rbd_users.append('glance')
 
-# create client.glance ceph user and keyring
-template '/etc/ceph/ceph.client.glance.keyring' do
-  source 'glance/ceph.client.glance.keyring.erb'
+  template '/etc/ceph/ceph.conf' do
```
tj90241, line 160: It feels weird to have a `ceph.conf` resource defined in the `glance` recipe, when e.g., `cinder` should also be making use of it.  Can we pull this into its own `ceph-client` recipe or something to that effect?  We could also dedup the resource defined in `nova-compute` in this way, I think.

Makefile
comay: Ships passing in the night - I noticed an issue earlier this evening when I was in fact doing a converged build. I pushed a fix for this particular issue. Please let me know if it makes sense.

Makefile
comay: Still thinking about what to do with this particular playbook - it's not part of the usual build and I want to do a bit more testing to see if a similar solution should be done here.

chef/cookbooks/bcpc/recipes/nova-head.rb
comay: I thought about that but for now I decided to mirror what's done in `nova-compute.rb`. Perhaps both of these could be minimized together.

chef/cookbooks/bcpc/recipes/glance.rb
comay: Yes, I went back and forth on this but decided to err on this implementation given the existing dance between `glance.rb` and `ceph-mon.rb`. Clearly if we were to ever separate Glance & Cinder being on the same host, this would definitely need to be reworked.

Let me go back over this and see if my earlier change can be made viable that did try to isolate the rendering of `ceph.conf` on the headnode and worknodes.

Makefile
comay: Okay, I've adjusted this one in a similar manner.

chef/cookbooks/bcpc/recipes/nova-head.rb
justinjpacheco: iirc, nova actually calls the `ceph` binary at some point which is why it is installed.

comay: Indeed, it did used to do that but it looks like you removed it (a `ceph auth get` cal) as part of introducing zone support in 00eeaa5246654bfdf0d820236df8bb0619625268

justinjpacheco: i meant that `nova` itself calls the `ceph` binary to gather some information

justinjpacheco: i ran into this issue when separating worknodes from storagenodes and i didn't want to install the ceph package. soon after doing so i saw nova throwing FileNotFound exceptions when looking for `/usr/bin/ceph`

comay: Interesting - I even wonder if that works nowadays without the admin key.

justinjpacheco: but by all means, give it a shot and see what happens. i'd be more than happy to be wrong on this :)

justinjpacheco: > Interesting - I even wonder if that works nowadays without the admin key.

i believe it uses the ceph username param in nova.conf to figure out which key it needs and passes that as an argument to /usr/bin/ceph

Makefile
comay: Okay, I believe I've addressed this in the `Makefile`.

chef/cookbooks/bcpc/recipes/nova-head.rb
comay: So it turns out that Cinder's `rbd` driver [needs](https://opendev.org/openstack/cinder/src/branch/stable/rocky/cinder/volume/drivers/rbd.py#L379) the `ceoh` binary in order to ascertain the mon addresses.

tj90241 (dismissed):

justinjpacheco (approved):

chef/cookbooks/bcpc/recipes/glance.rb
comay: As we discussed over a call, this isn't easy to do so I'm going to push this off for the moment. Longer term, we should definitely see if the situation can be improved through the use of Chef `include_recipe` statements and a recipe dedicated to rendering `/etc/ceph/ceph.conf`.

justinjpacheco (dismissed):

tj90241 (dismissed):

tj90241 (approved):

PR #1848, opened 2021-05-21
Address some lint issues with create-packer-box.sh

comay:


jingqgeng (approved):
Thank you for addressing the lint issues!

PR #1852, opened 2021-05-27
added message-ttl for versioned nova notifications

ajaytikoo:
Signed-off-by: Ajay Tikoo <atikoo@bloomberg.net>

**Description of changes**
Watcher relies on versioned notifications for keeping its cluster data model in
sync between periodic rebuilds. However, with versioned notifications
enabled, they remain in the rabbitmq queue as they are not consumed by
anything (though Watcher is using these messages, it may be requeing
them). In order to avoid the queues to grow indefinitely, we need to set a
`message-ttl` for the notification queues for versioned nova notifications.
The TTL has been set to 10 minutes, but can be changed in the attributes file.

**Testing performed**
These changes have been validated to be working as expected in the lab
environment

### DIFF
--- a/chef/cookbooks/bcpc/attributes/rabbitmq.rb
+++ b/chef/cookbooks/bcpc/attributes/rabbitmq.rb
@@ -19,3 +19,7 @@ default['bcpc']['rabbitmq']['ulimit']['nofile'] = 4096
 
 # Heartbeat timeout to detect dead RabbitMQ brokers
 default['bcpc']['rabbitmq']['heartbeat'] = 60
+
+# TTL for messages on the Nova (versioned) notifications queue that is consumed
+# by Watcher
+default['bcpc']['rabbitmq']['message_ttl']['watcher'] = 10 * 60 * 1000 # 10 minutes
--- a/chef/cookbooks/bcpc/recipes/rabbitmq.rb
+++ b/chef/cookbooks/bcpc/recipes/rabbitmq.rb
@@ -146,13 +146,62 @@ end
 rmqnodes = rmqnodes(all: true)
 ha_exactly = { 'ha-mode' => 'exactly', 'ha-params' => rmqnodes.length / 2 + 1 }
 ha_all = { 'ha-mode': 'all' }
-
 ha_policy = rmqnodes.length >= 3 ? ha_exactly : ha_all
 
 execute 'set rabbitmq ha policy' do
   command <<-DOC
     rabbitmqctl set_policy HA '^(?!(amq\.|[a-f0-9]{32})).*' '#{ha_policy.to_json}'
   DOC
+end
+
+if node['bcpc']['nova']['notifications']['format'] != 'unversioned'
+  # Nova versioned messages are enabled only for Watcher to be able to consume
+  # them. Messages remain in the queue even after Watcher has consumed them.
+  # This is because Watcher subscribes to the queue using the 'pool' option
+  # from the oslo_messaging. In order to prevent the queue from growing
+  # infinitely, we need to set a rabbitmq policy so that messages in the queue
+  # expire after a pre-defined time interval.
+  new_versioned_queue_policy = ha_policy.clone
+  new_versioned_queue_policy['message-ttl'] = node['bcpc']['rabbitmq'][
+    'message_ttl']['watcher']
+
+  # get the current policy settings for versioned notifications
+  begin
+    current_versioned_policy = JSON.parse(
+      shell_out!(
+        'rabbitmqctl list_policies --formatter json | jq -r \
+        \'.[] | select(.pattern == "^watcher_notifications.*")
+        | .definition\''
+      ).stdout.strip
+    )
+  rescue
+    current_versioned_policy = {}
+  end
+
+  # set the rabbitmq policy to define message-ttl
+  execute 'set ha and ttl policy for nova notification queues consumed by Watcher' do
+    command <<-DOC
+      rabbitmqctl set_policy "HA and TTL for queues that Watcher consumes" \
+        '^watcher_notifications.*' \
+        '#{new_versioned_queue_policy.to_json}' \
+        --apply-to queues --priority 1
+    DOC
+    only_if { current_versioned_policy.to_json != new_versioned_queue_policy.to_json }
+  end
+
+  current_versioned_message_ttl = current_versioned_policy['message-ttl']
+
+  # If the earlier policy did not set a message-ttl, purge the queues so that
+  # messages with no ttl defined do not exist in the queue.
+  ['watcher_notifications.info', 'watcher_notifications.error'].each do |queue|
+    execute "purge existing messages from #{queue}" do
+      command <<-DOC
+        rabbitmqctl purge_queue #{queue}
+      DOC
+      returns [0, 69]
+      only_if { current_versioned_message_ttl.nil? }
+    end
+  end
 end
 
 cookbook_file '/usr/local/bin/rabbitmqcheck' do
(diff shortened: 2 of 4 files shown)


chef/cookbooks/bcpc/attributes/rabbitmq.rb
```diff
@@ -19,3 +19,6 @@
 
 # Heartbeat timeout to detect dead RabbitMQ brokers
 default['bcpc']['rabbitmq']['heartbeat'] = 60
+
+# TTL for messages on the Nova versioned notifications queues
+default['bcpc']['rabbitmq']['nova_versioned_message_ttl'] = 60 * 60 * 1000 # 60 minutes
```
comay, line 24: Does this belong here or as a parameter under `['bcpc']['nova']['notifications']`? I realize it is a RabbitMQ queue but it's for Nova.

chef/cookbooks/bcpc/recipes/rabbitmq.rb
```diff
@@ -148,13 +148,47 @@
...
 
 execute 'set rabbitmq ha policy' do
   command <<-DOC
     rabbitmqctl set_policy HA '^(?!(amq\.|[a-f0-9]{32})).*' '#{ha_policy.to_json}'
   DOC
 end
 
+# check if TTL is already set for versioned notifications
+message_ttl_defined = shell_out!(
+  'rabbitmqctl list_policies | grep versioned_notifications | grep message-ttl | wc -l'
```
comay, line 164: It would be better to use the `--formatter json` argument to `rabbitmqctl` along with `jq` to parse this (not only its existence but the actual value).

chef/cookbooks/bcpc/attributes/rabbitmq.rb
ajaytikoo: > Does this belong here or as a parameter under `['bcpc']['nova']['notifications']`? I realize it is a RabbitMQ queue but it's for Nova.

It is a setting for a rabbitmq queue that is used by Nova. Since the setting is used by the rabbitmq recipe, I though that this would be the better place for it rather than nova.

chef/cookbooks/bcpc/recipes/rabbitmq.rb
ajaytikoo: adding this in the next commit

chef/cookbooks/bcpc/recipes/rabbitmq.rb
```diff
@@ -148,13 +148,77 @@
...
 
 execute 'set rabbitmq ha policy' do
   command <<-DOC
     rabbitmqctl set_policy HA '^(?!(amq\.|[a-f0-9]{32})).*' '#{ha_policy.to_json}'
   DOC
 end
 
+# get the current policy settings for versioned notifications
+begin
+  current_versioned_policy = JSON.parse(
```
tj90241, line 164: The `rabbitmq` recipe runs _before_ the OpenStack components, and thus before any queues have been defined for OpenStack.
That being said, how does this work for net-new builds if it needs to pick out and list the existing queues?

tj90241: I guess with a `3h3w`, the second headnode's run of `rabbitmq` will have benefited from the first headnode's chef run, but this does not work for a `1h1w` or similar configs - right?

tj90241: nevermind, I see this is just checking to see if there is a current policy in order to determine if queues have to be purged...

tj90241 (dismissed):

chef/cookbooks/bcpc/recipes/rabbitmq.rb
```diff
@@ -148,13 +148,77 @@
...
+  DOC
+  only_if do
+    node['bcpc']['nova']['notifications']['format'] != 'unversioned' and
+      current_versioned_policy.to_json != new_versioned_queue_policy.to_json
+  end
+end
+
+current_versioned_message_ttl = current_versioned_policy['message-ttl']
+
+execute 'purge existing messages from versioned_notifications.info' do
```
justinjpacheco, line 190: test configuring nova to only export watch_notifications.(info,error) so that we ca only set ttl options and purge that queue.

chef/cookbooks/bcpc/recipes/rabbitmq.rb
```diff
@@ -148,13 +148,77 @@
 ha_all = { 'ha-mode': 'all' }
 
 ha_policy = rmqnodes.length >= 3 ? ha_exactly : ha_all
+new_versioned_queue_policy = ha_policy.clone
```
justinjpacheco, line 151: lets group these changes together to that message queue purging is all together with a nice description of why we're doing this

chef/cookbooks/bcpc/recipes/rabbitmq.rb
```diff
@@ -148,13 +148,77 @@
...
   DOC
 end
 
+# get the current policy settings for versioned notifications
+begin
+  current_versioned_policy = JSON.parse(
+    shell_out!(
+      'rabbitmqctl list_policies --formatter json | jq -r \'.[]\' \\
+      | jq -r \'select(.pattern == "^(versioned|watcher)_notifications.*")\' \\
+      | jq -r \'.definition\''
```
comay, line 168: I don't believe this will work as intended with `jq(1)` as the comparison inside the `select()` can only be AFAIK against a string and not a regex pattern (`match(()` supports regex but that's not really want you want here). Also, you shouldn't need to run three `jq` commands but only a single one of the form

```
rabbitmqctl list_policies --formatter json | jq -r \
  '.[] |
   select(.pattern == "^versioned_notifications.*" or
          .pattern == "^watcher_notifications.*") |
   .definition'
   ```

Comment (ajaytikoo) [reply chef/cookbooks/bcpc/recipes/rabbitmq.rb:168]: @comay , I understand your concern with the select not supporting regex. However, the intention here is not to compare the regex, but the text itself. The regex is contained in the policy definition, and I am comparing the policy definition itself, so there is not reason to interpret regex here. I agree on combining the piped statements to one single jq command, and will make that change in the next commit.

Comment (comay) [reply chef/cookbooks/bcpc/recipes/rabbitmq.rb:168]: I'm afraid I don't understand - are you wanting to select only the policies where `pattern` is either `"^versioned_notificatons.*"` or `"watcher_notifications.*"`? If that's the case, then I don't believe the above `select()` works although perhaps I'm mistaken.

File: chef/cookbooks/bcpc/recipes/rabbitmq.rb (hunk 4 above)
Comment (ajaytikoo) [reply chef/cookbooks/bcpc/recipes/rabbitmq.rb]: included changes in the new commit.

File: chef/cookbooks/bcpc/recipes/rabbitmq.rb (hunk 5 above)
Comment (ajaytikoo) [reply chef/cookbooks/bcpc/recipes/rabbitmq.rb]: done in the new commit

File: chef/cookbooks/bcpc/recipes/rabbitmq.rb (hunk 6 above)
Comment (ajaytikoo) [reply chef/cookbooks/bcpc/recipes/rabbitmq.rb]: already discussed this with @comay

**Review by justinjpacheco — DISMISSED** [review]

File: chef/cookbooks/bcpc/templates/default/watcher/watcher.conf.erb [hunk 7]
```diff
@@ -36,3 +36,9 @@ auth_uri = <%= "https://#{@node['bcpc']['cloud']['fqdn']}:5000/" %>
 memcached_servers = <%= @headnodes.map{ |n| "#{n['service_ip']}:11211" }.join(',') %>
 auth_type = password
 
+<% if node['bcpc']['nova']['notifications']['format'] != 'unversioned' %>
+[watcher_decision_engine]
+# The default value for 'notifications_topics' is 'nova.versioned_notifications,watcher.watcher_notifications'.
+# However, we are changing it to match the queue name for versioned messages that we have configured in nova config file.
```
comay, line 42: This looks ready to go but before I approve the PR, can you make sure that this comment line (and please check the others you're adding as well) doesn't exceed 80 characters for readability.

comay (approved):

comay (approved):

PR #1857, opened 2021-06-08
configure-chrony: fix IPv4/6 compatibility issue

tj90241:
When the chef-bcpc `system_enable_ipv6` attribute is set to
`False`, it may be the case that `chrony` still tries to
open `AF_INET6` sockets to connect to timeservers.

This will result in `chrony` failing to start with the
following logged to the system journal:
```
Jun 08 21:37:38 <hostname> chronyd[5708]: Could not open IPv6 command socket : Address family not supported by protocol
```

To avoid this, configure `chrony` to start with `-4` when
`system_enable_ipv6` is set to `False`.

### DIFF
--- a/ansible/playbooks/roles/common/templates/chrony/default.j2
+++ b/ansible/playbooks/roles/common/templates/chrony/default.j2
@@ -1,0 +1,10 @@
+# This is a configuration file for /etc/init.d/chrony and
+# /lib/systemd/system/chrony.service; it allows you to pass various options to
+# the chrony daemon without editing the init script or service file.
+
+# Options to pass to chrony.
+DAEMON_OPTS="{{ '-4' if not system_enable_ipv6 else '' }}"
+
+# Sync system clock in containers or without CAP_SYS_TIME (likely to fail)
+# See /usr/share/doc/chrony/README.container for details.
+SYNC_IN_CONTAINER="no"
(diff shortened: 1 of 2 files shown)


peter-at (approved):

justinjpacheco (dismissed):

ansible/playbooks/roles/common/templates/chrony/default.j2
```diff
@@ -0,0 +1,10 @@
+# This is a configuration file for /etc/init.d/chrony and
+# /lib/systemd/system/chrony.service; it allows you to pass various options to
+# the chrony daemon without editing the init script or service file.
+
+# Options to pass to chrony.
+DAEMON_OPTS="{{ '-4' if not system_enable_ipv6 else '' }}"
+
+# Sync system clock in containers or without CAP_SYS_TIME (likely to fail)
+# See /usr/share/doc/chrony/README.container for details.
+SYNC_IN_CONTAINER="no"
```
comay, line 10: Is this a change from the defaults?

tj90241: This file is copy/pasted from the packaging's default configuration, so yes.  I did fix a small spelling error that was present in the comment string there, though.

comay: Ah, got it - I forgot you're replacing `/etc/default/chrony` rather augmenting some default values.

comay (approved):

PR #1860, opened 2021-06-14
packer: Install tools mated with kernel

tj90241:
Calico requires linux-tools-`uname -r` to be present.  With
upstream builds, it may be the case that a packer image is
generated on a specific date, with kernel x.y.z installed.
The Ansible automation will install the linux-tools-*
metapackage, which may install kernel x.y.theta tools.

To avoid this drift, install the tools mated with the
kernel at packer provisioning time.


comay (approved):
I don't believe Calico actually requires the package since `bpftool` isn't actually necessary. But this fix is good in that it eliminates noise in the logs.

justinjpacheco (approved):

PR #1874, opened 2021-07-12
increased retry timeout for watcher's vm migration

ajaytikoo:
**Describe your changes**
When migrating an instance, Watcher polls to check the status of the migrating instance for 120 seconds. If migration does not complete (or fail) within this time, Watcher considers that the migration failed, does not try to migrate any pending instances, and marks the action plan as failed. In a large production environment, it generally takes longer time for a migration to complete. In order to avoid Watcher failing a migration prematurely, this change makes the timeout and poll interval a configurable parameter, which defaults to the same values as are right now hard coded in the code (120 seconds and 1 second respectively). In environments where it normally takes longer to live migrate instances, Watcher can be configured with higher timeouts so that it will not consider a migration failed unless it actually fails or takes longer than the configured time to migrate.

Changes in the patched files:
1. **conf/__init__.py**: the changes to this file are two additional lines, one importing nova_helper from watcher.conf, and other registering its config options.
2. **conf/nova_helper.py**: this is a new file, which implements the configuration items for nova_helper.

**Testing performed**
Testing has been performed in the virtual lab environment.

### DIFF
--- a/chef/cookbooks/bcpc/files/default/watcher/conf/__init__.py
+++ b/chef/cookbooks/bcpc/files/default/watcher/conf/__init__.py
@@ -1,0 +1,62 @@
+# -*- encoding: utf-8 -*-
+# Copyright (c) 2016 b<>com
+# Copyright (c) 2016 Intel Corp
+#
+# Authors: Vincent FRANCOISE <vincent.francoise@b-com.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


tj90241 (changes requested):
Needs to be rebased against 8.2/python3.

tj90241:
Please reopen if this is still needed; cleaning up some old branches.

PR #1877, opened 2021-07-15
Upgrade to OpenStack Ussuri

tj90241:
Change the cluster build target from Rocky/Mimic to Ussuri/Octopus


tj90241:
Still have a TODO or two sprinkled in there, and there are some comments across config files that reference Pike/Rocky/Stein/Train/etc...

Otherwise, though, I do not envision a whole lot of changes here beyond what turns up in code review.

justinjpacheco (approved):
nice work 👍

justinjpacheco:
is it fair to say these commits are tightly coupled with one another? if that's the case, should they be squashed into a single "upgrade to ussrui" commit?

tj90241:
ya, I intend to squash at the end... just keeping it separated for review (and so when I push these lint fixes, anyone rebasing off this branch periodically won't have as much of a hassle).

chef/cookbooks/bcpc/attributes/placement.rb
```diff
@@ -0,0 +1,45 @@
+###############################################################################
```
comay, line 1: Some of these values seem to be unused (inherited from `nova.rb`?)

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -76,6 +76,27 @@
 end
 # create cinder openstack user ends
 
+ruby_block 'collect openstack service and endpoints list' do
```
comay, line 79: I'm wondering if this particular `ruby_block` could be put in a library file and reference as a function from there since it's repeated numerous times in the recipes?

chef/cookbooks/bcpc/attributes/placement.rb
tj90241: wow, completely overlooked that aspect... thanks!

chef/cookbooks/bcpc/recipes/cinder.rb
tj90241: Likely, yes, I'll give it a stab.  My (in)experience with Chef is showing here.

tj90241: Pushed a fix for this.

chef/cookbooks/bcpc/libraries/utils.rb
```diff
@@ -354,3 +354,28 @@ def cidr_to_reverse_zones(cidr)
...
+
+def openstack_endpoints
+    Chef::Resource::RubyBlock.send(:include, Chef::Mixin::ShellOut)
+    os_command = 'openstack endpoint list --format json'
+    os_command_out = shell_out(os_command, env: os_adminrc)
+    endpoint_list = JSON.parse(os_command_out.stdout)
+
+    # build a host of service_type => list of interfaces present
+    groups = endpoint_list.group_by { |e| e['Service Type'] }
+    endpoints = groups.map { |k, v| [k, v.map { |e| e['Interface'] }] }.to_h
```
justinjpacheco, line 366:
```suggestion
    groups.map { |k, v| [k, v.map { |e| e['Interface'] }] }.to_h
```

justinjpacheco: in ruby, the last statement is what will be returned, so you don't need to assign it

chef/cookbooks/bcpc/libraries/utils.rb
```diff
@@ -354,3 +354,28 @@ def cidr_to_reverse_zones(cidr)
...
+def openstack_services
+    Chef::Resource::RubyBlock.send(:include, Chef::Mixin::ShellOut)
+    os_command = 'openstack service list --format json'
+    os_command_out = shell_out(os_command, env: os_adminrc)
+    service_list = JSON.parse(os_command_out.stdout)
+
+    # build a hash of service_type => list of uris
+    services = service_list.map { |s| s['Type'] }
+
+    services
```
justinjpacheco, line 380:
```suggestion
    service_list.map { |s| s['Type'] }
```

justinjpacheco: same here as endpoints

chef/cookbooks/bcpc/libraries/utils.rb
```diff
@@ -354,3 +354,28 @@ def cidr_to_reverse_zones(cidr)
...
+    Chef::Resource::RubyBlock.send(:include, Chef::Mixin::ShellOut)
+    os_command = 'openstack endpoint list --format json'
+    os_command_out = shell_out(os_command, env: os_adminrc)
+    endpoint_list = JSON.parse(os_command_out.stdout)
+
+    # build a host of service_type => list of interfaces present
+    groups = endpoint_list.group_by { |e| e['Service Type'] }
+    endpoints = groups.map { |k, v| [k, v.map { |e| e['Interface'] }] }.to_h
+
+    endpoints
```
justinjpacheco, line 368:
```suggestion
    groups.map { |k, v| [k, v.map { |e| e['Interface'] }] }.to_h
```

justinjpacheco: the last statement is implicitly returned (in ruby) so you don't need to do this extra assignment

chef/cookbooks/bcpc/recipes/ceph-packages.rb
```diff
@@ -24,5 +24,10 @@
 
 package %w(
   ceph
-  ceph-deploy
 )
+
+# https://bugs.launchpad.net/ubuntu/+source/ceph/+bug/1892448
+execute 'workaround broken ceph-base packaging in the Ussuri UCA' do
```
justinjpacheco, line 30: nit: you can pass dpkg options to the `package` chef resource, but not a big deal

chef/cookbooks/bcpc/attributes/placement.rb
tj90241: Addressed in recent commit -- was able to cleanup the `placement` attributes and `placement.conf.erb` a good deal.

chef/cookbooks/bcpc/libraries/utils.rb
tj90241: thanks, will address both of these in the next lint push

tj90241: done

chef/cookbooks/bcpc/recipes/ceph-packages.rb
tj90241: thanks, changed it to use the `package` resource with `options`

chef/cookbooks/bcpc/recipes/placement.rb
```diff
@@ -0,0 +1,247 @@
...
+end
+
+execute 'enable placement-api' do
+  command 'a2ensite placement-api'
+  not_if 'a2query -s placement-api'
+end
+#
+# configure placement ends
+
+# create/manage placement databases starts
```
mboruta1, line 177: Assuming that we do want to route db accesses by the placement api via proxysql when its enabled, we should add the placement mysql user to ProxySQL (e.g. the two resources in present [here](https://github.com/bloomberg/chef-bcpc/blob/master/chef/cookbooks/bcpc/recipes/neutron-head.rb#L160)) before adding it to mysql.

chef/cookbooks/bcpc/recipes/placement.rb
```diff
@@ -0,0 +1,247 @@
...
+
+region = node['bcpc']['cloud']['region']
+config = data_bag_item(region, 'config')
+
+mysqladmin = mysqladmin()
+
+# used for database creation and access
+#
+database = {
+  'host' => node['bcpc']['mysql']['host'],
```
mboruta1, line 26: `host` and `port` should point to the mysql host iff proxysql is not enabled, otherwise to the proxysql primary (unless I am missing something and database accesses done by the placement service should not go via ProxySQL). Example of how I did this in neutron can be found [here](https://github.com/bloomberg/chef-bcpc/blob/master/chef/cookbooks/bcpc/recipes/neutron-head.rb#L33).

mboruta1 (commented):
(ProxySQL-related comments)

chef/cookbooks/bcpc/recipes/placement.rb
tj90241: Thanks, totally forgot to backport to account for that.

chef/cookbooks/bcpc/recipes/placement.rb
tj90241: Will do!

tj90241: done

chef/cookbooks/bcpc/recipes/placement.rb
tj90241: done

chef/cookbooks/bcpc/recipes/nova-compute.rb
```diff
@@ -239,7 +244,7 @@
 
 # install patched vif.py for libvirt
 # https://bugs.launchpad.net/nova/+bug/1893263
```
comay, line 246: This reference needs to be updatde to the Launchpad bug covering the vCPUs > 1 condition.

chef/cookbooks/bcpc/recipes/calico-work.rb
```diff
@@ -52,11 +52,18 @@
...
-cookbook_file '/usr/lib/python2.7/dist-packages/neutron/agent/linux/dhcp.py' do
+cookbook_file '/usr/lib/python3/dist-packages/neutron/agent/linux/dhcp.py' do
   source 'neutron/dhcp.py'
   notifies :restart, 'service[calico-dhcp-agent]', :immediately
 end
 
+# install patched etcdv3.py for networking-calico
+# https://github.com/projectcalico/networking-calico/pull/58
+cookbook_file '/usr/lib/python3.6/dist-packages/networking_calico/etcdv3.py' do
+  source 'calico/etcdv3.py'
```
comay, line 63: Don't you want the `pycompile(1)` step included here (as on `neutron-head.rb`)?

chef/cookbooks/bcpc/recipes/nova-compute.rb
tj90241: The intent indicated to me was that the same LP will be used for this follow-on fix.

comay: I know that sometimes an existing LP bug is reused that way but perhaps @desrod can comment on whether the folks at Canonical expect to do so here. I tend to think a new LP is preferable once the original fix has been back-ported and made available on other releases.

chef/cookbooks/bcpc/recipes/calico-work.rb
comay: Ah, thanks for that clarification.

chef/cookbooks/bcpc/recipes/nova-compute.rb
```diff
@@ -35,11 +35,16 @@
...
+  nova-compute-kvm
+  cpu-checker
+  libvirt-daemon-driver-storage-rbd
   nova-api-metadata
   ovmf
   pm-utils
   sysfsutils
-)
+) do
+  options '--no-install-recommends'
```
comay, line 46: So which package is the offender in pulling in those extraneous packages you mentioned?

chef/cookbooks/bcpc/attributes/placement.rb
```diff
@@ -0,0 +1,19 @@
...
+default['bcpc']['placement']['db']['max_overflow'] = 128
+default['bcpc']['placement']['db']['max_pool_size'] = 64
+
+# Placement debug toggle
+default['bcpc']['placement']['debug'] = false
+
+# "workers" parameters in nova are set to number of CPUs
+# available by default. This provides an override.
+default['bcpc']['placement']['workers'] = nil
+default['bcpc']['placement']['workers'] = nil
```
comay, line 16: Duplicate definition?

chef/cookbooks/bcpc/recipes/placement.rb
```diff
@@ -0,0 +1,265 @@
...
+template '/etc/haproxy/haproxy.d/placement.cfg' do
+  source 'placement/haproxy.cfg.erb'
+  variables(
+    headnodes: headnodes(all: true),
+    vip: node['bcpc']['cloud']['vip']
+  )
+  notifies :reload, 'service[haproxy-placement]', :immediately
+end
+
+# placement package installation and service defintiion
```
comay, line 137: `s/defintiion/definition/`

chef/cookbooks/bcpc/recipes/placement.rb
```diff
@@ -0,0 +1,265 @@
...
+package 'placement-api'
```
comay, line 138: Nit but could you add a blank line to separate the `package` & `service` definition(s)?

chef/cookbooks/bcpc/recipes/placement.rb
```diff
@@ -0,0 +1,265 @@
...
+service 'haproxy-placement' do
+  service_name 'haproxy'
+end
+
+# create policy.d dir for policy overrides
+directory '/etc/placement/policy.d' do
+  action :create
+end
+
+# TODO: @tstachecki: differs from /etc/placement/placement.conf?
```
comay, line 152: So I noticed that the ordering of things here differs from what was in `nova-head.rb`. In particular, I would have expected all of the database work to be done first before setting up the Placement service itself since the latter requires the former before the service is restarted.

chef/cookbooks/bcpc/templates/default/placement/placement-create-db.sql.erb
```diff
@@ -0,0 +1,17 @@
...
+
+GRANT ALL PRIVILEGES ON <%= @db['dbname'] %>.* \
+  TO '<%= @db['username'] %>'@'localhost' \
+  IDENTIFIED BY '<%= @db['password'] %>';
+
+GRANT ALL PRIVILEGES ON <%= @db['dbname'] %>.* \
+  TO '<%= @db['username'] %>'@'%' \
+  IDENTIFIED BY '<%= @db['password'] %>';
+
+GRANT ALL PRIVILEGES ON placement.* \
```
comay, line 11: Aren't lines 10-17 already covered by the previous two stanzas?

chef/cookbooks/bcpc/templates/default/placement/placement.conf.erb
```diff
@@ -0,0 +1,35 @@
+# for an overview of all the available configuration options for placement@stein
+# go to: https://docs.openstack.org/placement/stein/configuration/config.html
```
comay, line 2: `s/stein/ussuri/`

Also as a bonus, it would be nice if all of the references to `rocky` in the other template files were also switched to `ussuri` but that can be a follow up if you prefer.

chef/cookbooks/bcpc/recipes/ceph-mon.rb
```diff
@@ -184,3 +184,8 @@
     not_if "ceph osd tree | grep #{rack}"
   end
 end
+
+# clusters do not respect ceph.conf changes at this time
```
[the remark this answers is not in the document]
tj90241: I mentioned it in the attributes rather than here:
https://github.com/bloomberg/chef-bcpc/blob/ussuri/chef/cookbooks/bcpc/attributes/ceph.rb#L14

Essentially, there is a CVE pertaining to reuse of a `global_id` access token.  Enabling it too early can have significant consequences on clusters which are upgraded from Mimic, but which have not yet bounced their `qemu-*` processes yet via a live-migration or some other means:
_When this option is set to false (emphasis mine: when patching the vuln), then an unpatched client will not be able to reconnect to the cluster after an intermittent network disruption breaking its connect to a monitor, or be able to renew its authentication ticket when it times out (by default, after 72 hours)._

I'm not sure why one needs to run this command to actually get the changes to stick.  I would set `ceph.conf` without the setting and observe that it changed individually on each daemon via `ceph daemon ... config show`, but the warning about it being enabled potentially would persist until the `ceph config set...` (globally) was applied as per the CVE.

chef/cookbooks/bcpc/recipes/keystone.rb
```diff
@@ -210,11 +210,6 @@
   DOC
 end
 
-# use multi-domain keystone policy
```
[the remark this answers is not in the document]
tj90241: I believe so, but I have not tested it as I do not believe it is used by any deployment.
ref this LP though: https://bugs.launchpad.net/keystone/+bug/1806762

"_Ultimately, the policy.v3cloudsample.json policy file should be removed because it is obsolete with the advent of system-scope [0] and default roles [1]._"

chef/cookbooks/bcpc/recipes/nova-compute.rb
tj90241: `qemu-system-x86`, as of... Stein?
```
Recommends: qemu-system-gui (= 1:4.0+dfsg-0ubuntu9.8~cloud0), qemu-utils, ovmf, cpu-checker
Suggests: samba, vde2, qemu-block-extra (= 1:4.0+dfsg-0ubuntu9.8~cloud0), sgabios
```

Compare to Rocky:
```
Recommends: qemu-utils, cpu-checker
```

The `qemu-system-gui` package starts pulling in a dizzying amount of X11, GTK, etc. packages...

We can possibly remove the `options 'no-install-recommends'` if that's what you're insinuating -- this predates this commit being merged: https://github.com/bloomberg/chef-bcpc/commit/bceaa95e037d531ba101b2d74e2bcd28d827b91f

chef/cookbooks/bcpc/recipes/nova-head.rb
```diff
@@ -210,20 +137,20 @@
...
   nova-api
   nova-conductor
-  nova-consoleauth
   nova-novncproxy
-  nova-placement-api
   nova-scheduler
-)
+  novnc
+) do
+  options '--no-install-recommends'
```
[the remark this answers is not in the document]
tj90241: I can't recall the reason I had put this here offhand - I do recall bumping into some issue about `python3-novnc` not being a thing and the fact that this package is still python2-based.  I'll try to remove both packages and see if the build/upgrades are clean still.

chef/cookbooks/bcpc/attributes/placement.rb
tj90241: Fixing, thanks.

chef/cookbooks/bcpc/recipes/placement.rb
tj90241: yep, adding.

chef/cookbooks/bcpc/templates/default/placement/placement-create-db.sql.erb
tj90241: oof, yeah, this is some intentional copy/paste fallout.  Fixing.

chef/cookbooks/bcpc/templates/default/placement/placement.conf.erb
tj90241: Fixing everything that comes up in a grep.

chef/cookbooks/bcpc/recipes/ceph-mon.rb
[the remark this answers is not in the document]
comay: Thanks for the explanation. I understand about the CVE itself but I was curious why `ceph-mon` actually required the command to be run rather than picking up the setting in `ceph.conf` when they're restarted above. Perhaps it's just cosmetic and the warning isn't really accurate.

Resolving this but perhaps the comment could be updated to indicate we don't know why this is necessary as in

`# For some reason, Ceph doesn't honor the setting in ceph.conf at this time so manually set the value as well`

chef/cookbooks/bcpc/recipes/nova-compute.rb
comay: Yes, I was hoping we could remove the `--no-install-recommends` but I realized that our bare metal systems don't have `/etc/apt/apt.conf.d/99no-install-recommends` installed on them. 😭

I wonder if the following might work based on some reading

```
package %w(
  ceph
  nova-compute
  qemu-system-gui-
  nova-api-metadata
  ovmf
  pm-utils
  sysfsutils
)
```

At least on the command line, the `qemu-system-gui-` prevents that package (and its dependencies) from being installed.

tj90241: I find the `no-install-recommends` approach to be much more self-documenting for when we are able to drop this.

comay: At the very least, I'd like to see the packages that are covered by explicit dependencies dropped from the above list, namely `nova-compute-kvm`.

chef/cookbooks/bcpc/recipes/nova-head.rb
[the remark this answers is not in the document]
comay: I just ran an experiment with Ussuri and indeed `python-novnc` & `novnc` are installed even if `--no-install-recommends` is supplied.

chef/cookbooks/bcpc/recipes/placement.rb
comay: It looks like we're not consistent - in particular, the Cinder and Nova recipes have the database creation taking place before rendering the Apache and `service.conf` file while the others reverse that. Since `placement.rb` was birthed presumably from `nova-head.rb`, I guess I expected the ordering to resemble that.

It's true the services do start up after the package installation finishes (which I would argue is a bug given that the default configuration delivered by the package is undoubtedly not functional) but I was hoping to reduce the number of "false" (failed) restarts that occur when these services get restarted while things are put into place.

chef/cookbooks/bcpc/recipes/nova-head.rb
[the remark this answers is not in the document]
tj90241: yes, those two were carry-overs from the upgrade process, which requires them to be present to avoid breaking the VNC capabilities when purging packages as part of an in-place upgrade.  It can be removed from the build.

When I remove `no-install-recommends`, though, the following additional packages are present (on, at least, the headnodes):
```
+adwaita-icon-theme
+at-spi2-core
+cpu-checker
+dconf-gsettings-backend:amd64
+dconf-service
+docutils-doc
+fontconfig
+fontconfig-config
+fonts-dejavu-core
+glib-networking:amd64
+glib-networking-common
+glib-networking-services
+gsettings-desktop-schemas
+gstreamer1.0-plugins-base:amd64
+gstreamer1.0-plugins-good:amd64
+gstreamer1.0-x:amd64
+gtk-update-icon-cache
+hicolor-icon-theme
+humanity-icon-theme
+ipxe-qemu
+ipxe-qemu-256k-compat-efi-roms
+libaa1:amd64
+libasound2:amd64
+libasound2-data
+libasyncns0:amd64
+libatk-bridge2.0-0:amd64
+libatk1.0-0:amd64
+libatk1.0-data
+libatspi2.0-0:amd64
+libavc1394-0:amd64
+libbrlapi0.6:amd64
+libcaca0:amd64
+libcacard0:amd64
+libcairo-gobject2:amd64
+libcairo2:amd64
+libcdparanoia0:amd64
+libcolord2:amd64
+libcroco3:amd64
+libdatrie1:amd64
+libdconf1:amd64
+libdv4:amd64
+libepoxy0:amd64
+libfdt1:amd64
+libflac8:amd64
+libfontconfig1:amd64
+libgbm1:amd64
+libgdk-pixbuf2.0-0:amd64
+libgdk-pixbuf2.0-bin
+libgdk-pixbuf2.0-common
+libgraphite2-3:amd64
+libgstreamer-plugins-base1.0-0:amd64
+libgstreamer-plugins-good1.0-0:amd64
+libgstreamer1.0-0:amd64
+libgtk-3-0:amd64
+libgtk-3-bin
+libgtk-3-common
+libgudev-1.0-0:amd64
+libharfbuzz0b:amd64
+libiec61883-0:amd64
+libjack-jackd2-0:amd64
+libjson-glib-1.0-0:amd64
+libjson-glib-1.0-common
+libmp3lame0:amd64
+libmpg123-0:amd64
+libogg0:amd64
+libopus0:amd64
+liborc-0.4-0:amd64
+libpango-1.0-0:amd64
+libpangocairo-1.0-0:amd64
+libpangoft2-1.0-0:amd64
+libpciaccess0:amd64
+libpcsclite1:amd64
+libpixman-1-0:amd64
+libpmem1
+libproxy1v5:amd64
+libpulse0:amd64
+libraw1394-11:amd64
+librest-0.7-0:amd64
+librsvg2-2:amd64
+librsvg2-common:amd64
+libsamplerate0:amd64
+libshout3:amd64
+libslirp0:amd64
+libsndfile1:amd64
+libsoup-gnome2.4-1:amd64
+libsoup2.4-1:amd64
+libspeex1:amd64
+libspice-server1:amd64
+libtag1v5:amd64
+libtag1v5-vanilla:amd64
+libthai-data
+libthai0:amd64
+libtheora0:amd64
+libtwolame0:amd64
+libusbredirparser1:amd64
+libv4l-0:amd64
+libv4lconvert0:amd64
+libvirglrenderer0
+libvirt-daemon
+libvirt-daemon-driver-qemu
+libvirt-daemon-driver-storage-rbd
+libvisual-0.4-0:amd64
+libvorbis0a:amd64
+libvorbisenc2:amd64
+libvpx5:amd64
+libvte-2.91-0:amd64
+libvte-2.91-common
+libwavpack1:amd64
+libwayland-client0:amd64
+libwayland-cursor0:amd64
+libwayland-egl1:amd64
+libwayland-server0:amd64
+libx11-6:amd64
+libx11-data
+libxcb-render0:amd64
+libxcb-shm0:amd64
+libxcb1:amd64
+libxcomposite1:amd64
+libxcursor1:amd64
+libxdamage1:amd64
+libxext6:amd64
+libxfixes3:amd64
+libxi6:amd64
+libxinerama1:amd64
+libxkbcommon0:amd64
+libxml2-utils
+libxrandr2:amd64
+libxrender1:amd64
+libxtst6:amd64
+libxv1:amd64
+msr-tools
+ovmf
+python-chardet
+python-olefile
+python-pil:amd64
+python-pygments
+qemu-kvm
+qemu-system-common
+qemu-system-data
+qemu-system-gui:amd64
+qemu-system-x86
+seabios
+ubuntu-mono
+x11-common
```

So I will leave `no-install-recommends`, but remove the other two, so that `qemu-system-x86`, X11 components, etc. are not pulled in on the headnodes.

chef/cookbooks/bcpc/recipes/nova-compute.rb
tj90241: Removed `nova-compute-kvm` from the list and verified build.

chef/cookbooks/bcpc/recipes/placement.rb
tj90241: Discussed and opened a task to generally address this and the false starts that occur.

chef/cookbooks/bcpc/recipes/cinder.rb
comay: Thanks for making this update.

comay (dismissed):
Great work, Tyler - this is a monumental improvement to the project.

chef/cookbooks/bcpc/recipes/nova-compute.rb
tj90241: A new LP was raised this AM after all: https://bugs.launchpad.net/nova/+bug/1939604
Will toss it in the comments shortly.

comay (dismissed):

PR #1880, opened 2021-07-29
netplan: Drop match: macaddress from configuration

tj90241:
Specifying the MAC address of the transit interfaces can
create operational hurdles when NICs are replaced. As
we use deterministic network interface names vis-a-vi
`systemd`, we do not need to also match on the MAC
address.


tj90241:
I suppose this is fine for 8.2, but let me know if someone feels this should come later.

comay (approved):

justinjpacheco (approved):

PR #1881, opened 2021-08-10
`nova-api-metadata` can use a local `memcached` on the hypervisor

comay:

This proposed change adds `memcached` to worknodes and sets their configuration to use the local server for non-Keystone related objects, in particular Nova metadata. This is safe to do as `nova-api-metadata` when run directly on hypervisors should never be querying data except for their own instances.


tj90241 (approved):
Seems like a future item might be to break memory limit down by headnode and worknode.

justinjpacheco (approved):

PR #1890, opened 2021-08-25
Make ProxySQL user updates more reliable

mboruta1:
Previously a flag was set in the ProxySQL recipe when user attributes
needed to be updated, and read by subsequent recipes. This would require
a full chef run be completed every time a user attribute was changed,
otherwise subsequent chef runs would not have the said flag set. This
change makes it so that each recipe checks whether its ProxySQL user
needs to be updated.

**Describe your changes**

**Testing performed**

**Additional context**


tj90241 (changes requested):
The `vars-user.erb` file can be deleted now from this repository right?
Likewise, should we delete it from the filesystem(s) now that it's no longer referenced?

mboruta1:
Thank you Tyler for your review! Indeed we should delete vars-users from the repo; I also added a step at the top of the recipe to delete said file if present along with a note to remove said chef resource in the future once enough time has passed. Lastly, I added the `server_version` variable to the ProxySQL configuration file and attributes.

tj90241 (approved):
Thanks - looks great! 🎉

comay (approved):
Looks good - just a nit that you're very consistent using the proper noun ProxySQL but we should afford the same in the comments for the name MySQL. 😉

mboruta1:
Thank you David, updated all occurrences of `mysql` and `proxysql` in comments across the project accordingly :).

justinjpacheco (approved):

PR #1895, opened 2021-08-31
ceph: Add option to disable progress module

tj90241:
Reintroducing the ability to disable the progress module;
this second cut fixes a race condition that occurred when
bootstrapping a cluster.


tj90241:
Having a thought about a better way to guard/check for this, but this might have to do.

chef/cookbooks/bcpc/recipes/ceph-mgr.rb
```diff
@@ -36,3 +36,12 @@
   command "ceph-deploy mgr create #{node['hostname']}"
   creates "/var/lib/ceph/mgr/ceph-#{node['hostname']}/done"
 end
+
+execute 'configure progress module' do
+  command "ceph progress #{node['bcpc']['ceph']['module']['progress']['enabled'] ? 'on' : 'off'}"
+  not_if "ceph config get mgr mgr/progress/enabled | grep -w #{node['bcpc']['ceph']['module']['progress']['enabled']}"
+
+  # Guard to avoid racing for module configuration prior before mgrs are fully up
```
comay, line 44: I assume there isn't some sort of status you can check (maybe using `ceph config get mgr`?) to see if Ceph is ready?

tj90241: I was thinking about `ceph mgr module ls`, and means to query if the mgr is present in the cluster.  But do those necessarily mean that the mgr is ready to accept configuration commands?  (this is kind of where I'm stuck thinking, and wondering if it's best to just spin).

mboruta1 (approved):
Not ideal to spin but if there is no other clean option then I do not think we have a choice. Did something similar in ProxySQL. Looks good to me :).

PR #1898, opened 2021-09-09
configurable retry timeout for watcher's vm migration

ajaytikoo:
Signed-off-by: Ajay Tikoo <atikoo@bloomberg.net>

**Describe your changes**

> NOTE: This replaces earlier [PR: 1874](https://github.com/bloomberg/chef-bcpc/pull/1874)

When migrating an instance, Watcher polls to check the status of the migrating instance for 120 seconds. If migration does not complete (or fail) within this time, Watcher considers that the migration failed, does not try to migrate any pending instances, and marks the action plan as failed. In a large production environment, it generally takes longer time for a migration to complete. In order to avoid Watcher failing a migration prematurely, this change makes the timeout and poll interval a configurable parameter, which defaults to the same values as are right now hard coded in the code (120 seconds and 1 second respectively). In environments where it normally takes longer to live migrate instances, Watcher can be configured with higher timeouts so that it will not consider a migration failed unless it actually fails or takes longer than the configured time to migrate.

Changes in the patched files:

- conf/init.py: the changes to this file are two additional lines, one importing nova_helper from watcher.conf, and other registering its config options.
- conf/nova_helper.py: this is a new file, which implements the configuration items for nova_helper.
- common/nova_helper.py: the changes in this file are within the migration functions (live and non-live)

**Testing performed**

**Additional context**

### DIFF
--- a/chef/cookbooks/bcpc/recipes/watcher.rb
+++ b/chef/cookbooks/bcpc/recipes/watcher.rb
@@ -139,6 +139,52 @@ watcher_processes = if !node['bcpc']['watcher']['api_workers'].nil?
                       node['bcpc']['openstack']['services']['workers']
                     end
 
+# install patched conf/nova_helper.py
+# added configuration options for nova instance migration timeout and
+# polling interval. Default is 120 seconds timeout and 1 second polling
+# interval.
+cookbook_file '/usr/lib/python3/dist-packages/watcher/conf/nova_helper.py' do
+  source 'watcher/conf/nova_helper.py'
+  notifies :run, 'execute[pycompile-nova-helper-conf]', :immediately
+  notifies :restart, 'service[watcher-applier]', :delayed
+  notifies :restart, 'service[watcher-decision-engine]', :delayed
+end
+
+execute 'pycompile-nova-helper-conf' do
+  action :nothing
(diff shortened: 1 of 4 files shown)


chef/cookbooks/bcpc/recipes/watcher.rb
```diff
@@ -139,6 +139,52 @@
...
+
+execute 'pycompile-conf-init-file' do
+  action :nothing
+  command 'python3 -m py_compile /usr/lib/python2.7/dist-packages/watcher/conf/__init__.py'
+end
+
+# install patched nova_helper.py
+# uses configuration options for nova instance migration timeout and polling
+# interval.
+cookbook_file '/usr/lib/python2.7/dist-packages/watcher/common/nova_helper.py' do
```
tj90241, line 176: still references python2 - this code won't take effect... can you please also use `py3compile` vs. `python3 -m py_compile` to remain consistent with the rest of the recipes?

ajaytikoo: I will fix this in the next commit. Thanks for pointing this out.

comay:
Closing this as I don't believe we'll be making this change.

PR #1901, opened 2021-09-13
Various improvements to the ProxySQL automation

mboruta1:
- Allow in-place upgrades of ProxySQL
- When re-installing ProxySQL, back up the existing datadir
- Have consul switch the ProxySQL primary not only when ProxySQL is not
reachable, but also when a MySQL backend is not reachable via ProxySQL
- Expose several more ProxySQL configuration variables as chef attributes
- Increase the amount of time ProxySQL monitor logs are stored
- When uninstalling ProxySQL the process is never terminated; terminate
it at the end of a chef run
- Set fast_forward to false as default
- Updated comments

**Describe your changes**
See the description above.

**Testing performed**
Manual testing of additions on a local BVE + internal Jenkins pipeline.

**Additional context**

### DIFF
--- a/chef/cookbooks/bcpc/attributes/proxysql.rb
+++ b/chef/cookbooks/bcpc/attributes/proxysql.rb
@@ -52,11 +52,18 @@ default['bcpc']['proxysql']['server_capabilities'] = 569899
 default['bcpc']['proxysql']['enable_client_deprecate_eof'] = false
 default['bcpc']['proxysql']['enable_server_deprecate_eof'] = false
 default['bcpc']['proxysql']['multiplexing'] = true
+default['bcpc']['proxysql']['connect_timeout_server'] = 1000
+default['bcpc']['proxysql']['connect_timeout_server_max'] = 10000
+default['bcpc']['proxysql']['connect_retries_on_failure'] = 10
+default['bcpc']['proxysql']['connect_retries_delay'] = 10
+default['bcpc']['proxysql']['shun_on_failures'] = 5
+default['bcpc']['proxysql']['shun_recovery_time_sec'] = 10
 default['bcpc']['proxysql']['free_connections_pct'] = 10
 default['bcpc']['proxysql']['long_query_time'] = 10000
 default['bcpc']['proxysql']['default_query_timeout'] = 86400000
 default['bcpc']['proxysql']['max_transaction_time'] = 14400000
 default['bcpc']['proxysql']['query_retries_on_failure'] = 1
+default['bcpc']['proxysql']['monitor_history'] = 3600000
 
 # Galera Hostgroup Variables
 
@@ -91,8 +98,9 @@ default['bcpc']['proxysql']['mysql_servers']['compression'] = 0
 # The maximum number of connections ProxySQL will establish to each backend
 default['bcpc']['proxysql']['mysql_servers']['max_connections'] = node['bcpc']['mysql']['max_connections']
 
-# If greater than 0 and a backend's replication lag surpasses the given
-# threshold it will be shunned until it catches up.
+# If greater than 0, and a backend's replication lag surpasses the given
+# threshold, it will be shunned until it catches up. This is set to 0
+# (disabled) in favor of max_transactions_behind.
 default['bcpc']['proxysql']['mysql_servers']['max_replication_lag'] = 0
 
 # Whether or not to enable SSL on ProxySQL-MySQL connections
@@ -117,7 +125,7 @@ default['bcpc']['proxysql']['mysql_users']['transaction_persistent'] = 0
 
 # If set queries bypass the query processing layer and are sent directly to the
 # backend server
-default['bcpc']['proxysql']['mysql_users']['fast_forward'] = 1
+default['bcpc']['proxysql']['mysql_users']['fast_forward'] = 0
 
 # The maximum number of allowable frontend connections for a specific user
 default['bcpc']['proxysql']['mysql_users']['max_connections'] = 32768
--- a/chef/cookbooks/bcpc/recipes/proxysql.rb
+++ b/chef/cookbooks/bcpc/recipes/proxysql.rb
@@ -32,10 +32,30 @@ package 'remove proxysql2' do
 end
 
 # Remove ProxySQL if it is not enabled
+# NOTE: As of 2.2.2 uninstalling ProxySQL does not actually stop the ProxySQL
+# service, hence the delayed call to 'kill proxysql'.
+# NOTE: /etc/proxysql-admin.cnf is deleted manually but the data dir is kept
+# in order to ease recovery, for future log analysis, etc.
 package 'remove proxysql' do
   package_name 'proxysql'
   action :purge
   not_if { node['bcpc']['proxysql']['enabled'] }
+  notifies :run, 'execute[kill proxysql]', :delayed
+  notifies :delete, 'file[proxysql-admin]', :delayed
+end
+
+# Kill ProxySQL iff ProxySQL was uninstalled. Its execution is delayed to the
+# end of the chef run, after all relevant services have been switched over to
+# MySQL.
+execute 'kill proxysql' do
+  command 'killall proxysql || true'
+  action :nothing
+end
+
+# Remove the ProxySQL admin file on package purge
+file 'proxysql-admin' do
+  path '/etc/proxysql-admin.cnf'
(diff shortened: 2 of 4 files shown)


chef/cookbooks/bcpc/attributes/proxysql.rb
```diff
@@ -52,11 +54,18 @@
 default['bcpc']['proxysql']['enable_client_deprecate_eof'] = false
 default['bcpc']['proxysql']['enable_server_deprecate_eof'] = false
 default['bcpc']['proxysql']['multiplexing'] = true
+default['bcpc']['proxysql']['connect_timeout_server'] = 1000
+default['bcpc']['proxysql']['connect_timeout_server_max'] = 10000
+default['bcpc']['proxysql']['connect_retries_on_failure'] = 10
+default['bcpc']['proxysql']['connect_retries_delay'] = 10
```
mboruta1, line 60: The variables in this block are all ProxySQL defaults except for this one, which has been increase from 1 ms to 10.

chef/cookbooks/bcpc/attributes/proxysql.rb
```diff
@@ -52,11 +54,18 @@
...
+default['bcpc']['proxysql']['shun_on_failures'] = 5
+default['bcpc']['proxysql']['shun_recovery_time_sec'] = 10
 default['bcpc']['proxysql']['free_connections_pct'] = 10
 default['bcpc']['proxysql']['long_query_time'] = 10000
 default['bcpc']['proxysql']['default_query_timeout'] = 86400000
 default['bcpc']['proxysql']['max_transaction_time'] = 14400000
 default['bcpc']['proxysql']['query_retries_on_failure'] = 1
+default['bcpc']['proxysql']['monitor_history'] = 3600000
```
mboruta1, line 68: Increase default monitor history from 10 minutes to 1 hour to help debugging unexpected outages. Was hesitant to increase this further (e.g. 1 day) since the ping log table has an entry for every 8 seconds, which would result in 32k entries in the `mysql_server_ping_log` table (~10k for each backend). Can easily increase it later.

chef/cookbooks/bcpc/attributes/proxysql.rb
```diff
@@ -117,7 +127,7 @@
 
 # If set queries bypass the query processing layer and are sent directly to the
 # backend server
-default['bcpc']['proxysql']['mysql_users']['fast_forward'] = 1
+default['bcpc']['proxysql']['mysql_users']['fast_forward'] = 0
```
mboruta1, line 130: Enable the query processor by default; in fast forward mode ProxySQL is no longer SQL protocol aware and at that point you really are better off using HAProxy (see ProxySQL's own [blog entry](https://proxysql.com/blog/point-select-benchmark/)).

chef/cookbooks/bcpc/recipes/proxysql.rb
```diff
@@ -32,10 +32,30 @@
...
+
+# Kill ProxySQL iff ProxySQL was uninstalled. Its execution is delayed to the
+# end of the chef run, after all relevant services have been switched over to
+# MySQL.
+execute 'kill proxysql' do
+  command 'killall proxysql || true'
+  action :nothing
+end
+
+# Remove the ProxySQL admin file on package purge
```
mboruta1, line 55: The only files remaining after the ProxySQL uninstall are those in the data-dir, which are kept for debugging purposes (the recipe moves them to a proper location if ProxySQL is installed again, thus giving it a completely clean slate.

chef/cookbooks/bcpc/templates/default/consul/proxysql-check-helper.erb
```diff
@@ -14,10 +14,41 @@
...
+# Check connectivity via ProxySQL to a backend. Ideally this check will only
+# fail when all MySQL backends have encountered an issue, and thus switching
+# the primary will serve no purpose. However, as described in
+# https://github.com/sysown/proxysql/issues/3618, ProxySQL has at least one
+# bug that would require such a move.
+#
+# This check is forgiving because in a vast majority of cases ProxySQL will
+# handle backend failures just fine. It is designed to catch scenarios like
+# the one described in the bug report linked to above.
+MAX_ATTEMPTS=3
```
mboruta1, line 39: Consul still switches to another primary if this script (or the corresponding mysql check script) does not return in 2s. Consul's timeout can be increased accordingly if needed.

chef/cookbooks/bcpc/recipes/proxysql.rb
```diff
@@ -104,15 +144,15 @@
...
-# Delete the ProxySQL configuration DB
-execute 'delete default proxysql config db' do
-  command "rm -f #{node['bcpc']['proxysql']['default_datadir']}/proxysql.db"
-  only_if { node.run_state['proxysql_fresh_install'] }
+# Before ProxySQL is installed, move (and thus back up) the existing ProxySQL
+# data directory, if any.
+execute 'backup existing proxysql data directory' do
+  command "[ -d \"#{node['bcpc']['proxysql']['default_datadir']}\" ] && \
+    mv #{node['bcpc']['proxysql']['default_datadir']}/ \
+      #{node['bcpc']['proxysql']['default_datadir_parent_dir']}/proxysql-$(date +%s)"
```
tj90241, line 152: Could `::File.dirname` be used here instead of defining the attribute for parent directory?

chef/cookbooks/bcpc/recipes/proxysql.rb
```diff
@@ -32,10 +32,30 @@
...
   not_if { node['bcpc']['proxysql']['enabled'] }
+  notifies :run, 'execute[kill proxysql]', :delayed
+  notifies :delete, 'file[proxysql-admin]', :delayed
+end
+
+# Kill ProxySQL iff ProxySQL was uninstalled. Its execution is delayed to the
+# end of the chef run, after all relevant services have been switched over to
+# MySQL.
+execute 'kill proxysql' do
+  command 'killall proxysql || true'
```
tj90241, line 51: Alternatively, may be able to use `/var/lib/proxysql/proxysql.pid` to kill a specific PID (and then remove the state file).  I'm not sure it offers any advantages of this solution, though.

chef/cookbooks/bcpc/recipes/proxysql.rb
```diff
@@ -104,15 +144,15 @@
...
+# data directory, if any.
+execute 'backup existing proxysql data directory' do
+  command "[ -d \"#{node['bcpc']['proxysql']['default_datadir']}\" ] && \
+    mv #{node['bcpc']['proxysql']['default_datadir']}/ \
+      #{node['bcpc']['proxysql']['default_datadir_parent_dir']}/proxysql-$(date +%s)"
+  action :nothing
 end
 
-# Declare the service resource
-service 'proxysql'
```
tj90241, line 114: Just curious -- what's the reason for undermining the `service` resource?

chef/cookbooks/bcpc/recipes/proxysql.rb
mboruta1: Yes, that's true, and I also do not see any advantage. If I went the PID route I would have to trust that the file contents are accurate, while doing a killall requires me to trust that no other non-proxysql processes with the name `proxysql` exist on the system. Both are likely to be the case, but I think the former is slightly less likely (proxysql started manually for some reason, pid file never cleaned up for some reason).

tj90241: Sounds good -- just thought I'd mention it!

chef/cookbooks/bcpc/recipes/proxysql.rb
mboruta1: This resource does not do anything by itself and is not referenced by anything, so no need to declare/define it here. Forgot to clean it up some time before.

chef/cookbooks/bcpc/recipes/proxysql.rb
mboruta1: Yes, I think so; let me play around with chef a bit and if all is well push out the changes and remove the proposed `node['bcpc']['proxysql']['default_datadir_parent_dir']` attribute.

mboruta1: Updated to use `File.dirname`, thank you!

tj90241 (approved):

PR #1902, opened 2021-09-13
Add initial version of Github Actions Workflow.

psipika:
*Issue number of the reported bug or feature request: N/A *

This changeset introduces code migrating this repository away from TravisCI and into Github Actions.
The contents of the file being introduced (the Actions workflow in `.github/workflows/run_linters.yml`) are described, below:

- Define the name and list events upon which this workflow will be triggered (`push` [on any branch](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-a-single-event)).

```yaml
name: run_linters
on: [push]
```

- List the jobs within this workflow (there will be only one)

```yaml
jobs:
```

- The `run_linters` job will be executed on a *runner* with the latest version of Ubuntu.

```yaml
  run-linters:
    runs-on: ubuntu-latest
```

- The steps executed as part of the `run_linters` job will use the `v2` version of the `checkout` Action (provided by the Github community).

```yaml
    steps:
      - uses: actions/checkout@v2
```

- The steps will then call `install_linters.sh` and `run_linters.sh` from the `.github/scripts` directory.

```yaml
      - run: .github/scripts/install_linters.sh
      - run: .github/scripts/run_linters.sh
```

Testing includes submitting this Pull Request and monitoring the Action workflow as per [this description](https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions#viewing-the-jobs-activity).
In this case, the Actions' workflow can be seen [here](https://github.com/psipika/chef-bcpc/actions).

Github Actions reference material:
- [Understanding Github Actions](https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions).
- [Github Actions Workflow Syntax](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions).


psipika:
Will recreate based on comments to be a branch against `development`.

PR #1909, opened 2021-09-28
Install ansible via pip and use ansible-lint standalone.

psipika:
To continue work on Github Actions from PR #1903, this changeset uses code suggested in PR #1900 to introduce the following:
- During the linter installation step,  `ansible` is added to the `pip install` step. This is to prevent errors [like this one](https://github.com/bloomberg/chef-bcpc/runs/3722833869#step:4:27).
- Durint the linter run, move `ansible-lint` outside of the original `find` command (as suggested [here](https://github.com/bloomberg/chef-bcpc/pull/1900/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R38-R39)).

In addition, continuing cleanup of linting errors is the ultimate goal and so all contributors are requested to contribute to this changeset.

Manually running `.github/scripts/run_linters.sh` is the suggested testing to be performed prior to making a commit.


chef/cookbooks/bcpc/files/default/nova/vif.py
```diff
@@ -268,7 +268,7 @@ def _get_virtio_mq_settings(self, image_meta, flavor):
             max_tap_queues = self._get_max_tap_queues()
             if max_tap_queues:
                 vhost_queues = (max_tap_queues if flavor.vcpus > max_tap_queues
-                    else flavor.vcpus)
+                                else flavor.vcpus)
```
justinjpacheco, line 271: this seems indented too much

psipika: The `else` corresponds to the `if flavor.vcpus > max_tab_queues` "inline" `if` on line 270, unless I'm misreading the parentheses.

Here's the lint report:
```
./chef/cookbooks/bcpc/files/default/nova/vif.py:271:21: E128 continuation line under-indented for visual indent
```

PR #1920, opened 2021-11-01
move consul from chef to ansible.

npustchi:
Adding consul role to replace consul chef recipes, attributes, and templates.
Currently only conul-package is being added to ansible.
recipes:
- consul-package.rb
- consul.rb

attributes:
- consul.rb

templates:
- cloud-ip-watch.erb
- consul.service.erb
- if_leader.erb
- if_not_primary_mysql.sh.erb
- if_not_primary_proxysql.sh.erb
- if_primary_mysql.sh.erb
- if_primary_proxysql.sh.erb
- proxysql-check-helper.erb
- service-default.erb
- service-elect-watch.erb
- service-watch.erb

Tests:
I tested with 1h1w and 3h3w topologies in bve, built successfully. Also, I built 1h1w and 3h3w bve with chef in development and compared the configs generated, which is identical.
The failure (if the consul leader is any node than first node would fail chef, I tested with r3n1 being leader) prior to moving HAProxy to ansible is fixed as well.


ansible/playbooks/roles/consul/defaults/main.yml
```diff
@@ -0,0 +1,53 @@
+###############################################################################
+# consul files
+###############################################################################
+
+consul_filename: "\
+  {% for asset in assets_files %}\
```
tj90241, line 6: Just as a tip, instead of looping through, jinja has a lot of built-in filters -

```jinja
{{ assets_files | selectattr('name', '==', 'consul') | first }}"
```

ansible/playbooks/roles/consul/defaults/main.yml
```diff
@@ -0,0 +1,53 @@
...
+    {% if asset.name == 'consul' %}\
+      {{ asset.checksum }}\
+    {% endif %}\
+  {% endfor %}"
+
+executable: /usr/local/sbin/consul
+config_directory: /etc/consul/conf.d
+config_data_directory: /var/lib/consul
+bcpc_directory: /usr/local/bcpc/bin
+chef_databags_config_directory: /chef/databags/config.json
```
tj90241, line 22: This var is not referenced and can be deleted now, right?

ansible/playbooks/roles/consul/handlers/restart_consul.yml
```diff
@@ -0,0 +1,3 @@
+- name: restart consul
+  command: sudo systemctl restart consul.service
```
tj90241, line 2: You should use the `service` resource I think.  The linter will likely complain about using `systemctl` directly in the future.

```
- name: restart consul
  service:
    name: consul
    state: restarted
  become: yes
 ```

Comment (npustchi) [reply ansible/playbooks/roles/consul/handlers/restart_consul.yml:2]: yes, tried it, it fails, I think it is the permissions, searched online others had similar issues.

File: ansible/playbooks/roles/consul/defaults/main.yml (hunk 2 above)
Comment (npustchi) [reply ansible/playbooks/roles/consul/defaults/main.yml:22]: yap will delete it, thanks

File: ansible/playbooks/roles/consul/handlers/restart_consul.yml (hunk 3 above)
Comment (tj90241) [reply ansible/playbooks/roles/consul/handlers/restart_consul.yml:2]: Did you include `become: yes`?  Otherwise, it will execute as `operations` user and I'm not sure it has permissions to do that.

Comment (npustchi) [reply ansible/playbooks/roles/consul/handlers/restart_consul.yml:2]: yes. I did.

File: ansible/playbooks/roles/consul/defaults/main.yml (hunk 1 above)
Comment (npustchi) [reply ansible/playbooks/roles/consul/defaults/main.yml:6]: interesting will test it.

File: ansible/playbooks/roles/consul/handlers/restart_consul.yml (hunk 3 above)
Comment (npustchi) [reply ansible/playbooks/roles/consul/handlers/restart_consul.yml:2]: actually that is what I had and it kept failing, then I searched online and came up with this.

File: ansible/playbooks/roles/consul/templates/config.json.j2 [hunk 4]
```diff
@@ -0,0 +1,26 @@
...
+  "ports": {
+    "dns": {{ consul_dns_port }}
+  },
+  "recursors": [
+    "{{ cloud_ip }}"
+  ],
+  "bootstrap_expect": {{ ansible_play_hosts | length }},
+  "retry_join": [
+  {% for host in groups['headnodes']|difference([inventory_hostname]) %}
+  "{{ hostvars[host]['interfaces']['service']['ip'] }}"{{ "," if not loop.last }}
```
tj90241, line 23: Another jinja trick applicable here to avoid the `loop.last` and loop:
```jinja
groups['headnodes'] | difference([inventory_hostname]) | join(',')
```

tj90241 (approved):
lgtm, just approving with a few comments now as I'll be out the rest of this week.

ansible/playbooks/roles/consul/tasks/install-required-packages.yml
```diff
@@ -0,0 +1,4 @@
+- name: Install unzip
+  ansible.builtin.package:
```
psipika, line 2: Same thing here: is it possible to just use `package`?

ansible/playbooks/roles/consul/handlers/restart_consul.yml
psipika: I'm curious what the error message actually is, as a matter of learning ansible.

psipika (commented):
A few questions to check my understanding.

ansible/playbooks/roles/consul/defaults/main.yml
```diff
@@ -0,0 +1,52 @@
...
+    {% endif %}\
+  {% endfor %}"
+consul_checksum: "\
+  {% for asset in assets_files %}\
+    {% if asset.name == 'consul' %}\
+      {{ asset.checksum }}\
+    {% endif %}\
+  {% endfor %}"
+
+executable: /usr/local/sbin/consul
```
justinjpacheco, line 18: are these variables local to this role or are they global?

npustchi: it is local to this role but if we save the path /usr/local/sbin/ in base.yml might be better like https://github.com/bloomberg/chef-bcpc/blob/master/ansible/playbooks/roles/common/defaults/main/base.yml#L156

ansible/playbooks/roles/consul/tasks/install-required-packages.yml
npustchi: should work.

ansible/playbooks/roles/consul/handlers/start_consul.yml
```diff
@@ -0,0 +1,5 @@
+- name: start consul
+  ansible.builtin.service:
```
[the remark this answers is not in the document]
npustchi: should work.

ansible/playbooks/roles/consul/defaults/main.yml
npustchi: renamed the variables to consul prefix.

psipika (dismissed):

ansible/playbooks/roles/consul/tasks/install-consul.yml
```diff
@@ -0,0 +1,97 @@
...
+  register: temp_directory
+
+- name: unarchive consul
+  unarchive:
+    src: "{{ assets_download_dir }}/{{ consul_filename }}"
+    dest: "{{ temp_directory.path }}/"
+    creates: "{{ temp_directory.path }}/consul"
+    mode: 0755
+
+- name: install consul
```
mboruta1, line 14: Not an expert in ansible; but will this run every time the `configure-consul-cluster` target is run? If so, could we have some way of detecting whether or not what is present in the asset dir isn't already installed, and thus avoid an unnecessary re-installation and restart every time `make all` is run?

npustchi: it only installs if different,

PR #1925, opened 2021-11-18
cinder: Default rbd_exclusive_cinder_pool to true

tj90241:
As part of its periodic tasks, cinder polls each RBD
volume's size and accumulates a sum to report to the
cinder services. This is very inefficient.

Instead, if the Ceph pool(s) allocated to cinder are
guaranteed to be exclusively used by RBD volumes (and not
other Ceph objects), cinder can be told to just ask Ceph
for the size of the pool and be done with it.

This commit makes the necessary change, which is the
default in a future version of OpenStack/cinder anyways.


justinjpacheco (approved):

mboruta1 (approved):
+100

psipika (approved):

PR #1929, opened 2021-11-23
added powerdns enable/disable

justinjpacheco:


mboruta1 (approved):
Looks good to me; I am assuming that in our particular setup, when we flip this flag, we will uninstall powerdns manually?

justinjpacheco:
> Looks good to me; I am assuming that in our particular setup, when we flip this flag, we will uninstall powerdns manually?

@mboruta1 yep, that's correct

PR #1930, opened 2021-11-23
add custom neutron notification queue support

justinjpacheco:
- additional neutron notification queues can now be published to with
     neutron specific event information


mboruta1 (approved):
Looks good to me. The only thing I would change would be to indicate in the config that retry and transport_url are set to their defaults, and what those defaults are (retry and transport url, so for example `#retry = -1`). I am a fan of configs that specify defaults for important unset config options, like an ubuntu machines `/etc/ssh/sshd_config`, because I do not need to look elsewhere for the service's full configuration.

mboruta1 (dismissed):

mboruta1 (approved):

PR #1932, opened 2021-12-01
Pin version of ansible and ansible-lint.

psipika:

Pinning the version of `ansible-lint` and `ansible-core` to prevent issues [like this one](https://github.com/bloomberg/chef-bcpc/runs/4383150260?check_suite_focus=true#step:4:21).

Ran the fix on [my fork](https://github.com/psipika/chef-bcpc/runs/4384436341?check_suite_focus=true#step:4:20)


kpfleming:
As requested.

PR #1933, opened 2021-12-02
toggle powerdns forwarding

justinjpacheco:
- added support to toggle powerdns forwarding for the local zone

### DIFF
--- a/ansible/playbooks/roles/common/defaults/main/base.yml
+++ b/ansible/playbooks/roles/common/defaults/main/base.yml
@@ -295,3 +295,8 @@ system_parameters_net.nf_conntrack_max: 262144  # noqa var-naming
 
 # Read-ahead value for all block devices, in kb
 system_readahead_kb: 512
+
+###############################################################################
+# unbound
+###############################################################################
+unbound_powerdns_enabled: true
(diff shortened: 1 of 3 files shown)


mboruta1 (dismissed):

comay (approved):

tj90241 (commented):
Should the `bcpc::powerdns` chef recipe also be skipped if this is false?

I don't think the enabled flag propagates down to the Chef recipe at present.

ansible/playbooks/roles/common/defaults/main/base.yml
```diff
@@ -295,3 +295,9 @@ system_parameters_net.nf_conntrack_max: 262144  # noqa var-naming
```
mboruta1, line 295: Shouldnt we also change unbound such that it does not attempt to create DNS zones if powerdns is disabled? https://github.com/bloomberg/chef-bcpc/blob/development/chef/cookbooks/bcpc/recipes/unbound.rb#L29

justinjpacheco: yep, good point, @mboruta1

justinjpacheco: done

justinjpacheco:
> Should the `bcpc::powerdns` chef recipe also be skipped if this is false?
> 
> I don't think the enabled flag propagates down to the Chef recipe at present.

@tj90241 that was added here: https://github.com/bloomberg/chef-bcpc/commit/1ec6ea15a39ab55d3864491e64333a9df6473035

mboruta1 (dismissed):

mboruta1 (approved):

PR #1934, opened 2021-12-17
Moving haproxy recipe to ansible role.

npustchi:
Signed-off-by: npustchi <npustchi@bloomberg.net>

Moving HAProxy from chef to ansible.
- Removing HAProxy recipe, attributes, files, templates and chef environments and creating an ansible role to configure HAProxy.
- Removing Keystone HAProxy configuration in chef and creating a keystone role in ansible to configure keystone in HAProxy.
- Removing glance HAProxy configuration in chef and creating a keystone role in ansible to configure glance in HAProxy.
- Removing neutron HAProxy configuration in chef and creating a keystone role in ansible to configure neutron in HAProxy.
- Removing placement HAProxy configuration in chef and creating a keystone role in ansible to  configure placement in HAProxy.
- Removing nova HAProxy configuration in chef and creating a keystone role in ansible to configure nova in HAProxy.
- Removing cinder HAProxy configuration in chef and creating a keystone role in ansible to configure cinder in HAProxy.
- Removing heat HAProxy configuration in chef and creating a keystone role in ansible to configure heat in HAProxy.
- Removing watcher HAProxy configuration in chef and creating a keystone role in ansible to configure watcher in HAProxy.

to test this you can use the following patch for bve prop
https://bbgithub.dev.bloomberg.com/BCPC/chef-bcpc-prop/pull/597

Tests:
I tested with 1h1w and 3h3w topologies in bve, built successfully. Also, I built 1h1w and 3h3w bve with chef in development and compared the haproxy configs generated, which is identical.


tj90241:
As part of adding these in Ansible, you should also remove them in Chef, e.g. this glance configuration for HAProxy:
https://github.com/bloomberg/chef-bcpc/blob/master/chef/cookbooks/bcpc/recipes/glance.rb#L124

Makefile
```diff
@@ -104,6 +105,29 @@ configure-common-node :
 		-i ${inventory} ${playbooks}/configure-common-node.yml \
 		--limit cloud
 
+run-configure-haproxy : \
```
tj90241, line 108: If the target is named `run-configure-haproxy`, then I might suggest changing the names of the targets which it pulls in -- later on, when we move things like `keystone` in its entirety to Ansible, the terminology here will be confusing.

Also, I would suggest using `-f 1` here to step through headnodes one at a time.

ansible/playbooks/roles/common/defaults/main/base.yml
```diff
@@ -222,6 +222,8 @@ networking:
 use_primary_transit_interface: true
 configure_service_ip: true
 configure_network_interfaces: true
+ 
+networking_networks_primary_aggregate_cidr: []
```
tj90241, line 226: Note that pre-existing overrides will have to be moved from the Chef environment to Ansible overrides...

ansible/playbooks/roles/haproxy/defaults/main.yml
```diff
@@ -0,0 +1,93 @@
...
+
+# The number of HTTP requests resulting in a 4xx error allowed per client
+# within http_err_rate_period
+haproxy_qos_http_err_rate: 30
+
+# A list of CIDRs and/or IPs to exempt from QoS
+haproxy_qos_exemptions: []
+
+# SLO link returned in 429 responses
+haproxy_qos_slo_url: null
```
tj90241, line 90: Ditto about pre-existing Chef environment overrides in this case.

ansible/playbooks/roles/haproxy/tasks/add-apt-repo.yml
```diff
@@ -0,0 +1,10 @@
+- name: import haproxy apt key
+  apt_key:
+    data: "{{ lookup('file', '{{ haproxy_repo_key_data }}') if haproxy_repo_key_url is not defined else omit }}"
+    url: "{{ haproxy_repo_key_url if haproxy_repo_key_url is defined else omit }}"
```
tj90241, line 4: I don't think this task will work if `haproxy_repo_key_*` is not defined, perhaps just assume it is defined and let Ansible explode if not?

tj90241 (changes requested):
Some suggestions, but definitely need to remove things from Chef as they are ported to Ansible as noted in one comment.

ansible/playbooks/roles/haproxy/tasks/add-apt-repo.yml
tj90241: or, if you really wanted to make it conditional, I think these task items need to be in a block that is guarded by `when: ... is not defined`.

npustchi:
> Some suggestions, but definitely need to remove things from Chef as they are ported to Ansible as noted in one comment.

yap that would be the last step, after adding all services.

Makefile
npustchi: why do we need to go one headnode at a time.

ansible/playbooks/roles/haproxy/tasks/add-apt-repo.yml
npustchi: tested with and without, it works. it should be as we have internal repo and public url repo.

npustchi: sure, I look into block if it fits this situation.

ansible/playbooks/roles/common/defaults/main/base.yml
npustchi: yes I would create tasks for that.

ansible/playbooks/roles/haproxy/defaults/main.yml
npustchi: created the task to follow up RDTIBCC-3475

Makefile
tj90241: The Makefiles can be used for release deployment; if you deploy to all headnodes at once, you can cause all HAproxy instances to become unavailable simultaneously.  The knock-on effect here is that things like the Consul healthchecks can then fail and the VIP will be withdrawn, etc...

ansible/playbooks/roles/haproxy/tasks/add-apt-repo.yml
tj90241: I removed the definition in defaults.yml for the `haproxy` role, it fails with:
```
TASK [haproxy : import haproxy apt key] ********************************************************************************************************************************************************************************************
fatal: [r1n1]: FAILED! => {}

MSG:

The task includes an option with an undefined variable. The error was: 'haproxy_repo_key_data' is undefined

The error appears to be in '<BUILD_PATH>/modules/chef-bcpc/ansible/playbooks/roles/haproxy/tasks/add-apt-repo.yml': line 1, column 3, but may
be elsewhere in the file depending on the exact syntax problem.
```

with the diff being:
```
--- a/ansible/playbooks/roles/haproxy/defaults/main.yml
+++ b/ansible/playbooks/roles/haproxy/defaults/main.yml
@@ -6,8 +6,8 @@ haproxy_creds: "{{ chef_databags | selectattr('id', 'equalto', 'config') | map(a
 
 # Installation-related configuration options
 haproxy_repo_enabled: true
-haproxy_repo_public_url: ppa:vbernat/haproxy-1.8
-haproxy_repo_key_data: files/haproxy.key
+#haproxy_repo_public_url: ppa:vbernat/haproxy-1.8
+#haproxy_repo_key_data: files/haproxy.key
 haproxy_dependencies:
```

npustchi: it should not be removed from main.yml, add-apt-repo task is designed to work with chef-bcpc-prop to use internal haproxy repo and without to use public repo. So for testing, I would recommend to remove it from chef-bcpc-prop overrides.yml and build again. Just note that if you already built the env, then you must remove the internal url from apt repo.

Makefile
tj90241: Understood, I have some similar enhancements that we did not do because of this.

ansible/playbooks/roles/haproxy/tasks/add-apt-repo.yml
tj90241: I understand that it should not be removed - that's why I'm suggesting that you can change:
```
haproxy_repo_key_url if haproxy_repo_key_url is defined else omit
```

into:
```
haproxy_repo_key_url
```

It is included as a base attribute - it will always be defined. In the event it is undefined (which it should never be), the playbook fails anyways as I demonstrated above.

npustchi: so the way we get the public repo is using the key in the repo which data should be filled and the internal repo should be used with url. basically the omit serves to use this task with these options, in each case the other parameter, "data" or "url is omitted.

Makefile
npustchi: will apply -f 1

ansible/playbooks/roles/haproxy/filter_plugins/utils.py
```diff
@@ -0,0 +1,39 @@
+"""
```
tj90241, line 1: My Ansible-fu isn't up to snuff: Is it possible to move this `filter_plugin` to a more applicable role?  `get_hosts_chef_role` is something that other things will use in the future and it seems weird to have it in here.

Makefile
tj90241: Just reviewing again now and I don't see this yet?

Makefile
```diff
@@ -104,6 +105,71 @@ configure-common-node :
...
+	configure-heat-haproxy \
+	configure-watcher-haproxy
+
+configure-haproxy :
+
+	ansible-playbook -v \
+		-i ${inventory} ${playbooks}/site.yml \
+		-t configure-haproxy --limit headnodes
+
+configure-keystone-haproxy :
```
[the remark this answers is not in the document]
npustchi: these roles will have the entire recipes for each service, but they only configure haproxy now. I added these tags to represent the current status. When it represents the whole service we can change it to what suits in the future.

ansible/playbooks/roles/haproxy/filter_plugins/utils.py
npustchi: it is, but if we need to we can use role hierarchy.

npustchi: I actually searched this at the time of design and see role hierarchy is used in such cases.

npustchi:
> Some suggestions, but definitely need to remove things from Chef as they are ported to Ansible as noted in one comment.

this has been addressed.

tj90241 (approved):
Please just add an item to backfill moving the filter task.

ansible/playbooks/roles/haproxy/templates/haproxy.pem.j2
```diff
@@ -0,0 +1 @@
+{{ chef_databags[1]['ssl']['key'] | b64decode }}{{ chef_databags[1]['ssl']['crt'] | b64decode }}{{ chef_databags[1]['ssl']['intermediate'] | b64decode if chef_databags[1]['ssl']['intermediate'] is not none }}
```
comay, line 1: The ordering doesn't matter but it would be nice to write this out in the same order they were before: crt, intermediate (if defined) and then key.

Also, does this all need to be bunched up in a single line or can there be a return between each one?

tj90241: The ordering changed since the code was written, though yes it should be as it appears in Chef now.

Makefile
```diff
@@ -113,6 +114,71 @@ configure-common-node :
 		-i ${inventory} ${playbooks}/configure-common-node.yml \
 		--limit cloud
 
+run-configure-haproxy : \
```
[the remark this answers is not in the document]
npustchi: The list of clients is based on the order it was done by the chef. I am not sure sorting that may introduce a bug.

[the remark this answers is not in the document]
npustchi: Though I sorted the rest.

Makefile
```diff
@@ -113,6 +114,71 @@ configure-common-node :
...
+	configure-heat-haproxy \
+	configure-watcher-haproxy
+
+configure-haproxy :
+
+	ansible-playbook -v \
+		-i ${inventory} ${playbooks}/site.yml \
+		-t configure-haproxy -f 1 --limit headnodes
+
+configure-keystone-haproxy :
```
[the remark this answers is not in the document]
npustchi: we agreed to keep chef ordering.

ansible/playbooks/headnodes.yml
```diff
@@ -2,6 +2,15 @@
   gather_facts: no
   serial: "{{ step | default(ansible_play_batch | length) }}"
   roles:
-    - common
     - chef-node
+    - cinder
+    - common
```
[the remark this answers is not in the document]
comay: I understand you sorted this list and generally I support that but 1) it's inconsistent with all of the other node types and 2) the original intent was to specify the common role and then the `chef-node` type itself which is sort of the base set of roles for all node types. At this point, the other roles can be listed.

ansible/playbooks/roles/haproxy/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
+- import_tasks: add-apt-repo.yml
+  become: true
+  tags: [never,configure-haproxy]
+  when: haproxy_repo_enabled
+
```
comay, line 5: It appears there's trailing white space at the end of the line.

ansible/playbooks/roles/haproxy/templates/429.http.j2
```diff
@@ -0,0 +1,9 @@
+HTTP/1.0 429 Too Many Requests
```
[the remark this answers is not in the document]
comay: Reopening as the file is still not quite correct I think, in particular there are additional newlines but not carriage returns - how was this file created? Did you run

```
git mv \
  chef/cookbooks/bcpc/templates/default/haproxy/429.http.erb \
  ansible/playbooks/roles/haproxy/templates/429.http.j2
```

before making the parameter updates because it's definitely lost the trailing carriage returns. You may need to use `vim` to edit the file as I know it will preserve the trailing carriage returns.

Here's what I have in doing the above - you can see the two trailing carriage returns at the end of the first six lines.

```
> cat ansible/playbooks/roles/haproxy/templates/429.http.j2
HTTP/1.0 429 Too Many Requests
Cache-Control: no-cache
Connection: close
Content-Type: text/html
Retry-After: {{ haproxy_qos_retry_after }}

<html><body><h1>429 Too Many Requests</h1>
You have executed too many requests and have been rate limited. Please see the s
ervice SLO{{ " at " + haproxy_qos_slo_url if haproxy_qos_slo_url is not none }}
for more information.
</body></html>

>
> cat -e -t -v ansible/playbooks/roles/haproxy/templates/429.http.j2
HTTP/1.0 429 Too Many Requests^M^M$
Cache-Control: no-cache^M^M$
Connection: close^M^M$
Content-Type: text/html^M^M$
Retry-After: {{ haproxy_qos_retry_after }}^M^M$
^M^M$
<html><body><h1>429 Too Many Requests</h1>$
You have executed too many requests and have been rate limited. Please see the s
ervice SLO{{ " at " + haproxy_qos_slo_url if haproxy_qos_slo_url is not none }}
for more information.$
</body></html>$
$
>
```

ansible/playbooks/roles/haproxy/templates/haproxy.pem.j2
```diff
@@ -0,0 +1 @@
+{{ chef_databags[1]['ssl']['crt'] | b64decode }}{{ chef_databags[1]['ssl']['key'] | b64decode }}{{ chef_databags[1]['ssl']['intermediate'] | b64decode if chef_databags[1]['ssl']['intermediate'] is not none }}
```
comay, line 1: Apologies, I missed your question on Slack late yesterday. According to https://github.com/bloomberg/chef-bcpc/blob/development/chef/cookbooks/bcpc/recipes/haproxy.rb#L46-L68, the certs are written out in the order of `crt`, then `intermediate` if it's set, and then `key`. Also checking `/etc/haproxy/haproxy.pem` appears to confirm that.

comay (approved):

PR #1936, opened 2021-12-27
Vagrant box build fix

tj90241:
`packer` can leave artifacts in `libvirt` that prevent box
builds from succeeding in some cases, without leaving an
appropriate error message.

Perform some pre-flight activities to increase the success
rate of box builds.

### DIFF
--- a/virtual/packer/bin/create-packer-box.sh
+++ b/virtual/packer/bin/create-packer-box.sh
@@ -56,6 +56,12 @@ if [ -z "$base_box_exists" ]; then
     fi
 fi
 
+# prevent vagrant-libvirt from failing if there's scrapnel lying around
+if [ "$VAGRANT_DEFAULT_PROVIDER" == "libvirt" ]; then
+    virsh destroy output-vagrant_source || true
+    virsh undefine output-vagrant_source || true
+    virsh vol-delete --pool default output-vagrant_source.img || true
+fi
 
 # create the packer box
 # Use the script path to find the packer directory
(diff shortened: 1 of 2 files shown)


virtual/packer/bin/create-packer-box.sh
```diff
@@ -37,6 +37,10 @@ base_box_exists=$(
     || true
 )
 if [ -z "$base_box_exists" ]; then
+    # prevent vagrant-libvirt from failing if there's scrapnel lying around
```
jingqgeng, line 40: Should we check if the base box provider is libvirt before we call virsh commands?

justinjpacheco: @jingqgeng i believe the `|| true` will always succeed regardless of virsh being there or not.

justinjpacheco: never mind, i just tried this and it errored out

justinjpacheco: never mind again. while it does show an error message, the error code is still 0

```
> which virsh
> virsh || true
-bash: virsh: command not found
> echo $?
0
> virsh
-bash: virsh: command not found
> echo $?
127
```

jingqgeng (dismissed):

tj90241:
Had to also add `virsh destroy ...`, otherwise noticed that it would undefine domains but leave them running in some cases.

tj90241:
That's a very good idea, thanks, I'll add that.

tj90241:
Sorry for all the pushes, think I covered all the above.

Also discovered that the volume name that we attempt to purge seems off, perhaps some new vagrant/packer change.

comay (dismissed):

jingqgeng (approved):

comay (approved):

PR #1938, opened 2022-01-03
Hard-pin version of ansible to match ansible-lint.

psipika:

Making sure to use `ansible` with the right version vs `ansible-core`...

**Testing performed**
[Successful run here](https://github.com/psipika/chef-bcpc/runs/4692458781?check_suite_focus=true)


comay (approved):
Only in the Ansible world does 5.2.1 == 5.1.0 😹

PR #1939, opened 2022-01-06
Upgrade to Calico 3.21.2

comay:


justinjpacheco (dismissed):

tj90241 (changes requested):
https://github.com/bloomberg/chef-bcpc/blob/master/chef/cookbooks/bcpc/attributes/calico.rb#L7

We need to bump the PPA version to 3.21, right?

tj90241 (dismissed):
Just recommend on holding off merging until we decide if there's another 8.2.x release or not; don't want a new Calico whilst 8.3.0 rolls.

tj90241 (approved):

PR #1943, opened 2022-01-18
Upload and download packer box from a s3 bucket 

jingqgeng:
Add make target to upload/download a packer box from a s3 bucket. Also moved all config files under `packer/config`


tj90241 (approved):
Just suggest that you remind everyone that the config file location is changing once this is cut into a release.

PR #1944, opened 2022-01-24
Configure HW EDAC polling 

tj90241:
Configure polling of MCA banks and use of EDAC controllers.

Optionally disable polling/EDAC by default, as most server
platforms have BMCs (Baseboard Management Controllers)
managing hardware status out-of-band.

### DIFF
--- a/ansible/playbooks/roles/common/tasks/configure-hardware.yml
+++ b/ansible/playbooks/roles/common/tasks/configure-hardware.yml
@@ -1,0 +1,103 @@
+- name: Gather a list of hardware/virtualization facts
+  setup:
+    gather_subset:
+      - '!all'
+      - '!any'
+      - hardware
+      - virtual
+
+# IPMI module configuration and loading
+- name: Configure the kernel's IPMI module
+  block:
+    - name: Load ipmi_devintf kernel module
+      modprobe:
+        name: ipmi_devintf
+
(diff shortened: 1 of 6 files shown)


ansible/playbooks/roles/common/tasks/configure-hardware.yml
```diff
@@ -0,0 +1,180 @@
...
+        src: hardware/edac.conf.j2
+        dest: /etc/modprobe.d/edac.conf
+        owner: root
+        group: root
+        mode: '0644'
+      vars:
+        modules: "{{ kernel_edac_modules }}"
+  when: not hardware_edac_polling
+
+- name: Enable EDAC controllers
```
tj90241, line 74: as a note to the reviewer, I am explicitly not `modprobe`-ing EDAC modules here. I cannot think of a way to know _which_ module needs to be probe'd, and probing them all results in redundant modules being loaded. If the system is rebooted after this change, the appropriate module is loaded... which is probably sufficient?

justinjpacheco (dismissed):

comay (dismissed):
Looks good although I can't comment on the specifics to be done to disable things both dynamically and for subsequent boots but that's your area of expertise. 😉

tj90241:
Removed the need to update the CE banks dynamically, and just dynamically ripping out the EDAC modules now.

comay (approved):

PR #1945, opened 2022-01-24
move consul from chef to ansible.

npustchi:
Adding consul role to replace consul chef recipes, attributes, and templates.
Currently only conul-package is being added to ansible.
recipes:

consul-package.rb
consul.rb
attributes:

consul.rb
templates:

cloud-ip-watch.erb
consul.service.erb
if_leader.erb
if_not_primary_mysql.sh.erb
if_not_primary_proxysql.sh.erb
if_primary_mysql.sh.erb
if_primary_proxysql.sh.erb
proxysql-check-helper.erb
service-default.erb
service-elect-watch.erb
service-watch.erb

**Tests**
I tested with 1h1w and 3h3w topologies in bve, built successfully. Also, I built 1h1w and 3h3w bve with chef in development and compared the configs generated, which is identical.
The failure (if the consul leader is any node than first node would fail chef, I tested with r3n1 being leader) prior to moving HAProxy to ansible is fixed as well.


ansible/playbooks/roles/consul/templates/services.json.j2
```diff
@@ -0,0 +1,58 @@
+{
```
comay, line 1: Could you please reintroduce part of the comment from `chef/cookbooks/bcpc/attributes/consul.rb` here as in:

```
{#
 NOTE: In order to more easily switch between a ProxySQL-enabled installation
 and one where MySQL is used directly, the ProxySQL service is defined in
 Consul regardless of whether or not it is enabled. The ProxySQL Consul service
 will only be available when ProxySQL is enabled.

 Service definitions reference:
 https://www.consul.io/docs/agent/services.html
#}
```

ansible/playbooks/roles/consul/templates/watches.json.j2
```diff
@@ -0,0 +1,52 @@
+{
```
comay, line 1: Likewise here:

```
{#
 NOTE: In order to more easily switch between a ProxySQL-enabled installation
 and one where MySQL is used directly, the ProxySQL service is defined in
 Consul regardless of whether or not it is enabled. The ProxySQL Consul service
 will only be available when ProxySQL is enabled.

 Watch definitions reference:
 https://www.consul.io/docs/agent/watches.html
#}
```

ansible/playbooks/headnodes.yml
```diff
@@ -2,6 +2,7 @@
   gather_facts: no
   serial: "{{ step | default(ansible_play_batch | length) }}"
   roles:
-    - common
     - chef-node
+    - common
```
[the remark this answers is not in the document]
npustchi: Just ordered it alphabetically.

comay (dismissed):

comay (dismissed):

mboruta1:
Removed my name from list of requesters in order to reduce noise; please re-request when PR is ready for review.

PR #1948, opened 2022-01-27
[RDTIBCC-2881] Allow upload and download from s3

jingqgeng:


virtual/packer/bin/create-packer-box.sh
```diff
@@ -20,7 +20,7 @@ set -xe
 packer_dir=$(dirname "$(dirname "$0")")
 
 # Check if an official base box is added to vagrant
-config_variables="${packer_dir}/variables.json"
+config_variables="${packer_dir}/config/variables.json"
```
tj90241, line 23: Would make the team aware of this change, as the file location is changing it impacts the build pipeline and other things that assume the location of where `variables.json` file needs to exist...

tj90241 (approved):
see comment

PR #1950, opened 2022-02-20
Address installation issue with `etcd3gw`

comay:
On a new or reprovisioned worknode, the initial `chef-client` run always fails during the run of `bcpc::etcd3gw`. The underlying issue is that while the Chef recipe in question installs all of the requirements via Ubuntu packages, one of the packages - `requests` - isn't at the requested level. When `pip` is executed to install the downloaded tar file, it fails with a proxy issue when it tries to download a newer version of `requests`. And in any case, we would prefer to be using the native packaged versions of the Python requirements.

It turns out the updated `requests` requirement was introduced via

```
commit 7f6c949bc0136cbc3c4d4af1d2a3b208d655e172
Author: Davanum Srinivas <davanum@gmail.com>
Date:   Thu Jul 30 09:10:17 2020 -0400

    Fix for CVE-2018-18074 - bump requests

    CVE-2018-18074
    moderate severity
    Vulnerable versions: <= 2.19.1
    Patched version: 2.20.0

    The Requests package through 2.19.1 before 2018-09-14 for Python sends an HTTP Authorization header to an http URI upon receiving a same-hostname https-to-http redirect, which makes it easier for remote attackers to discover credentials by sniffing the network.

    Signed-off-by: Davanum Srinivas <davanum@gmail.com>
```

Canonical actually back-ported the fix for CVE-2018-18074 into their 2.18.4 version of the `python[3]-requests` package and so the installed version actually addresses this issue. As such, we can simply adjust the `requirements.txt` file to specify 2.18.4 instead of 2.20.0.

This fix has been tested on a new provisioned physical host which exhibited the installation failure previously.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/etcd3gw.rb
+++ b/chef/cookbooks/bcpc/recipes/etcd3gw.rb
@@ -1,7 +1,7 @@
 # Cookbook:: bcpc
 # Recipe:: etcd3gw
 #
-# Copyright:: 2021 Bloomberg Finance L.P.
+# Copyright:: 2022 Bloomberg Finance L.P.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


justinjpacheco (approved):

justinjpacheco:
nice work, @comay 👍

PR #1951, opened 2022-02-23
Fix mgr election timer

tj90241:
The mgr aims to beacon to the mons each `mgr_tick_period`
 seconds. If a beacon is not seen by the mons in the last
`mon_mgr_beacon_grace` seconds, then the mons mark the
currently acting mgr as failed and elect a new one in its
place.

In a previous commit, I had put the `mon_mgr_beacon_grace`
timer erroneously in the mgr section, but it's really a mon
section item as the mons are the ones which are looking for
mgr becaons and acting on a lack of response.

Moreover, after auditing the code to confirm this, I also
became aware that the mons *also* consider the mgr's tick
period when they themselves, the mons, see their own tick
period goes beyond that of the `mon_mgr_beacon_grace` less
the `mgr_tick_period`. In this case, they will reset the
mgr beacon timeoutos to avoid a potential false alarm.


tj90241:
You can audit this code, here:
https://github.com/ceph/ceph/blob/octopus/src/mon/MgrMonitor.cc#L696

This is where the mons are monitoring for the mgrs and the use of these variables is referenced.

And here is the corresponding tick for the mgrs:
https://github.com/ceph/ceph/blob/octopus/src/mgr/DaemonServer.cc#L311

comay (approved):

PR #1959, opened 2022-03-05
Fix parameter names in configure-resolver.yml, cinder.conf.erb, misc spelling

comay:
Dusted this off from an old workspace and did a sweep based on subsequent changes. The only actual interesting corrections are in configure-resolver.yml & cinder.conf.erb (the latter misspelling is ignored but we're saved by the default working out for us).


chef/cookbooks/bcpc/templates/default/cinder/cinder.conf.erb
```diff
@@ -2,7 +2,7 @@
 host = bcpc
 allow_availability_zone_fallback = true
 rootwrap_config = /etc/cinder/rootwrap.conf
-api_paste_confg = /etc/cinder/api-paste.ini
```
justinjpacheco, line 5: oh my

justinjpacheco (approved):

PR #1965, opened 2022-03-10
Disable felix logging to syslog and stdout

mboruta1:
Note that felix will continue to log to /var/log/calico/felix.log.

**Describe your changes**
Disable logging to syslog and stdout for felix. Logs for felix will now only be present in /var/log/calico/felix.log.

**Testing performed**
Tested on a local BVE.

**Additional context**
This reduces log duplication.

### DIFF
--- a/chef/cookbooks/bcpc/templates/default/calico/felix.cfg.erb
+++ b/chef/cookbooks/bcpc/templates/default/calico/felix.cfg.erb
@@ -19,13 +19,13 @@ EtcdCertFile = <%= node['bcpc']['etcd'][@cert_type]['crt']['filepath'] %>
 EtcdKeyFile = <%= node['bcpc']['etcd'][@cert_type]['key']['filepath'] %>
 
 # The log severity above which logs are sent to the stdout.
-LogSeverityScreen = Warning
+LogSeverityScreen = Fatal
 
 # The log severity above which logs are sent to the log file.
 LogSeverityFile = Warning
 
 # The log severity above which logs are sent to the syslog.
-LogSeveritySys = Warning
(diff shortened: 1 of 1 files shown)


chef/cookbooks/bcpc/templates/default/calico/felix.cfg.erb
```diff
@@ -19,13 +19,13 @@ EtcdCertFile = <%= node['bcpc']['etcd'][@cert_type]['crt']['filepath'] %>
 EtcdKeyFile = <%= node['bcpc']['etcd'][@cert_type]['key']['filepath'] %>
 
 # The log severity above which logs are sent to the stdout.
-LogSeverityScreen = Warning
+LogSeverityScreen = none
```
mboruta1, line 22: Note that we need to disable stdout logging as well because, even with `LogSeveritySys` disabled log messages printed by felix to stdout will be sent to syslog. The alternative would be to add a custom file to `/etc/rsyslog`, which is not as clean.

tj90241:
Love these kinds of changes!

tj90241 (dismissed):

chef/cookbooks/bcpc/templates/default/calico/felix.cfg.erb
mboruta1: After talking with David Comay I updated this such that we log only Fatal errors to syslog. Since errors in felix may result in system-wide network instability it would make sense to have such errors appear in syslog as well.

comay (approved):

comay (approved):

justinjpacheco (approved):

PR #1967, opened 2022-03-15
Upgrade haproxy from 1.8 to 2.4

mboruta1:
haproxy: Replace deprecated use of reqadd/reqrep

These have been deprecated for awhile in HAproxy, and now
no longer work at all in the 2.4 series. Replace them with
suggested equivalents.

haproxy: Jump to version 2.4 if using external PPA

There are no HAproxy 1.8.x releases built against Focal,
and 1.8.x is EOL sometime next year anyways. Jump to 2.4.x
as it has releases for both Bionic and Focal both, so
operators can maintain version parity when they swing from
Bionic to Focal.

haproxy: Render PEM file in documented fashion

haproxy indicates that the PEM file should be structured
such that the public key appears first, immediate (if
present) next, and finally the private key.

**Describe your changes**
Changes required to upgrade haproxy to version 2.4

**Testing performed**
Tested on a local BVE and physical test cluster.

**Additional context**
N/A

### DIFF
--- a/chef/cookbooks/bcpc/attributes/haproxy.rb
+++ b/chef/cookbooks/bcpc/attributes/haproxy.rb
@@ -4,7 +4,7 @@
 
 # Installation-related configuration options
 default['bcpc']['haproxy']['repo']['enabled'] = false
-default['bcpc']['haproxy']['repo']['url'] = 'http://ppa.launchpad.net/vbernat/haproxy-1.8/ubuntu'
+default['bcpc']['haproxy']['repo']['url'] = 'http://ppa.launchpad.net/vbernat/haproxy-2.4/ubuntu'
 default['bcpc']['haproxy']['repo']['key'] = 'haproxy/haproxy.key'
 
 # QoS protection-related configuration options
--- a/chef/cookbooks/bcpc/recipes/haproxy.rb
+++ b/chef/cookbooks/bcpc/recipes/haproxy.rb
@@ -46,12 +46,13 @@ end
 begin
 
   certs = []
-  certs.push(Base64.decode64(config['ssl']['key']))
   certs.push(Base64.decode64(config['ssl']['crt']))
 
   if config['ssl']['intermediate']
     certs.push(Base64.decode64(config['ssl']['intermediate']))
   end
+
+  certs.push(Base64.decode64(config['ssl']['key']))
 
   template '/etc/haproxy/haproxy.pem' do
     source 'haproxy/haproxy.pem.erb'
--- a/chef/cookbooks/bcpc/templates/default/haproxy/haproxy.cfg.erb
+++ b/chef/cookbooks/bcpc/templates/default/haproxy/haproxy.cfg.erb
@@ -44,7 +44,7 @@ frontend https
   stats hide-version
   stats realm Haproxy\ Statistics
   stats auth <%= @user['username'] %>:<%= @user['password'] %>
-  reqadd X-Forwarded-Proto:\ https
+  http-request add-header X-Forwarded-Proto https if { ssl_fc }
   acl url_horizon path_beg /horizon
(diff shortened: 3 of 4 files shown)


chef/cookbooks/bcpc/attributes/haproxy.rb
```diff
@@ -4,7 +4,7 @@
 
 # Installation-related configuration options
 default['bcpc']['haproxy']['repo']['enabled'] = false
```
comay, line 6: Is there any reason at this point to not set this to `true`?

comay: Right, but I guess what I'm asking is are we effectively only testing HAProxy 2.4 from this point forward and so should we change the default to use the project's repo versus Canonical now?

tj90241: Also worth noting that the last time this was reviewed, it was suggested to change this to `false` on the basis "optional features [are] disabled by default" [1]. This issue has come up in the past as well... I believe we should have a common consensus of how to treat this now and going forward, and stick to it...

[1] https://github.com/bloomberg/chef-bcpc/pull/1740"

comay: Aye. And I'm OK with leaving this as `false` by default as long as we know the automation works with that version. I also agree we should be consistent here in terms of the defaults (although we obviously don't even offer a choice in some components where we don't even allow use of the distribution's version).

mboruta1: @tj90241 I thought that statement was regarding QoS; in that QoS is an optional feature and should be disabled by default  (see [this comment](https://github.com/bloomberg/chef-bcpc/pull/1740#issuecomment-728976524)).

@comay Looking at the code I see the following repo definitions:
```
chef/cookbooks/bcpc/attributes/percona.rb:6:default['bcpc']['percona']['repo']['enabled'] = true
chef/cookbooks/bcpc/attributes/proxysql.rb:9:default['bcpc']['proxysql']['repo']['enabled'] = true
chef/cookbooks/bcpc/attributes/rabbitmq.rb:6:default['bcpc']['rabbitmq']['source']['repo']['enabled'] = false
chef/cookbooks/bcpc/attributes/openstack.rb:5:default['bcpc']['openstack']['repo']['enabled'] = true
chef/cookbooks/bcpc/attributes/ceph.rb:5:default['bcpc']['ceph']['repo']['enabled'] = false
chef/cookbooks/bcpc/attributes/haproxy.rb:6:default['bcpc']['haproxy']['repo']['enabled'] = false
```

We technically test only specific versions of mysql, proxysql, rmq, openstack, ceph, and haproxy. Regardless of whether or not we set these attributes to 'true', the repos being pointed to change, new minor versions uploaded, and thus may no work with our automation.

That being said, there should be some consistency; I would be for changing all these `false` values to `true`

comay: Yes, I mean the `repo` enabling and not QoS specifically.

I'm OK with either default but we should be consistent here. And while *not this PR*, I think if we don't actually test/work if we switch out components when the `enabled` flag is changed, then we should remove the flag and its use form the source base.

mboruta1: David and I talked out of band and the following was clarified:
- HAProxy 2.4 and 1.8 will work with the above changes; meaning that enabled or not, a working build will be produced on ubuntu 18.04 and above.
- We should make the above flags consistent across the project.

Resolving this.

tj90241 (approved):

justinjpacheco (approved):

justinjpacheco:
nice config syntax updates around the regex stuff 👍

justinjpacheco (approved):

PR #1969, opened 2022-03-16
Provide a way to specify proxies during the Packer build

comay:

This changeset provides a way to specify proxies to be used during the Packer build. It can be used, for example, to build the box behind a firewall.

This was tested in this configuration and the resulting box was used to do an upstream `1h1w` build.


psipika (approved):

virtual/network/proxyconfig.sh
```diff
@@ -1,30 +0,0 @@
...
-        sudo tee -a /etc/apt/apt.conf
-    echo 'http_proxy='"${http_proxy}" | sudo tee -a /etc/environment
-fi
-if [ "${https_proxy}" != "" ] ; then
-    echo 'Acquire::https::Proxy '\""${https_proxy}"\"';' | \
-        sudo tee -a /etc/apt/apt.conf
-    echo 'https_proxy='"${https_proxy}" | sudo tee -a /etc/environment
-fi
-
-sudo rm -f /etc/apt/apt.conf.d/70debconf
```
jingqgeng, line 30: Do we need to remove this file in the packer box, too?

virtual/packer/packer_box_provision.sh
```diff
@@ -36,11 +38,23 @@ function main {
...
-    if [ -n "$apt_key_url" ]; then
-        /usr/bin/wget -qO - "$apt_key_url" | /usr/bin/apt-key add -
+    if [ -n "${apt_key_url}" ]; then
+        /usr/bin/wget -qO - "${apt_key_url}" | /usr/bin/apt-key add -
     fi
 
-    if [ -n "$apt_url" ]; then
+    if [ -n "${http_proxy_url}" ]; then
+        echo 'Acquire::http::Proxy "'"${http_proxy_url}"'";' \
+            > /etc/apt/apt.conf.d/proxy
```
jingqgeng, line 47: Do we want to print this out to stdout with `tee`, like how it used to be done when provisioning the `network` vm?

virtual/network/proxyconfig.sh
comay: I looked into this and concluded to *not* remove it as it aligns things closer to what we do in bare metal provisioning (where the file is not removed).

virtual/packer/packer_box_provision.sh
comay: I don't think it's necessary given the script sets the `-x` option unless I've misunderstood?

jingqgeng: Just wanted to confirm. I was mostly comparing the statement to that in `proxyconfig.sh` and saw that `proxyconfig.sh` has `#!/bin/bash -x` but still used `tee`.

jingqgeng (approved):

PR #1974, opened 2022-03-22
hide floating ip networks from non-admins

justinjpacheco:
- this update alters the rbac policy for each floating ip network
    so that only users from the `admin` project will be able to see them


comay (dismissed):
Sweet!

comay (approved):

npustchi (approved):

mboruta1 (approved):

PR #1976, opened 2022-03-25
Upgrade from PXC5.7 to PXC8.0

tj90241:
* Query cache is no more, remove options for it.

  * Remove strange tunables that do not seem to exist
    anymore (e.g., wsrep_drupal_282555_workaround).

  * Replace all use of GRANT ALL ... IDENTIED BY with
    CREATE USER IF NOT EXISTS ... IDENTIFIED BY and
    a subsequent GRANT ALL. The former is not supported
    in PXC8.0 anymore.

  * Remove configuration for SST authentication, it seems
    to be handled differently now (TODO: verify this).


tj90241:
Very rough... and proxysql not playing ball with this yet for some reason.

tj90241:
Looks like we will also want to install `percona-xtradb-cluster-client` to use 8.0 client libraries.

tj90241:
BCPC build now (both with/without ProxySQL), but less SST/galera clustering...

`wsrep_sst_auth` and provisioning of authentication for SST is no more, as per this guide:
https://www.percona.com/blog/2019/10/03/percona-xtradb-cluster-8-0-new-feature-wsrep_sst_auth-removal/

Figuring out what's needed in it's place, though, as SST is not firing up right now.

tj90241:
Everything here _should_ be mostly straightforward based on the commit message.

The one change that may jump out is the removal of the `/var/lib/mysql/done` file (and replacement with `init_mysql?` method). I ended up doing this because I kept bumping into the fact that SST syncs the donor quicker now (because we're no longer provisioning the SST account as part of the bootstrap script!)... as a result, the bootstrap script would fail on anything other than the first node because the MySQL password, etc. has already been changed due to the SST from the initial node.

tj90241:
And this PR/use of PXC8.0 for production means probably needs a little more tuning/testing because it seems a LOT has changed to the configuration. But I at least wanted to checkpoint here because there are already so many changes.

chef/cookbooks/bcpc/recipes/mysql.rb
```diff
@@ -20,93 +20,109 @@
...
+    node.run_state['mysql_restart'] = true
+  end
 end
 
-execute 'add mysqlchk to /etc/services' do
-  command <<-DOC
-    printf "mysqlchk\t3307/tcp\n" >> /etc/services
-  DOC
-  not_if 'grep mysqlchk /etc/services'
+service 'mysql' do
```
justinjpacheco, line 97: does it matter if this is defined here? i think the first pass of the chef's initialization will parse/initialize this code ref before the actual execution run. might be worth experiencing with.

tj90241: It is probably possible to replace/move it back up, though I need a notification to the service resource to restart after it is moved back up in that case. The service is here because the restart has to happen after all the config is dropped, but before the initial init script is run. And the restart should only happen once.

justinjpacheco: ah, yah, just noticed `action :restart`, ok, nm

PR #1983, opened 2022-04-12
OpenStack baseline upgrade changes

tj90241:
Newer versions of UCA have changes that are incompatible
with `chef-bcpc`.

This PR contains a series of changes which are meant to
"true-up" `chef-bcpc` with the latest version the Ussuri
UCA.


tj90241:
Covered in: https://github.com/bloomberg/chef-bcpc/pull/1983

PR #1997, opened 2022-06-14
Default to 9K MTU for transit networks on virtual builds

tj90241:
In order to test 9K MTU feasibility in a development
environment, we actually need to raise the MTU to 9K by
default.

Existing VMs can continue to operate at 1500 MTU so long
as the Neutron provider network is configured as such
(and it remains so by default).

### DIFF
--- a/ansible/playbooks/roles/common/defaults/main/base.yml
+++ b/ansible/playbooks/roles/common/defaults/main/base.yml
@@ -221,7 +221,7 @@ host_aggregates: []
 ###############################################################################
 
 networking:
-  mtu: 1500
+  mtu: 9000
 
 use_primary_transit_interface: true
 configure_service_ip: true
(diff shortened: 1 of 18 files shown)


tj90241:
Still need to amend/test Virtualbox configurations...

tj90241:
Tested both `virtualbox` and `libvirt` builds, both have 9k MTU transits now, and both retain 1500 MTU for management networks (this must remain this way for compatibility with the underlying host).

One nuance... the `Vagrantfile` changed only for `libvirt`, because AFAIK there is no way to signify to `virtualbox` that the included DHCP server should specify a DHCP option for MTU. `vagrant-libvirt`, OTOH, does push the MTU to `dnsmasq` and the guest obtains the MTU setting via DHCP options.

mboruta1 (approved):
Looks good to me; running Jenkins pipelines (one libvirt, one vbox) just for sanity.

mboruta1:
Both pipelines passed.

PR #1998, opened 2022-06-15
neutron: Pick up MTU settings of the host

tj90241:
Currently, Ansible configures the MTU of the transit links
via the networking['mtu'] variable.

Neutron has a similar-ish configuration tunable for this,
which is `global_physnet_mtu` (default 1500, the standard
for ethernet networks).

The implication of this variable is that one cannot set the
MTU of an OpenStack network more than `global_physnet_mtu`.
And, moreover, all networks created inherent this MTU by
default.

Add plumbing to BCPC such that such that we set the Neutron
tunable to the same value that Ansible configures the host
with. Also add a Chef attribute that retains a 1500 MTU for
newly created networks so that we do not change the default
yet, and allow operators to create networks by default with
a 'standard' MTU even when the host supports jumbo frames.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/neutron-head.rb
+++ b/chef/cookbooks/bcpc/recipes/neutron-head.rb
@@ -318,12 +318,14 @@ end
 # create networks starts
 node['bcpc']['neutron']['networks'].each do |network|
   fixed_network = network['name']
+  default_network_mtu = node['bcpc']['neutron']['network']['default_network_mtu']
 
   raise "#{fixed_network}: no subnets defined" unless network.key?('fixed')
 
   # build the create network command line options
   network_create_opts = [
     '--provider-network-type local',
+    "--mtu #{default_network_mtu}",
   ]
 
   # networks are shared by default unless explicitly set to false in the config
@@ -388,6 +390,7 @@ node['bcpc']['neutron']['networks'].each do |network|
       new_fn_output=$(openstack network create #{float_network} \
                         --external \
                         --format shell \
+			--mtu #{default_network_mtu} \
                         --prefix 'fn_')
 
       # evaluate the shell output so we can access the values
--- a/ansible/playbooks/roles/common/defaults/main/chef.yml
+++ b/ansible/playbooks/roles/common/defaults/main/chef.yml
@@ -21,6 +21,8 @@ chef_environment:
         cmdline_linux:
           - net.ifnames=0
           - biosdevname=0
+      networking:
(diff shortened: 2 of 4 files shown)


mihalis68 (dismissed):

mboruta1 (approved):

chef/cookbooks/bcpc/recipes/neutron-head.rb
```diff
@@ -318,12 +318,14 @@
...
 node['bcpc']['neutron']['networks'].each do |network|
   fixed_network = network['name']
+  default_network_mtu = node['bcpc']['neutron']['network']['default_network_mtu']
 
   raise "#{fixed_network}: no subnets defined" unless network.key?('fixed')
 
   # build the create network command line options
   network_create_opts = [
     '--provider-network-type local',
+    "--mtu #{default_network_mtu}"
```
comay, line 328: Should this be `default_network_mtu` or `global_physnet_mtu`? I thought the network needed to have their MTU set to 9000 as well.

chef/cookbooks/bcpc/recipes/neutron-head.rb
```diff
@@ -388,6 +390,7 @@
       new_fn_output=$(openstack network create #{float_network} \
                         --external \
                         --format shell \
+			--mtu #{default_network_mtu} \
```
comay, line 393: Same comment here. It seems the given defaults in the properties aren't consistent.

chef/cookbooks/bcpc/recipes/neutron-head.rb
tj90241: It is intended to be `default_network_mtu`.

If `--mtu` is not specified here, OpenStack itself would default to `global_physnet_mtu`. But the Ansible playbooks now default `global_physnet_mtu` to 9k. Thus without something to say `--mtu 1500`, all newly created networks would have a 9k MTU instead of 1500. I don't think we are ready to do that, yet.

comay: I think I understand - so this is a transition step of sorts or laying down scaffolding? We're setting `global_physnet_mtu` to default to 9000 but for the time being, we'll still limit all of the networks to the current default of 1500?

tj90241: Exactly.

tj90241:
Fixed Chef lint issues.

comay (approved):

PR #1999, opened 2022-06-22
Install VBoxGuestAdditions into the packer image

tj90241:


tj90241:
But this now breaks `libvirt` builds because of the `packer` provisioning change so...

tj90241:
Virtualbox is unsupported, now.

PR #2001, opened 2022-06-29
neutron/model_query: Fix --share/--no-share filter

tj90241:
When splitting the DB query from the OUTER JOIN into the
two components that are UNION'd late, the 'is_shared'
filter got lost when the admin queries against network-
related collections. This results in networks, subnets,
etc. being listed unconditionally when an admin queries
for items, regardless of whether `--share` or `--no-share`
is specified.


tj90241:
I'll rebase the full patch once it passes upstream's muster, but this fixes some edge-case bug related to `admin` functionality in the interim.

comay (approved):

PR #2010, opened 2022-07-27
`/etc/cron.d/logrotate-apache2` has hours/minutes transposed

comay:

While looking into the `cookstyle` warning concerning `chef/cookbooks/bcpc/recipes/apache2.rb`, I noticed that the hours and minutes field in the `cron` line were transposed. This fixes that issue and also clears the `cookstyle` error by switching to using a `cron_d` resource.


tj90241 (approved):
hmm, I'm surprised I didn't notice the `cron_d` resource before! Nice job.

PR #2012, opened 2022-07-28
Correct pinning for latest version of `ansible` that's Python 3.6 compatible

comay:


psipika (approved):
Per https://github.com/ansible/ansible-lint/issues/1723 it looks like `noqa var-naming` won't work until 5.4.0. However, with the changes suggested here, consistency with Python 3.6-supported libraries is maintained. Thank you!

PR #2018, opened 2022-08-08
Add ability to set MTU of Calico OpenStack nets

tj90241:
Add an annotation to the etcd datamodel used to describe
WorkloadEndpoints for Calico OpenStack VMs that specifies
the MTU of the OpenStack network.

This effectively adds non-default (>1500) MTU support to
Calico OpenStack.

This also fixes an upstream bug in `nova` where the MTU
element of the interface in the `libvirt` domain XML is not
retained during a live migration, which causes the live-
migration to fail when it otherwise would not.

### DIFF
--- a/chef/cookbooks/bcpc/files/default/nova/migration.py
+++ b/chef/cookbooks/bcpc/files/default/nova/migration.py
@@ -451,7 +451,9 @@ def _update_vif_xml(xml_doc, migrate_data, get_vif_config):
         for index, dest_interface_subelem in enumerate(dest_interface_elem):
             # NOTE(mnaser): If we don't have an MTU, don't define one, else
             #               the live migration will crash.
-            if dest_interface_subelem.tag == 'mtu' and mtu is None:
+            if dest_interface_subelem.tag == 'mtu':
+                if mtu is not None:
+                    interface_dev.insert(index, mtu)
                 continue
             interface_dev.insert(index, dest_interface_subelem)
         # And finally re-insert the hw address.
--- a/chef/cookbooks/bcpc/files/default/neutron/datamodel_v3.py
+++ b/chef/cookbooks/bcpc/files/default/neutron/datamodel_v3.py
@@ -1,0 +1,297 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2018 Tigera, Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


mboruta1:
Very nice! Can you please provide diffs of `datamodel_v3.py`, `dhcp_agent.py`, `endpoints.py`? Would like to know exactly what these patches change.

tj90241:
The diffs of those can be seen here:
https://github.com/projectcalico/calico/pull/6487/files

tj90241:
Pushed a review upstream for the `nova` issue:
https://review.opendev.org/c/openstack/nova/+/852365

tj90241:
Worth mentioning here that upstream rejected the Nova commit on the premise that `libvirt` does not support changing the MTU at the time of live-migration (what I "fixed"). However, `neutron` pays no mind to what `nova` does and continues building the network with the MTU specified -- that which `nova` would otherwise switch to if we didn't have to pin it. This creates a kind of split-brain situation.

Fortunately, with Calico, this does not matter. Because `nova` is in charge of plugging the taps and controls the MTU of the same, it all "just works". The problem seems to be existent for OVS, etc. because `neutron` is managing the MTU of network elements, like bridges and switches, separately from the VIF or port that `nova` is plugging.

There is a long term feature/plan in the works to allow `libvirt` (or at least OpenStack) to support a dynamic MTU extension of sorts, but it sounds like it's a bit off and so we're on our own shall we choose to permit VMs to live-migrate when their current MTU mismatches with that of the network (i.e., by deploying this patch).

mboruta1 (approved):
Looks good to me, great work!

comay:
I'll try and take a look at this over the weekend.

chef/cookbooks/bcpc/files/default/nova/migration.py
```diff
@@ -451,7 +451,9 @@ def _update_vif_xml(xml_doc, migrate_data, get_vif_config):
         for index, dest_interface_subelem in enumerate(dest_interface_elem):
             # NOTE(mnaser): If we don't have an MTU, don't define one, else
             #               the live migration will crash.
-            if dest_interface_subelem.tag == 'mtu' and mtu is None:
+            if dest_interface_subelem.tag == 'mtu':
```
comay, line 454: The upstream proposed PR includes an informative comment that should be included here.

comay (approved):
I only have one minor comment about adding in the comment from the upstream Nova PR.

tj90241:
This feature was included in the upstream Calico repository.

PR #2020, opened 2022-08-15
Pick some more changes from Focal branch

tj90241:
Found some more changes that apply well to Bionic.

### DIFF
--- a/ansible/playbooks/roles/common/files/apt/99chef
+++ b/ansible/playbooks/roles/common/files/apt/99chef
@@ -1,0 +1,7 @@
+Package: chef
+Pin: version 14.*
+Pin-Priority: 999
+
+Package: chef-server-core
+Pin: version 12.*
+Pin-Priority: 999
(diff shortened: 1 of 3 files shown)


mboruta1 (approved):
Just a question, why lower the swap for the headnodes to 4 GB? Simply unnecessary?

tj90241:
Snuck part of the reasoning in the review comments (Focal Bento box is 32gb, instead of 64gb). After a day or more of being up, the ceph-mons complain about low free space on the root volume. This should buy ample room, I hope. And yes, there are only like dozens of MBs used by swap at any given time... even 1GB of swap would be plenty, but not sure how low we want to go.

comay:
> Snuck part of the reasoning in the commit* comments (Focal Bento box is 32gb, instead of 64gb). After a day or more of being up, the ceph-mons complain about low free space on the root volume. This should buy ample room, I hope. And yes, there are only like dozens of MBs used by swap at any given time... even 1GB of swap would be plenty, but not sure how low we want to go.

Although I don't have strong feelings about the swap space, would it make more sense to increase the size of the root file systems, at least for the headnodes given the amount of logging that takes place on them?

PR #2021, opened 2022-08-16
update ansible to publish license list to chef nodes.

npustchi:
Signed-off-by: npustchi <npustchi@bloomberg.net>

**Describe your changes**
We need to pass license list from inventory.yml in ansible to chef node. This will update it to chef.

**Testing performed**
Tested with jenkins for 1h1w toplogy.

### DIFF
--- a/ansible/playbooks/roles/common/filter_plugins/util.py
+++ b/ansible/playbooks/roles/common/filter_plugins/util.py
@@ -62,6 +62,12 @@ def update_chef_node_host_vars(a, *args, **kw):
     node_details['normal']['host_vars'] = {}
     node_details['normal']['host_vars'].update({'interfaces': interfaces})
 
+    licenses = hostvars.get('licenses')
+
+    if licenses is not None:
+        licenses = {'licenses': licenses}
+        node_details['normal'].update(licenses)
+
     # add aggregate to top level node attributes section
     aggregate = hostvars.get('aggregate')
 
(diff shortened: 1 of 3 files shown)


ansible/playbooks/roles/common/filter_plugins/util.py
```diff
@@ -62,6 +62,12 @@ def update_chef_node_host_vars(a, *args, **kw):
     node_details['normal']['host_vars'] = {}
     node_details['normal']['host_vars'].update({'interfaces': interfaces})
 
+    license = hostvars.get('license')
```
comay, line 65: Shouldn't this actually be `licenses` given each host may have multiple licenses associated with it?

comay (commented):
Also, I guess I expected there to be a `node_licenses()` function defined in `chef/cookbooks/bcpc/libraries/utils.rb`? Or is this going to be accessed in a different manner?

ansible/playbooks/roles/common/filter_plugins/util.py
npustchi: it is a type like in inventory we have run_list or transit.

npustchi:
> Also, I guess I expected there to be a `node_licenses()` function defined in `chef/cookbooks/bcpc/libraries/utils.rb`? Or is this going to be accessed in a different manner?

there is :
https://github.com/bloomberg/chef-bcpc/blob/development/chef/cookbooks/bcpc/libraries/utils.rb#L136

comay (approved):

PR #2022, opened 2022-08-17
Candidate fix for live-migration issue

tj90241:
`qemu` suffers from a race condition that bubbles up to
`libvirt`, and subsequently `nova-compute`. Specifically,
when this race condition is observed, `libvirt` raises
an (unhandled) exception that causes `nova-compute` to
leave shrapnel lying around, and the VM to enter a mostly
undefined state.

Simply retrying the failing `libvirt` invocation is enough
to workaround the race condition, effectively, based on a
reproducer that we have running locally - which is exactly
what this commit does.

If `libvirt` appears stuck in this condition (something we
have not seen with our reproducers), assume that the
migration completes to allow `nova-compute` to continue
making progress and cleaning up the domain.

### DIFF
--- a/chef/cookbooks/bcpc/files/default/nova/guest.py
+++ b/chef/cookbooks/bcpc/files/default/nova/guest.py
@@ -1,0 +1,1011 @@
+# Copyright 2010 United States Government as represented by the
+# Administrator of the National Aeronautics and Space Administration.
+# All Rights Reserved.
+# Copyright (c) 2010 Citrix Systems, Inc.
+# Copyright (c) 2011 Piston Cloud Computing, Inc
+# Copyright (c) 2012 University Of Minho
+# Copyright (c) 2013 Hewlett-Packard Development Company, L.P.
+# Copyright (c) 2015 Red Hat, Inc
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+"""
+Manages information about the guest.
+
+This class encapsulates libvirt domain provides certain
+higher level APIs around the raw libvirt API. These APIs are
+then used by all the other libvirt related classes
+"""
+
+import time
+
+from lxml import etree
+from oslo_log import log as logging
+from oslo_service import loopingcall
+from oslo_utils import encodeutils
+from oslo_utils import excutils
+from oslo_utils import importutils
+import six
+
+from nova.compute import power_state
+from nova import exception
+from nova.i18n import _
+from nova.virt import hardware
+from nova.virt.libvirt import config as vconfig
+
+libvirt = None
+
+LOG = logging.getLogger(__name__)
+
+VIR_DOMAIN_NOSTATE = 0
+VIR_DOMAIN_RUNNING = 1
+VIR_DOMAIN_BLOCKED = 2
+VIR_DOMAIN_PAUSED = 3
+VIR_DOMAIN_SHUTDOWN = 4
+VIR_DOMAIN_SHUTOFF = 5
+VIR_DOMAIN_CRASHED = 6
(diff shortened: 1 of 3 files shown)


tj90241:
said patch:
```
diff --git a/guest.py b/guest.py
index 7dd8497..ffa283a 100644
--- a/guest.py
+++ b/guest.py
@@ -642,7 +642,7 @@ class Guest(object):
         libvirtmod_qemu.virDomainQemuMonitorCommand(
             self._domain._o, 'announce_self', 1)
 
-    def get_job_info(self):
+    def get_job_info(self, retries=10):
         """Get job info for the domain
 
         Query the libvirt job info for the domain (ie progress
@@ -667,6 +667,16 @@ class Guest(object):
                     # away completclsely
                     LOG.debug("Domain has shutdown/gone away: %s", ex)
                     return JobInfo(type=libvirt.VIR_DOMAIN_JOB_COMPLETED)
+                elif ex.get_error_code() == libvirt.VIR_ERR_INTERNAL_ERROR:
+                    # http://review.opendev.org/c/openstack/nova/+/852002
+                    #
+                    # There is a (suspected) race condition in qemu which can
+                    # result in libvirt raising an internal error. Based on
+                    # current data, this condition is transient and subsequent
+                    # virDomainGetJobStats invocations will succeed.
+                    LOG.warning(f"BCPC: virDomainGetJobStats")
+                    if retries == 0:
+                        raise
                 else:
                     LOG.debug("Failed to get job stats: %s", ex)
                     raise
@@ -675,6 +685,11 @@ class Guest(object):
                 LOG.debug("Missing local virDomainGetJobStats: %s", ex)
                 JobInfo._have_job_stats = False
                 return JobInfo._get_job_stats_compat(self._domain)
+
+            # Need to retry calling virtDomainGetJobStats after backing off.
+            time.sleep(0.1)
+            LOG.warning(f"BCPC: virDomainGetJobStats: retries #{retries}")
+            return self.get_job_info(retries=retries-1)
         else:
```

chef/cookbooks/bcpc/files/default/nova/guest.py
```diff
@@ -0,0 +1,1010 @@
...
+                    LOG.debug("Domain has shutdown/gone away: %s", ex)
+                    return JobInfo(type=libvirt.VIR_DOMAIN_JOB_COMPLETED)
+                elif ex.get_error_code() == libvirt.VIR_ERR_INTERNAL_ERROR:
+                    # http://review.opendev.org/c/openstack/nova/+/852002
+                    #
+                    # There is a (suspected) race condition in qemu which can
+                    # result in libvirt raising an internal error. Based on
+                    # current data, this condition is transient and subsequent
+                    # virDomainGetJobStats invocations will succeed.
+                    LOG.warning("BCPC: virDomainGetJobStats")
```
comay, line 726: I know this warning text may not be long for this world, but I'd suggest something a bit more expansive such as

```
LOG.warning("BCPC: virDomainGetJobStats indicates transient 'no RAM info was set' QEMU condition")
```

chef/cookbooks/bcpc/files/default/nova/guest.py
```diff
@@ -0,0 +1,1010 @@
...
+                    raise
+            except AttributeError as ex:
+                # Local python binding doesn't support new API
+                LOG.debug("Missing local virDomainGetJobStats: %s", ex)
+                JobInfo._have_job_stats = False
+                return JobInfo._get_job_stats_compat(self._domain)
+
+            # Need to retry calling virtDomainGetJobStats after backing off.
+            time.sleep(0.1)
+            LOG.warning(f"BCPC: virDomainGetJobStats: retries={retries}")
```
comay, line 740:
```
LOG.warning(f"BCPC: Retrying virDomainGetJobStats after transient failure: retries={retries}")
```

chef/cookbooks/bcpc/files/default/nova/guest.py
tj90241: It already looks like this, mostly: 😛 
```
2022-08-18 14:52:30.385 881533 WARNING nova.virt.libvirt.guest [req-1d3ca2a6-c841-4aad-8462-947768085aa5 - - - - -] BCPC: virDomainGetJobStats: libvirt.libvirtError: internal error: migration was active, but no RAM info was set
```

(`oslo_log` will use the backtrace to include details about the exception that fired.)

Can chuck a few extra words in there, though.

comay: Ah, never mind then - thanks for the clarification!

comay (dismissed):

comay (approved):

PR #2031, opened 2022-08-22
Add conditions for running `make create` against focal

jingqgeng:
- Add instructions for 20.04 virtual build in README.md
- Change the provisioning scripts 
- Add more example config files


README.md
```diff
@@ -110,13 +110,25 @@ make create-packer-box
...
 
-To create a libvirt build:
+To create a libvirt build with Ubuntu 18.04:
 
 ```shell
 sudo apt-get install build-essential dnsmasq libguestfs-tools libvirt-dev pkg-config qemu-utils
 vagrant plugin install vagrant-libvirt vagrant-mutate
 vagrant box add bento/ubuntu-18.04 --box-version 202005.21.0 --provider virtualbox
 vagrant mutate bento/ubuntu-18.04 libvirt
+export VAGRANT_DEFAULT_PROVIDER=libvirt VAGRANT_VAGRANTFILE=Vagrantfile.libvirt
```
comay, line 120: Perhaps there's some other difference but it seems it's only the `vagrant box add` & `vagrant mutate` commands that are different so perhaps lines 116-135 could be consolidated to a single set of instructions along the lines of

To create a libvirt build:

```
sudo apt-get install build-essential dnsmasq libguestfs-tools libvirt-dev pkg-config qemu-utils
vagrant plugin install vagrant-libvirt vagrant-mutate
```

If you are using Ubuntu 18.04, a base Bento box for Vagrant should be added:

```
vagrant box add bento/ubuntu-18.04 --box-version 202005.21.0 --provider virtualbox
vagrant mutate bento/ubuntu-18.04 libvirt
```

Or if you are using Ubuntu 20.04:

```
vagrant plugin install vagrant-libvirt vagrant-mutate
vagrant box add bento/ubuntu-20.04 --box-version 202206.03.0 --provider virtualbox
```

After the base Bento box has been added for Vagrant:

```
export VAGRANT_DEFAULT_PROVIDER=libvirt VAGRANT_VAGRANTFILE=Vagrantfile.libvirt
make generate-chef-databags
make create-packer-box
make create all
```

ansible/playbooks/localhost.yml
```diff
@@ -2,4 +2,4 @@
   gather_facts: false
   roles:
     - common
-    - localhost
+    - localhost
```
comay, line 5: Looks like the trailing newline is missing from this file (Emacs?)

virtual/network/provisioner.sh
```diff
@@ -96,7 +124,10 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
-apt_configuration
+check_os_distribution
+if [ "${os_distribution}" == "bionic" ]; then
```
comay, line 128: I would suggest continuing to always call `apt_configuration` here but instead put this `bionic` check in that function.

virtual/vagrantbox.json.20.04.example
```diff
@@ -0,0 +1,4 @@
+{
+  "vagrant_box": "bcc-ubuntu-20.04",
+  "vagrant_box_version": "0"
+}
```
comay, line 4: Looks like the trailing newline is missing from this file (Emacs?)

ansible/playbooks/localhost.yml
jingqgeng: Yes

virtual/packer/packer_box_provision.sh
```diff
@@ -37,6 +40,16 @@ function main {
     sync
 }
 
+function check_os_distribution {
+    os_version=$(awk -F=  '$1=="PRETTY_NAME" {print $2;}' /etc/os-release)
```
tj90241, line 44: You can also use `lsb_release -sc` inline instead of the if/else/fi logic to determine it here, too.

```shell
$ lsb_release -sc
bionic
```

Might need to install a package (`lsb-release`) to provide this tool, but I think we end up installing stuff as part of BCPC that ends up pulling it in anyways -- so installing it in the base box and using it here should be fine.

This should help address @comay 's comments where you have "${os_distribution}" as you can use `lsb_release -sc` in its place.

virtual/network/provisioner.sh
```diff
@@ -33,7 +44,17 @@ switch_config() {
     if on_edge "${1}"; then
         # add masquerading source NAT rule on spines or super spines
         sudo iptables -A POSTROUTING -j MASQUERADE -o eth0 -t nat
-        sudo iptables-save | sudo tee /etc/iptables/rules.v4
+        if [ "${os_distribution}" == "bionic" ]; then
```
tj90241, line 47: I think the "Focal" method here should work equally well in Bionic - may be able to trim it out.

virtual/packer/packer_box_provision.sh
```diff
@@ -105,6 +118,9 @@ function cleanup_image {
     # blank netplan machine-id (DUID) so machines get unique ID generated on
     # boot
     truncate -s 0 /etc/machine-id
+    if [ "${os_distribution}" == "focal" ]; then
```
tj90241, line 121: Should work fine in Bionic, too.

virtual/network/provisioner.sh
jingqgeng: I'm actually not too sure why we needed this `/etc/iptables/rules.v4` file. Is it okay to just not generate it?

tj90241: It's needed to make the host act as a L3 router (see that `-A POSTROUTING -j MASQERADE` rule -- that's what we are saving and making persistent).

jingqgeng: @comay Any comment on this?

tj90241: ??

virtual/network/provisioner.sh
```diff
@@ -33,7 +34,14 @@ switch_config() {
     if on_edge "${1}"; then
         # add masquerading source NAT rule on spines or super spines
         sudo iptables -A POSTROUTING -j MASQUERADE -o eth0 -t nat
-        sudo iptables-save | sudo tee /etc/iptables/rules.v4
+    fi
+
+    if [ "$(lsb_release -sc)" == "focal" ]; then
```
tj90241, line 39: Does not need to be guarded for Focal - works on Bionic.

tj90241:
Looks good after the last two netfilter changes to reduce the if/else divergences.

tj90241 (dismissed):
🚀

ansible/playbooks/localhost.yml
comay: Can this file be un-staged given there are no real changes being made?

virtual/network/provisioner.sh
```diff
@@ -21,30 +21,43 @@ set -eux
...
+
 on_edge() {
     [[ ${on_edge_flag} == 1 ]]
 }
 
 switch_config() {
     # enable IPv4 forwarding
-    sudo sed -i 's/#\(net.ipv4.ip_forward\)/\1/g' /etc/sysctl.d/99-sysctl.conf
-    sudo sysctl -q -p /etc/sysctl.d/99-sysctl.conf
+    echo 'net.ipv4.ip_forward=1' > /etc/sysctl.d/90-bcpc.conf
```
comay, line 31: Can these two lines be run as `vagrant`? I'm surprised a `sudo` isn't required here but it is elsewhere in the script?

virtual/packer/packer_box_provision.sh
```diff
@@ -55,11 +56,11 @@ function configure_apt {
         > /etc/apt/apt.conf.d/99no-install-recommends
 
     if [ -n "${apt_url}" ]; then
-cat << EOF > /etc/apt/sources.list
-deb ${apt_url} bionic main restricted universe multiverse
-deb ${apt_url} bionic-backports main restricted universe multiverse
-deb ${apt_url} bionic-security main restricted universe multiverse
-deb ${apt_url} bionic-updates main restricted universe multiverse
+ cat << EOF > /etc/apt/sources.list
```
comay, line 59: Looks like a stray space was inserted to the start of this line?

virtual/vagrantbox.json.20.04.example
comay: Can the trailing newline be added to the file - perhaps using `touch` will fix it?

virtual/network/provisioner.sh
tj90241: From my Focal box and environment:
```
vagrant@network:~$ cat /etc/sysctl.d/90-bcpc.conf 
net.ipv4.ip_forward=1
```

```
 28 switch_config() {
 29     # enable IPv4 forwarding
 30     echo 'net.ipv4.ip_forward=1' > /etc/sysctl.d/90-bcpc.conf
 31     sysctl -qp /etc/sysctl.d/90-bcpc.conf
 ```

Comment (jingqgeng) [reply virtual/network/provisioner.sh:31]: Looks like `sudo` is not necessary in provisioning [shell](https://www.vagrantup.com/docs/provisioning/shell) scripts:
> privileged (boolean) - Specifies whether to execute the shell script as a privileged user or not (sudo). By default this is "true". Windows guests use a scheduled task to run as a true administrator without the WinRM limitations.

I can clean up all the `sudo` in our provisioning scripts.

File: virtual/network/provisioner.sh [hunk 11]
```diff
@@ -64,19 +77,20 @@ systemd_configuration() {
...
     for nameserver in ${nameservers}; do
         echo "nameserver ${nameserver}"
-    done | sudo tee /etc/resolv.conf
+    done | tee /etc/resolv.conf
 }
 
 apt_configuration() {
     # ref: ansible/playbooks/roles/common/tasks/configure-bgp.yml
-    sudo cp "/vagrant/apt-preferences" /etc/apt/preferences.d/98-bird
+    cp "/vagrant/apt-preferences" /etc/apt/preferences.d/98-bird
```
tj90241, line 85: Not sure how I missed this earlier, but this should only be done for Bionic.

virtual/network/provisioner.sh
```diff
@@ -64,19 +77,20 @@ systemd_configuration() {
...
 }
 
 package_installation() {
     dpkg --remove-architecture i386
-    apt="sudo DEBIAN_FRONTEND=noninteractive apt-get -y"
+    apt="sudo DEBIAN_FRONTEND=noninteractive\
```
jingqgeng, line 90: I need this `sudo` here because otherwise bash thinks `DEBIAN_FRONTEND=noninteractive` was a command and complain about not being able to find the command. I tried a few things but this seems to be the simplest solution.

tj90241 (dismissed):

virtual/network/provisioner.sh
comay: Can you please add a space before the trailing backslash?

virtual/network/provisioner.sh
```diff
@@ -96,7 +110,9 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
-apt_configuration
+if [ "$(lsb_release -sc)" == "bionic" ]; then
```
comay, line 113: Per Tyler's comment around line 83 and my earlier comment, I think we should always be calling `apt_configuration` but put the OS changes *there* for whatever we want do take place (in this case, for Bionic only). Perhaps in the future, we'll need to do some sort of `apt` pinning for Focal and so concentrating the logic there makes sense.

virtual/vagrantbox.json.20.04.example
jingqgeng: Opening those files with `vim` fixed the diffs.

comay (approved):

PR #2032, opened 2022-08-22
Parallelize several `Makefile` tasks

tj90241:
* Some of the variables defined in the `common` role
    have nothing to do with that role.

  * By adding a 'site' playbook that imports the role
    common to all nodes separately from node-specific
    roles, we can coerce Ansible into parallelizing
    work that was previously executed serially on a
    role-by-role basis.


tj90241:
So, while looking at this, I inadvertently discovered that variables are being shared across roles where it really does not seem like they should be.

Example:
`cloud_images` is only used by `headnodes`, yet it's in the `common` role:
https://github.com/bloomberg/chef-bcpc/blob/e7639156abc127d1636341cadc03d1be01e46d67/ansible/playbooks/roles/common/defaults/main/base.yml#L64

Probably because the file assets are also in the `common` role. OK...

But I think in some cases, this is way too over-extended. Things like the `chef-roles` are also in `common`, but really are only pertinent to the `bootstrap` role:
https://github.com/bloomberg/chef-bcpc/blob/e7639156abc127d1636341cadc03d1be01e46d67/ansible/playbooks/roles/common/defaults/main/chef.yml#L44

This may be because `cloud_vars` is in `common` and it pulls in that?
https://github.com/bloomberg/chef-bcpc/blob/e7639156abc127d1636341cadc03d1be01e46d67/ansible/playbooks/roles/common/defaults/main/base.yml#L50

Once this is all done, `playbooks/site.yml` just becomes:
```
- hosts: cloud
  gather_facts: no
  roles:
    - common
    - chef-node

- import_playbook: localhost.yml
- import_playbook: bootstraps.yml
- import_playbook: headnodes.yml
```

... and we drop playbooks for most roles. Because we don't even need them. They're just `common` and `chef-node` anyways.

And the majority of the playbook is still parallelized like we want, because the roles common to all nodes are imported at once, and roles that are unique are imported uniquely.

PR #2035, opened 2022-09-02
ansible changes to make development branch focal compatible

mzhang741:

**Describe your changes**
Based on the differences between`development` and `development-focal` branches, added conditionals based on `ansible_distribution_version` to differentiate ansible template and config files so that we could have one `development` branch that is compatible with both bionic and focal release. 

**Testing performed**
tested in bve and was able to make successful bionic build and focal build

**Additional context**

### DIFF
--- a/ansible/playbooks/roles/common/defaults/main/base.yml
+++ b/ansible/playbooks/roles/common/defaults/main/base.yml
@@ -255,7 +255,7 @@ ubuntu_apt:
 bird:
   distribution:
     name: bionic-backports
-  when: ansible_distribution_version == "18.04"
+  when: ansible_distribution_release == "bionic"
 
 ###############################################################################
 # license traits
@@ -281,7 +281,7 @@ grub_cmdline_linux: []
 ###############################################################################
 
 # Select desired I/O scheduler to be applied at startup (deadline, noop, cfq)
-hardware_io_scheduler: "{{ 'mq-' if ansible_distribution_version == '20.04' else ''}}deadline"
+hardware_io_scheduler: "{{ 'mq-deadline' if ansible_distribution_release == 'focal' else 'deadline'}}"
 
 # Select kernel NVMe multipath support.
 # Disabling addresses the issue with controller vs. disk/device naming issues,
--- a/ansible/playbooks/roles/common/defaults/main/base.yml
+++ b/ansible/playbooks/roles/common/defaults/main/base.yml
@@ -239,13 +239,13 @@ ubuntu_apt:
     - universe
     - multiverse
   sources:
-    - dist: bionic
+    - dist: "{{ ansible_distribution_release }}"
       url: http://archive.ubuntu.com/ubuntu
-    - dist: bionic-backports
+    - dist: "{{ ansible_distribution_release }}-backports"
       url: http://archive.ubuntu.com/ubuntu
-    - dist: bionic-security
+    - dist: "{{ ansible_distribution_release }}-security"
       url: http://security.ubuntu.com/ubuntu
-    - dist: bionic-updates
+    - dist: "{{ ansible_distribution_release }}-updates"
       url: http://archive.ubuntu.com/ubuntu
 
 ###############################################################################
@@ -255,6 +255,7 @@ ubuntu_apt:
 bird:
   distribution:
     name: bionic-backports
+  when: ansible_distribution_version == "18.04"
 
 ###############################################################################
 # license traits
@@ -280,7 +281,7 @@ grub_cmdline_linux: []
 ###############################################################################
 
 # Select desired I/O scheduler to be applied at startup (deadline, noop, cfq)
-hardware_io_scheduler: deadline
+hardware_io_scheduler: "{{ 'mq-' if ansible_distribution_version == '20.04' else ''}}deadline"
 
 # Select kernel NVMe multipath support.
 # Disabling addresses the issue with controller vs. disk/device naming issues,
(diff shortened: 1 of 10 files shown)


ansible/playbooks/roles/common/tasks/configure-system.yml
```diff
@@ -169,6 +191,37 @@
...
+    - ubuntu-advantage-tools
+    - update-notifier-common
+    - ubuntu-release-upgrader-core
+    - udisks2
+    - upower
+  when: ansible_distribution_version == "20.04"
+
+# Purge off dependencies of uneeded packages
+- name: Purge unused packages
+  apt:
```
jingqgeng, line 221: It should be fine to keep this task for both 18.04 and 20.04

ansible/playbooks/roles/common/templates/tinyproxy/tinyproxy.conf.j2
```diff
@@ -29,15 +29,27 @@ Listen {{ local_proxy['config']['listen'] }}
 # The upstream rules allow you to selectively route upstream connections
 # based on the host/domain of the site being accessed.
 {% for upstream in local_proxy['config'].get('upstream',[]) %}
+{% if ansible_distribution_version == '20.04' %}
```
jingqgeng, line 32: Maybe move the if statement outside of the `for upstream...` statement, so the distribution version is checked only once?

jingqgeng (commented):
Overall this looks good! 

A nit-picking suggestion would be checking `ansible_distribution_release` instead of `ansible_distribution_version` for the if conditions. The version numbers, 18.04 and 20.04, could be of a different OS. Checking `ansible_distribution_release (bionic or focal)` probably better fits our goals here.

ansible/playbooks/roles/common/defaults/main/base.yml
```diff
@@ -280,7 +281,7 @@ grub_cmdline_linux: []
 ###############################################################################
 
 # Select desired I/O scheduler to be applied at startup (deadline, noop, cfq)
-hardware_io_scheduler: deadline
```
comay, line 283: I think this reads more clearly as

```
hardware_io_scheduler: "{{ 'mq-deadline' if ansible_distribution_version == '20.04' else 'deadline'}}"
```

ansible/playbooks/roles/common/templates/ssh/sshd_config.j2
```diff
@@ -33,7 +33,7 @@ LogLevel INFO
 
 # Authentication:
 LoginGraceTime 120
-PermitRootLogin yes
+PermitRootLogin {{ 'prohibit-password' if ansible_distribution_release == 'focal' else 'yes' }}
```
comay, line 36: The default value of this parameter appears to be `prohibit-password` in both Bionic and Focal so if the consensus is use `prohibit-password.`, I would suggest dropping this line altogether so the default value is used. Tagging @tj90241 for confirmation.

ansible/playbooks/roles/common/defaults/main/chef.yml
```diff
@@ -111,6 +111,7 @@ chef_roles:
     json_class: Chef::Role
     chef_type: role
     run_list:
+      - recipe[bcpc::chef]
```
comay, line 114: This requires a corresponding set of changes under `chef/cookbooks/bcpc` and I suggest deferring this to another PR.

ansible/playbooks/roles/common/tasks/configure-chrony.yml
```diff
@@ -21,6 +21,7 @@
     path: /etc/systemd/system/chrony.service.d
     recurse: yes
     state: directory
+  when: ansible_distribution_release == "bionic"
```
comay, line 24: I'm unconvinced this work-around is incorrect even for Focal so I would suggest dropping the changes in this file for now until things are resolved.

ansible/playbooks/roles/common/tasks/configure-system.yml
```diff
@@ -158,6 +158,28 @@
     - rpcbind
     - snapd
   ignore_errors: yes  # noqa ignore-errors
+  when: ansible_distribution_release == "bionic"
+
+- name: Stop and disable unused services
```
comay, line 163: Please change the Bionic `name` line (line 149) for this as

```
Stop and disable unused services (Bionic)
```

and make this line read

```
Stop and disable unused services (Focal)
```

ansible/playbooks/roles/common/tasks/configure-system.yml
```diff
@@ -169,6 +191,36 @@
     - ubuntu-advantage-tools
     - update-notifier-common
     - ubuntu-release-upgrader-core
+  when: ansible_distribution_release == "bionic"
+
+# Purge some Ubuntu stock packages we don't need
+- name: Purge unused service packages
```
comay, line 197: Please change the Bionic `name` line (line 185) for this as

```
Purge unused service packages (Bionic)
```

and make this line read

```
Purge unused service packages (Focal)
```

ansible/playbooks/roles/common/templates/chrony/default.j2
```diff
@@ -3,7 +3,7 @@
 # the chrony daemon without editing the init script or service file.
 
 # Options to pass to chrony.
-DAEMON_OPTS="{{ '-4' if not system_enable_ipv6 else '' }}"
+DAEMON_OPTS="{{ '-F 1' if ansible_distribution_release == 'focal' else '' }}{{ '-4' if not system_enable_ipv6 else '' }}"
```
comay, line 6: I'd like to understand from @tj90241 whether or not we can just make this change for both Bionic and Focal - both have a default value of `0` and I'm unaware of system differences between the two releases where `-F 1` would work on one but not the other.

ansible/playbooks/roles/common/templates/unbound/server.conf.j2
```diff
@@ -2,6 +2,12 @@
 # managed by ansible
 ###############################################################################
 
+{% if ansible_distribution_release == 'focal' %}
```
comay, line 5: As with `sshd_config`, this change would appear to be useful in both Bionic and Focal so if that's indeed the case, let's just make it for both. Tagging @tj90241 for confirmation.

mzhang741:
changes are picked into this PR https://github.com/bloomberg/chef-bcpc/pull/2044, which has been merged into development

PR #2039, opened 2022-09-08
Vagrant downloads should verify certificates

comay:

RDTIBCC-4544

We really shouldn't be telling Vagrant to ignore SSL/TLS certificates when downloading.

This PR also includes some Copyright clean-up I forgot in my last commit and also I adjusted one of the patterns in `.gitignore` to account for separate `output-vagrant-<version>` directories.


mboruta1:
Should `--insecure` also be removed from [here](https://github.com/bloomberg/chef-bcpc/blob/master/virtual/packer/bin/create-packer-box.sh#L42)?

comay:
> Should `--insecure` also be removed from [here](https://github.com/bloomberg/chef-bcpc/blob/master/virtual/packer/bin/create-packer-box.sh#L42)?

It sure looks like it - I'll make and test the change.

mboruta1 (approved):

psipika (approved):
👍

PR #2044, opened 2022-09-23
Dual OS Build

jingqgeng:
Support both Bionic & Focal.


ansible/playbooks/roles/common/defaults/main/base.yml
```diff
@@ -280,7 +281,7 @@ grub_cmdline_linux: []
 ###############################################################################
 
 # Select desired I/O scheduler to be applied at startup (deadline, noop, cfq)
-hardware_io_scheduler: deadline
+hardware_io_scheduler: "{{ 'mq-' if ansible_distribution_version == '20.04' else ''}}deadline"
```
comay, line 286: I think this reads more clearly as:

```
hardware_io_scheduler: "{{ 'mq-deadline' if ansible_distribution_version == '20.04' else 'deadline'}}"
```

ansible/playbooks/roles/common/defaults/main/chef.yml
```diff
@@ -111,6 +111,7 @@ chef_roles:
     json_class: Chef::Role
     chef_type: role
     run_list:
+      - recipe[bcpc::chef]
```
comay, line 113: Suggestion - as the `cpu.rb` module being patched upstream is part of the `chef/ohai` project and this is a *very* targeted patch, I would call the recipe `ohai-cpu` since we're not really patching Chef but rather a different project.

ansible/playbooks/roles/common/tasks/configure-resolver.yml
```diff
@@ -49,23 +49,3 @@
...
-###############################################################################
-
-- name: stop, disable and mask systemd-resolved
-  service:
-    name: systemd-resolved
-    state: stopped
-    enabled: false
-    masked: true
-
-- name: install original resolved config file
```
comay, line 67: I believe it's only lines 67-71 that need to be removed from this playbook - the earlier lines (52-67) are still relevant.

ansible/playbooks/roles/common/tasks/configure-system.yml
```diff
@@ -158,6 +158,28 @@
     - rpcbind
     - snapd
   ignore_errors: yes  # noqa ignore-errors
+  when: ansible_distribution_version == "18.04"
+
+- name: Stop and disable unused services
```
comay, line 163: Please change the Bionic `name` line (line 149) to this

```
Stop and disable unused services (Bionic)
```

and make this line read

```
Stop and disable unused services (Focal)
```

ansible/playbooks/roles/common/tasks/configure-system.yml
```diff
@@ -169,6 +191,37 @@
     - ubuntu-advantage-tools
     - update-notifier-common
     - ubuntu-release-upgrader-core
+  when: ansible_distribution_version == "18.04"
+
+# Purge some Ubuntu stock packages we don't need
+- name: Purge unused service packages
```
comay, line 197: Please change the Bionic `name` line (line 185) to this

```
Purge unused service packages (Bionic)
```

and make this line read

```
Purge unused service packages (Focal)
```

ansible/playbooks/roles/common/tasks/configure-system.yml
```diff
@@ -169,6 +191,37 @@
...
+    - udisks2
+    - upower
+  when: ansible_distribution_version == "20.04"
+
+# Purge off dependencies of uneeded packages
+- name: Purge unused packages
+  apt:
+    autoremove: yes
+    purge: yes
+  when: ansible_distribution_version == "20.04"
```
comay, line 224: This shouldn't be limited to just Focal.

ansible/playbooks/configure-common-node.yml
```diff
@@ -11,7 +11,7 @@
 # In the future, we should move Consul, etcd, etc. out of Chef and into
 # Ansible so these dependencies can be resolved more elegantly.
 - hosts: "{{ groups.get('etcdnodes', groups.get('headnodes', [])) }}"
-  gather_facts: no
+  gather_facts: yes
```
comay, line 14: Do facts actually need to be gathered in this playbook?

chef/cookbooks/bcpc/recipes/ceph-mon.rb
```diff
@@ -14,7 +14,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
comay, line 17: I'd prefer we leave the blank line following the Copyright so this file should be reverted.

chef/cookbooks/bcpc/recipes/chef.rb
```diff
@@ -0,0 +1,23 @@
+# Cookbook:: bcpc
```
comay, line 1: As mentioned above, I think this file should be renamed to something else such as `ohai.rb` or `ohai-cpu.rb` to more closely indicate why it's here. Also, line 2 should be updated to reflect the chosen name.

chef/cookbooks/bcpc/recipes/etcd3gw.rb
```diff
@@ -14,37 +14,40 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
comay, line 16: Please add back the blank line separating the Copyright from the code.

chef/cookbooks/bcpc/recipes/neutron-head.rb
```diff
@@ -14,7 +14,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
comay, line 17: Please restore this blank line too.

virtual/network/provisioner.sh
```diff
@@ -21,40 +21,50 @@ set -eux
 
 on_edge_flag=0
 
+
```
comay, line 24: Please remove this extraneous blank line so the `on_edge_flag` assignment and the function are a bit closer to each other.

virtual/network/provisioner.sh
```diff
@@ -21,40 +21,50 @@ set -eux
...
-    sudo systemctl restart bird
+    cp "/vagrant/bird/${1}.conf" /etc/bird/bird.conf
+    systemctl restart bird
 }
 
 base_config() {
-    for s in rpcbind lxcfs snapd lxd iscsid; do
-        sudo systemctl stop ${s}
-        sudo systemctl disable ${s}
+    if [ "$(lsb_release -sc)" == "bionic" ]; then
```
comay, line 48: I'd like to suggest assigning the value of this command to a variable earlier in the script ala

```
CODENAME=$(lsb_release -sc)
```

and then use `${CODENAME}` in all of the references within the script.

virtual/packer/packer_box_provision.sh
```diff
@@ -56,10 +56,10 @@ function configure_apt {
 
     if [ -n "${apt_url}" ]; then
 cat << EOF > /etc/apt/sources.list
-deb ${apt_url} bionic main restricted universe multiverse
-deb ${apt_url} bionic-backports main restricted universe multiverse
-deb ${apt_url} bionic-security main restricted universe multiverse
-deb ${apt_url} bionic-updates main restricted universe multiverse
+deb ${apt_url} $(lsb_release -sc) main restricted universe multiverse
```
comay, line 59: As with my comment around `virtual/network/provisioner.sh`, I'd like to suggest assigning a variable such as `CODENAME` that contains the command output and use that throughout the script.

virtual/packer/packer_box_provision.sh
```diff
@@ -128,9 +129,13 @@ function cleanup_image {
 function download_debs {
     # Resynchronize package index files after above cleanup
     apt-get update
-    apt-get install --download-only -y -t bionic-backports \
-        bird2 init-system-helpers
-    apt-get install --download-only -y chrony tinyproxy unbound
+    if [ "$(lsb_release -sc)" == "bionic" ]; then
```
comay, line 132: I'd like to suggest this be slightly rearranged to look along the lines of (using my earlier suggestion about using a variable for the so-called codename)

```
    if [ "${CODENAME}" == "bionic" ]; then
        apt-get install --download-only -y -t bionic-backports \
            bird2 init-system-helpers
    elif [ "${CODENAME}" == "focal" ]; then
        apt-get install --download-only -y bird2
    fi
    apt-get install --download-only -y chrony tinyproxy unbound
```

ansible/playbooks/roles/common/defaults/main/chef.yml
jingqgeng: I think this was intended to pin chef version. I can double check.

ansible/playbooks/roles/common/tasks/configure-resolver.yml
jingqgeng: I thought we're getting rid of systemd-resolved all together? Maybe I misunderstood.

virtual/packer/packer_box_provision.sh
jingqgeng: Why `CODENAME`? I don't mind using a variable as long as we have a good name. It seems that `lsb_release -sc` is more self-explanatory in terms of what we're trying to do here.

ansible/playbooks/roles/common/defaults/main/chef.yml
comay: Ah yes, there's two things we need to do. Personally, I would still separate them out into a `chef.rb` recipe for doing the pinning and a `ohai.rb` to patch the module.

ansible/playbooks/roles/common/tasks/configure-resolver.yml
comay: It's part of the `systemd` package so it's not possible to get rid of it completely but we still want to disable, mask, etc it.

virtual/packer/packer_box_provision.sh
comay: I suggested that after reading `lsb_release(1)`

```
       -c, --codename
              Display the code name of the currently installed distribution.
```

but maybe something like `DISTRIBUTION` might make more sense (and sort of mirrors the name Ansible uses, `ansible_distribution_release`.

comay: Or maybe for consistency with the Ansible and Chef recipes, instead we should just say

```
RELEASE=$(lsb_release -rs)
```

and test `"${RELEASE}" against either `"18.04"` or `"20.04"`.

tj90241: I think codename is fine... is there any reason to change it to release?

ansible/playbooks/roles/common/defaults/main/chef.yml
jingqgeng: I tried to move this into ansible, but eventually felt like this might be the best we can do? I think this patch needs to be copied over after chef is installed on all nodes, aka, after the make target `configure-chef-nodes`. Because otherwise it'll be overwritten when chef is installed. Going through the make target list, the next target that runs on all nodes is `configure-common-node`, which is already how this is executed. 

What's the harm to patching ohai in chef? I understand that since chef is already installed, this might not be effective immediately. But at least it should be loaded the next time we chef all the nodes?

ansible/playbooks/roles/common/defaults/main/base.yml
tj90241: Please change this to `hardware_io_scheduler: mq-deadline`, as that's the default for kernel 5.4 and what all the nodes have used by default. `deadline` was the default for the Bionic 4.x kernel, which we do not and have not used for awhile.

virtual/network/provisioner.sh
jingqgeng: I changed that to `distribution_codename`. Used all lower case so it's consistent with other variables in the script.

ansible/playbooks/roles/common/defaults/main/chef.yml
tj90241: Yes -- as you've insinuated, the other way to approach it would be run `chef-client -o bcpc::<name_of_recipe>` before doing the first client run (as part of another make target or not)  to make sure that the changes are on-disk before we truly converge the node.

virtual/packer/packer_box_provision.sh
tj90241: We need `bird2` (and its dependent package, `init-system-helpers`) from `bionic-backports`. Really, it's Focal's `bird2` that we run on Bionic already... that's the reason for the conditional, if that makes sense?

ansible/playbooks/roles/common/defaults/main/chef.yml
tj90241: @jingqgeng  mentioned that this is getting run as part of `role[node]` during `configure-common-node` -- so it's sufficient as is, if we'd like to keep this in Chef.

jingqgeng: Yes. I talked with Tyler offline. All we need is to ensure `bcpc::ohai` runs before `bcpc::nova-compute`, which is already how this code works: we run `chef-client -o role[node]` [here](https://github.com/bloomberg/chef-bcpc/blob/master/ansible/playbooks/configure-common-node.yml#L29) in `ansible-playbook/configure-common-node`  and that happens before `bcpc::nova-compute` is called when we chef worknodes in `run-chef-client-worknodes`.

tj90241: Physical hosts still get run with `make all`, which includes `configure-common-node` -- there's no need for any specific steps for them.

virtual/vagrantbox.json
```diff
@@ -1,4 +1,4 @@
 {
-  "vagrant_box": "bcc-ubuntu-18.04",
+  "vagrant_box": "bcc-ubuntu-20.04",
```
comay, line 2: It seems strange for the default here to be `20.04` but there also be a new example file for `20.04`. One suggestion is to restore `vagrant/virtualbox.json` and then `git mv` it to `virtual/vagrantbox.json.18.04.example` and update the instructions to specify one always needs to create `virtual/vagrantbox.json` (need to update `.gitignore` too).

Alternatively, if we want to say that the default continues to be `18.04`, then simply restore this file.

jingqgeng: Will opt for defaulting build to `18.04` for now.

comay (approved):
Great work, @jingqgeng and a tip of the hat to @tj90241 for his initial investigating here.

jingqgeng:
> Great work, @jingqgeng and a tip of the hat to @tj90241 for his initial investigating here.

Thank you both!! Will merge in after two final builds, one for bionic & focal, finish in a bit.

PR #2045, opened 2022-09-27
etcdnodes: Use CNAMEs in lieu of service IPs

tj90241:
Conditionally allow the use of CNAMEs instead of service
IPs when building the `etcd` cluster and associated
proxies. This allows organizations to reuse CAs when
migrating `etcd` to different service IPs.


tj90241:
An optional Ansible overrides can be specified like so:
```
# etcd configuration

etcd:
  host_indexes:
    r1n1: 1
    r2n1: 2
    r3n1: 3
```

This will result in the Ansible adding host aliases for `r1n1` and friends in the form of `etcdnode-1.__cloud_domain__` -> `r1n1`'s service IP, etc. Similarly, `etcd` will be configured with the former DNS RR/alias rather than the service IP.

If this override is not specified, then the `etcd` cluster is configured with service IPs as before.

The mapping of hosts to indexes is specified statically, rather than dynamically, to facilitate migrations/cutover between different nodes.

ansible/playbooks/roles/common/templates/etc/hosts.j2
```diff
@@ -1,3 +1,4 @@
+{%- set host_to_etcdnode_cname = etcd.get('host_indexes', {}) | etcdnode_to_cnames(cloud_domain) -%}
```
tj90241, line 1: derp, this should be a Jinja variable expressed outside of the template...

comay (approved):
Nice work, Tyler.

chef/cookbooks/bcpc/templates/default/etcd/etcd.service.erb
```diff
@@ -21,10 +21,10 @@ ExecStart=/usr/local/bin/etcd \
     --trusted-ca-file=<%= node['bcpc']['etcd']['ca']['crt']['filepath'] %> \
     --cert-file=<%= node['bcpc']['etcd']['server']['crt']['filepath'] %> \
     --key-file=<%= node['bcpc']['etcd']['server']['key']['filepath'] %> \
-    --advertise-client-urls=https://<%= node['service_ip'] %>:2379 \
+    --advertise-client-urls=https://<%= etcd_advertised_name() %>:2379 \
     --listen-client-urls=https://<%= node['service_ip'] %>:2379,https://127.0.0.1:2379 \
     --listen-peer-urls=https://<%= node['service_ip'] %>:2380 \
```
mboruta1, line 26: We still use `node['service_ip']` here and directly above because we cannot use a domain name? Etcd does a bind(2) call on this input I gather from https://etcd.io/docs/v3.1/op-guide/configuration/#--listen-peer-urls.

mboruta1 (approved):
Nice :)

chef/cookbooks/bcpc/templates/default/etcd/etcd.service.erb
tj90241: correct, those are just the addresses which are bound and listened on. The ones above are the ones which are advertised to peers.

PR #2046, opened 2022-09-28
`chrony`: Fix some issues with startup/config

tj90241:
* `chrony` will persistently leave servers offline if
    they are not routable at startup. Thus, ensure that the
    servers are onlined after we get a default route.
    
* Use `server` instead of `pool` in the configuration, as
    the Ansible configuration suggests we are using that.

### DIFF
--- a/ansible/playbooks/roles/common/tasks/configure-chrony.yml
+++ b/ansible/playbooks/roles/common/tasks/configure-chrony.yml
@@ -22,23 +22,34 @@
     recurse: yes
     state: directory
 
-- name: install systemd dependency work-around for chrony
-  copy:
-    content: |
-      [Unit]
-      After=
-      After=network.target network-online.target
-      Wants=network-online.target
-    dest: /etc/systemd/system/chrony.service.d/custom.conf
-    mode: 0644
-  register: chrony_custom
+- name: install service to start monitoring sources after getting a default route
+  block:
+    - name: install the systemd unit
+      copy:
+        content: |
+          [Unit]
+          Description=Wait for default route and bring Chrony sources online
+          After=bird.service chrony.service
+          Requires=chrony.service
 
-- name: reload chrony
-  systemd:
-    name: chrony
-    daemon_reload: yes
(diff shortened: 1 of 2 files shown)


ansible/playbooks/roles/common/tasks/configure-chrony.yml
```diff
@@ -22,23 +22,33 @@
...
-    dest: /etc/systemd/system/chrony.service.d/custom.conf
-    mode: 0644
-  register: chrony_custom
+- name: install service to start monitoring sources after getting a default route
+  block:
+    - name: install the systemd unit
+      copy:
+        content: |
+          [Unit]
+          Description=Wait for default route
```
comay, line 31: Perhaps `Wait for default route and bring Chrony sources online`?

ansible/playbooks/roles/common/tasks/configure-chrony.yml
```diff
@@ -22,23 +22,33 @@
...
 
-- name: reload chrony
-  systemd:
-    name: chrony
-    daemon_reload: yes
-    enabled: yes
-  when: chrony_custom.changed  # noqa no-handler
+          [Service]
+          Type=oneshot
+          ExecStart=/bin/sh -c 'while : ; do /usr/sbin/birdc show route | grep -q 0.0.0.0/0; if [ $? -eq 0 ]; then break; fi; sleep 1; done'
```
comay, line 36: You can simplify this a bit with

```
ExecStart=/bin/sh -c 'while [ -z "$(ip route list default)" ]; do sleep 1; done'
```

tj90241: This is actually probably better in that it checks against the routing table and not `bird`, too... ok

ansible/playbooks/roles/common/tasks/configure-chrony.yml
tj90241: 👍

comay (approved):

PR #2048, opened 2022-10-10
Calico with MTU fixes

tj90241:
Currently, setting the MTU of a cloud does not really work:

  * The MTU observed in the DHCP offer will always be 1500,
    regardless of how the OpenStack network is configured.

  * Live-migrations break for any VMs which were born on
    an OpenStack network with one MTU, but which was since
    changed to another.

This PR addresses both issues (see commit messages for
details).

### DIFF
--- a/chef/cookbooks/bcpc/recipes/nova-compute.rb
+++ b/chef/cookbooks/bcpc/recipes/nova-compute.rb
@@ -261,6 +261,9 @@ end
 
 # Add feature to rewrite monitor addresses in the domain XML using values from
 # migration.conf instead to faciliate Ceph monitor migration exercises
+#
+# Also, ensure that the MTU is not changed during a live-migration.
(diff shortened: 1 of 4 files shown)


chef/cookbooks/bcpc/recipes/nova-compute.rb
```diff
@@ -261,6 +261,8 @@
 
 # Add feature to rewrite monitor addresses in the domain XML using values from
 # migration.conf instead to faciliate Ceph monitor migration exercises
+#
+# Also, ensure that the MTU is not changed during a live-migration.
```
comay, line 265: Would it be possible to add a reference to one of the most appropriate Launchpad issue that's been filed on this, even if it's not going to be addressed?

tj90241: added

comay (approved):

PR #2052, opened 2022-10-12
change markupsafe version to 2.0.1

mzhang741:
Focal Jenkins builds failed at the deployment stage due to this  ImportError: cannot import name 'soft_unicode' from 'markupsafe'. This is because rally package requires a specific version of markupsafe to be installed

**Describe your changes**
add pip install markupsafe 2.0.1 before install rally 

**Testing performed**
Jenkens focal build with 1h1w succeeded

### DIFF
--- a/chef/cookbooks/bcpc/recipes/rally.rb
+++ b/chef/cookbooks/bcpc/recipes/rally.rb
@@ -93,6 +93,7 @@ execute 'install rally in virtualenv' do
     pip install 'pip>=19.1.1'
     pip install 'decorator<=4.4.2'
     pip install 'jinja2<3.0.0'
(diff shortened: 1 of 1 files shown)


chef/cookbooks/bcpc/recipes/rally.rb
```diff
@@ -93,6 +93,7 @@
     pip install 'pip>=19.1.1'
     pip install 'decorator<=4.4.2'
     pip install 'jinja2<3.0.0'
+    pip install markupsafe==2.0.1
```
mboruta1, line 96: If you don't mind, can you please single quote this to be consistent with the other pip calls above?

mboruta1 (approved):

PR #2054, opened 2022-10-17
networking_calico: Do not checksum DHCP packets

tj90241:
Very old versions of isc-dhcp-client broke when UDP
packets had a checksum field not properly filled in as
part of GSO/checksum offload support being introduced
for virtio_net:
https://lwn.net/Articles/373209/

The missing/incorrect checksum in the UDP header was
problematic for isc-dhcp-client (and possibly other DHCP
clients) because they use raw sockets, and were unaware
of how to handle TP_STATUS_CSUMNOTREADY.

OpenStack has historically worked around this by adding an
iptables mangle rule to always rewrite the UDP header for
DHCP packets. But, isc-dhcp-client has also been patched to
account for this since 2014, and likewise has qemu (for
virtio_net) since 2009:
https://git.qemu.org/?p=qemu.git;a=commit;h=1d41b0c

The consequence of patching the mangle rule in iptables is
that the underlying OpenStack mechanisms tend to add the
rules in an order that upsets Felix. When Felix sees this
(often during live-migration, when the VM has no
networking), it can spend a fair amount of time resyncing
the rules... thus leaving a live-migrating instance with
no networking for an appreciable amount of time.

As this should not be needed for any Linux distribution in
the last decade and it can perturb Felix, simply do not
fill in DHCP UDP checksums here with the assumption that
it has been fixed/is being done elsewhere.

### DIFF
--- a/.github/scripts/run_linters.sh
+++ b/.github/scripts/run_linters.sh
@@ -23,6 +23,7 @@ function main {
     find . -name "*.sh" -print0 | xargs -0 -t shellcheck;
     find . -name "*.sh" -print0 | xargs -0 -t bashate -e E006;
     find . -name "*.py" \
+        ! -path "./chef/cookbooks/bcpc/files/default/calico/dhcp.py" \
         ! -path "./chef/cookbooks/bcpc/files/default/cinder/rbd.py" \
         ! -path "./chef/cookbooks/bcpc/files/default/etcd3gw/watch.py" \
         ! -path \
--- a/chef/cookbooks/bcpc/files/default/calico/dhcp.py
+++ b/chef/cookbooks/bcpc/files/default/calico/dhcp.py
@@ -1,0 +1,250 @@
+# Copyright 2015 Metaswitch Networks
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+import copy
+import netaddr
+import os
+import re
+import sys
+import time
+
+from neutron.agent.linux import dhcp
+from oslo_log import log as logging
+
+from networking_calico.compat import constants
+
+LOG = logging.getLogger(__name__)
+
+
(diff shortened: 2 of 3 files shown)


tj90241:
```
--- a/networking-calico/networking_calico/agent/linux/dhcp.py
+++ b/networking-calico/networking_calico/agent/linux/dhcp.py
@@ -220,21 +220,31 @@ class CalicoDeviceManager(dhcp.DeviceManager):
         pass
 
     def fill_dhcp_udp_checksums(self, *args, **kwargs):
-        retries = 10
-        while retries > 0:
-            try:
-                super(CalicoDeviceManager, self).fill_dhcp_udp_checksums(
-                    *args, **kwargs)
-            except RuntimeError:
-                # fill_dhcp_udp_checksums() can fail transiently if iptables
-                # is modified concurrently, especially with an aggressive
-                # iptables writer such as Felix running.
-                LOG.exception("Failed to insert checksum rule, may retry...")
-                time.sleep(0.1)
-                retries -= 1
-            else:
-                LOG.debug("Inserted DHCP checksum rule.")
-                break
-        else:
-            LOG.error("Failed to insert DHCP checksum rule. Exiting...")
-            sys.exit(1)
+        # NOTE(tstachecki): Very old versions of isc-dhcp-client broke when
+        # UDP packets had a checksum field not properly filled in as part of
+        # GSO/checksum offload support being introduced for virtio_net:
+        # https://lwn.net/Articles/373209/
+        #
+        # The missing/incorrect checksum in the UDP header was problematic for
+        # isc-dhcp-client (and possibly other DHCP clients) because they use
+        # raw sockets, and were unaware of how to handle
+        # TP_STATUS_CSUMNOTREADY.
+        #
+        # OpenStack has historically worked around this by adding an iptables
+        # mangle rule to always rewrite the UDP header for DHCP packets. But,
+        # isc-dhcp-client has also been patched to account for this since 2014,
+        # and likewise has qemu (for virtio_net) since 2009:
+        # https://git.qemu.org/?p=qemu.git;a=commit;h=1d41b0c
+        #
+        # The conequence of patching the mangle rule in iptables is that the
+        # underlying OpenStack mechansisms tend to add the rules in an order
+        # that upsets Felix. When Felix sees this (often during live-migration,
+        # when the VM has no networking), it can spend a fair amount of time
+        # resyncing the rules... thus leaving a live-migrating instance with
+        # no networking for an appreciable amount of time.
+        #
+        # As this should not be needed for any Linux distribution in the last
+        # decade and it can perturb Felix, simply do not add in an iptables
+        # rule to fill in DHCP UDP checksums here with the assumption that it
+        # it has been fixed/is being done elsewhere.
+        pass
```

comay (approved):
Nice find - I was well aware of the UDP checksum shenanigans but didn't realized it caused Felix such indigestion.

I added a comment to the upstream Calico PR asking about addressing this for real in upstream OpenStack so eventually Calico can drop their code as well.

PR #2055, opened 2022-10-20
ansible/common: Add tasks for partitioning volumes

tj90241:
Add some tasks for formatting NVMe namespaces, partitioning
volumes, formatting filesystems, and configuring mounts as
part of configure-node.

This allows operators to provision disparate mounts for
services such as MySQL, etcd, etc.


tj90241:
An overrides can be added to assign groups => volume configuration like so:

```
# volumes
volumes:
  worknodes:
    - block_device: /dev/nvme0n1
      mount_path: /var/lib/etcd
      volume_label: etcd
```

The volume(s) are then configured prior to automation such that individual service recipes/roles do not need to modified in order be isolated on the specified volume.

I thought about instead having a per-node mount configuration in the inventory, or a chassis model => volume mapping rather than a role => mapping, but given the primary use-case of this is to configure ephemeral storage volumes for `etcd`, `mysql` and other services... everything else seemed overkill. I'm happy to take suggestions on how else this might be specified, though.

tj90241:
To test, I added a hardware override to assign each headnode 4 additional drives, and then added this to my `overrides.yml`:
```
+
+volume_configs:
+  headnodes:
+    - block_device: /dev/sdb
+      volume_label: ceph-data
+      mount_path: /var/lib/ceph
+
+    - block_device: /dev/sdc
+      volume_label: etcd-data
+      mount_path: /var/lib/etcd
+
+    - block_device: /dev/sdd
+      volume_label: mysql-data
+      mount_path: /var/lib/mysql
+
+    - block_device: /dev/sde
+      volume_label: rabbitmq-data
+      mount_path: /var/lib/rabbitmq
+
```

After the cluster built with `make destroy create all`, the headnodes look like so:
```
/dev/sdb1                        7.9G   36M  7.4G   1% /var/lib/ceph
/dev/sdc1                        7.9G   36M  7.4G   1% /var/lib/etcd
/dev/sdd1                        7.9G   36M  7.4G   1% /var/lib/mysql
/dev/sde1                        7.9G   36M  7.4G   1% /var/lib/rabbitmq
```

mboruta1 (approved):

PR #2056, opened 2022-10-22
ansible: Always enumerate all headnode groups

tj90241:
To support mix n' match of different classes of hosts and
to avoid conditional logic scattered about to support both
"converged" and "dedicated" classes of services like Ceph,
etcd, RabbitMQ, always enumerate all groups of hosts,
even when that group's role is fulfilled by another group
of nodes.


tj90241:
Suppose we have a cloud with nodes who share `etcdnodes` and `rmqnodes` roles.

The nodes, before, could be put only under the `rmqnodes` group in `inventory.yml`. Why? Because that group came first? 🙃 

This PR fixes that by reorg-ing the contract of the inventory a bit so that you can *always* specify things like `ansible -i inventory.yml etcnodes -m ping` as an operator, and even on a `3h3w` cluster where there are no dedicated etcdnodes per se, the nodes with an etcd runlist are still matched.

That is done via changing the inventory as follows (in this case, the "data" for the nodes is actually still filled out under `rmqnodes` - but Ansible will merge the definitions for `etcdnodes` and `rmqnodes` such that the groups always reference the same thing):
```
diff --git a/ansible/inventory.yml b/ansible/inventory.yml
index 3642bc9..4bf0ee6 100644
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -32,6 +32,11 @@ all:
               - role[bootstrap]
               zone: null
+        etcdnodes:
+          hosts:
+            rmqnode-1-h: {}
+            rmqnode-2-h: {}
+            rmqnode-3-h: {}
         rmqnodes:
           hosts:
             rmqnode-2-h:
               ...
```

However, the contract with how Ansible is used changes ever so slightly. If I have a cluster which has etcd running on headnodes (that is, no discreet/dedicated etcdnodes), to select _just the etcdnodes_ (but not the headnodes), I have to do a limit of 'etcdnodes:!headnodes'.
```
$ ansible -i inventory.yml cloud -m ping --list-hosts | grep rmqnode-2-h
    rmqnode-2-h
$ ansible -i inventory.yml etcdnodes -m ping --list-hosts | grep rmqnode-2-h
    rmqnode-2-h
$ ansible -i inventory.yml 'etcdnodes:!headnodes' -m ping --list-hosts | grep rmqnode-2-h
[WARNING]: No hosts matched, nothing to do
```

tj90241:
ah nuts, still have to think about this. If a node is both an `etcdnode` _and_ `rmqnode`, it'll never get Chef'd.

tj90241:
Fixed the last issue by Chef'ing `rmqnodes`, `etcdnodes`, and `storageheadnodes` all in the same batch. There was no need to separate them, AFAIK, as none of them have any reliance upon each other.

mboruta1 (commented):
Looks good to me, will approve once the infrastructure tests pass.

PR #2059, opened 2022-10-27
Revert `ceph-mon` migration support

tj90241:
The changes to support `ceph-mon` migration are very
invasive and require custom patches to several files. Thus,
remove it from the tree at this time but leave it in a
commit for reverting at a future point in time if needed.


jingqgeng (approved):

comay (approved):
Great work all the way on this, Tyler.

PR #2063, opened 2022-11-15
Apply the endpoint check bug fix

jingqgeng:
Apply the [bug fix](https://github.com/projectcalico/calico/pull/6979 ) for looking up endpoints in calico in chef-bcpc

Signed-off-by: Jing Geng <jgeng25@bloomberg.net>

### DIFF
--- a/.github/scripts/run_linters.sh
+++ b/.github/scripts/run_linters.sh
@@ -24,6 +24,7 @@ function main {
     find . -name "*.sh" -print0 | xargs -0 -t bashate -e E006;
     find . -name "*.py" \
         ! -path "./chef/cookbooks/bcpc/files/default/calico/dhcp.py" \
+        ! -path "./chef/cookbooks/bcpc/files/default/calico/status.py" \
         ! -path "./chef/cookbooks/bcpc/files/default/etcd3gw/watch.py" \
         ! -path \
             "./chef/cookbooks/bcpc/files/default/neutron/external_net_db.py" \
--- a/chef/cookbooks/bcpc/files/default/calico/status.py
+++ b/chef/cookbooks/bcpc/files/default/calico/status.py
@@ -1,0 +1,200 @@
+
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2015 Metaswitch Networks
+# Copyright (c) 2018 Tigera, Inc. All rights reserved.
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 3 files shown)


chef/cookbooks/bcpc/recipes/calico-felix.rb
```diff
@@ -27,6 +27,19 @@
   action :delete
 end
 
+if platform?('ubuntu')
```
tj90241, line 30: `calico-felix.rb` impiles this change goes on _every_ node, but this code is only subject to `neutron-server(.rb)` -- perhaps it should go there, instead? WDYT?

chef/cookbooks/bcpc/files/default/calico/status.py
```diff
@@ -0,0 +1,200 @@
...
+        self.processing_snapshot = True
+        return old_endpoints_by_host
+
+    def _post_snapshot_hook(self, old_endpoints_by_host):
+        # Look for previous endpoints that are no longer present...
+        for hostname, ep_ids in old_endpoints_by_host.items():
+            LOG.info("host: %s", hostname)
+            # Check for particular endpoints that have disappeared, and
+            # signal those.
+            for ep_id in ep_ids:
```
tj90241, line 106: In the sense of leaving it (even) better than you found it -- slight refactoring here?

Hoist above the loop:
```python
new_ep_ids = self._endpoints_by_host.get(hostname, set())
```

Then, you can say `for ep_id in [e for e in ep_ids if e not in new_ep_ids]`, which itself reduces "for stuff in set X but not in set Y', or set notation:  X – Y: `for ep_id in ep_ids.difference(new_ep_ids)`.

.github/scripts/run_linters.sh
```diff
@@ -23,6 +23,7 @@ function main {
     find . -name "*.sh" -print0 | xargs -0 -t shellcheck;
     find . -name "*.sh" -print0 | xargs -0 -t bashate -e E006;
     find . -name "*.py" \
+        ! -path "./chef/cookbooks/bcpc/files/default/calico/status.py" \
```
comay, line 26: Please keep the list sorted.

chef/cookbooks/bcpc/recipes/calico-felix.rb
comay: Agreed, particularly since that recipe is what delivers/installs the ML2 driver.

chef/cookbooks/bcpc/files/default/calico/status.py
comay: +1 on the latter construct.

jingqgeng: Will also refactor in the PR to the upstream project

jingqgeng:
Will squash the commits when the PR is approved.

comay (approved):
Great work, Jing!

PR #2066, opened 2022-11-17
`neutron-server`: Avoid spurious build failures

tj90241:
Under Focal, `neutron-server` is observed to sometimes
fail to start cleanly the first few runs (exits with a
non-zero code). In response to this, `systemd` attempts to
the process itself a few times before signaling an error
upwards.

Unfortunately, during some builds, it is observed that
start/retry burst limits (configurable via the [Unit]
StartLimitIntervalSec and StartLimitBurst parameters)
are being violated, which ultimately causes `systemd` to
raise an error upwards, and which subsequently causes
`chef-client` to fail.

Thus, bump the service restart interval upwards and
double the default burst limits (as StartLimitIntervalSec
must be > StartLimitBurst * RestartSec to avoid restarting
the service indefinitely in the event of a true failure).

This improves the success rate of Focal builds.

### DIFF
--- a/chef/cookbooks/bcpc/files/default/neutron/override.conf
+++ b/chef/cookbooks/bcpc/files/default/neutron/override.conf
@@ -1,0 +1,6 @@
+[Unit]
+StartLimitIntervalSec=20s
+StartLimitBurst=10
+
+[Service]
+RestartSec=3s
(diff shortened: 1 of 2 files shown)


comay (dismissed):
I assume you set upon these new values by trial and error?

tj90241:
Kind of; I looked at how long it took for the issue to (almost) resolve itself based on the `journalctl` logs (when it was formally was trying to restart it every 100ms and rapidly hit the limit...).

It looked like it took about 3 seconds-ish, but as noted in the PR text, I also had to bump up StartLimitIntervalSec and StartLimitBurst to avoid the infinite-restart-loop condition, and found that doubling them got us out of that hole.

tj90241:
(no changes... I just rebased onto the development branch incorrectly... argh)

PR #2070, opened 2022-11-22
ceph: Allow mon_max_pool_pg_num tuning

tj90241:
Allow `mon_max_pool_pg_num` to be tuned, and raise the default from 64k to 128k.


tj90241:
Can make it a prop tunable if desired, but most of the other things in `ceph.conf` are not the Ceph defaults, but rather things that we found we needed...

mboruta1 (approved):

PR #2074, opened 2022-12-08
Move `cpu.rb` installation into the `chef-node` task

comay:

Tested with a one-of-everything configuration.


ansible/playbooks/roles/chef-node/tasks/chef-node.yml
```diff
@@ -8,6 +8,12 @@
   apt:
     deb: "/var/tmp/{{ chef_client_package.filename }}"
 
+- name: replace cpu.rb to account for changes in /proc/cpuinfo
+  template:
```
tj90241, line 12: `file:` and not `template:`? We generally only use the latter if Jinja templating is needed, and suffix the file with a `.j2` extension. You will also have to move `cpu.rb` to  `chef-node/files`.

ansible/playbooks/roles/chef-node/tasks/chef-node.yml
```diff
@@ -8,6 +8,12 @@
...
+    src: cpu.rb
+    dest: /opt/chef/embedded/lib/ruby/gems/2.5.0/gems/ohai-14.8.12/lib/ohai/plugins/cpu.rb
+    mode: 0664
```
tj90241, line 15: `0664` may need to be quoted (I forget why, maybe because it'll treat it as an octal value?)

ansible/playbooks/roles/chef-node/tasks/chef-node.yml
comay: 👍🏻

ansible/playbooks/roles/chef-node/tasks/chef-node.yml
comay: You only need to quote it if it doesn't have a leading zero.

tj90241: ahhh okay, that's the case when I get tripped up, then.

tj90241 (approved):
Thank you for doing this!

tj90241 (approved):

PR #2076, opened 2022-12-12
Mon config changes

wstevens11:
With the move to 4 copies the number of PGs/osd in a Maintenance scenario we need to up the max pgs per osd up to 450
mon max pgs per osd = 450

moved the default pg and pgp_num from 8 to 1024


comay:
Closing this PR as it's been subsumed by #2078 and #2081.

PR #2080, opened 2022-12-21
Correctly reference the TLS information in the HAProxy recipe

comay:

This should fix the upstream builds when it comes to the HAProxy recipe.


mboruta1 (approved):

mihalis68 (approved):
Pulled this into my build of development and it built clean!

PR #2083, opened 2023-01-06
Add initial support for Ubuntu Jammy

tj90241:
This commit allows Jammy-based stubnodes to be
provisioned (virtual builds not yet supported, though).


tj90241:
Once the Bionic bits are cleaned up, this should only result in a 2-line change to add support for Jammy.

1 of those lines can be reverted once Calico builds for Jammy are present (currently, we have to use the Calico packages from Focal as they are not yet available for Jammy).

The Jammy builds use the same kernel (5.15.x) as the Focal builds -- it's just userspace that changes.

mboruta1 (approved):
Nice!

comay (approved):

PR #2088, opened 2023-01-17
nova: Prevent misconfiguration of tap MTU

tj90241:
When `nova-compute` is restarted, it calls `plug_vif` on
all VIFs. When this is done, the MTU of the Neutron network
is used.

In cases where instances were spawned prior to the MTU of
the Neutron network MTU changing, this will result in the
tap getting the incorrect MTU. The tap, itself, is used
by `calico-dhcp-agent` to configure `dnsmasq` for the
instances on the host, so it must be set correctly.


tj90241:
This is a pretty grave bug I stumbled on tonight!

On an existing hypervisor where `nova-compute` had been bounced previously, I noticed that all the tuntap MTUs were at 9000 even though I had spawned instances when the Neutron network had an MTU of 1500. When I live-migrated instances away and back to the (same) host, the tuntaps were at 1500 as expected, because of the logic added previously ().

However, simply restarting `nova-compute` cases all the VIFs to get re-plugged, and `nova-compute` just wants to do whatever Neutron is doing (at the time). So all the VIFs follow the Neutron network, even if the VM is still at something lesser. This is problematic, because the tuntap MTU is what `calico-dhcp-agent` uses to setup `dnsmasq`, among other things.

The diff is as follows:
```
--- /usr/lib/python3/dist-packages/nova/virt/libvirt/driver.py	2022-05-16 17:52:02.000000000 +0000
+++ /usr/lib/python3/dist-packages/nova/virt/libvirt/driver.py	2023-01-17 02:08:39.808288714 +0000
@@ -1149,6 +1149,28 @@
     def plug_vifs(self, instance, network_info):
         """Plug VIFs into networks."""
         for vif in network_info:
+            # tstachecki: If the nova-compute is restarting, we need to retain
+            # the existing MTU already set for the VIF. This is provided via
+            # the instance configuration. If not for this, we derive MTU from
+            # the Neutron network, which may mismatch with how the instance is
+            # presently configured.
+            try:
+                guest = self._host.get_guest(instance)
+                cfg = self.vif_driver.get_config(instance, vif,
+                                                 instance.image_meta,
+                                                 instance.flavor,
+                                                 CONF.libvirt.virt_type,
+                                                 self._host)
+                iface = guest.get_interface_by_cfg(cfg)
+                network = vif.get('network')
+                if network:
+                    new_network = copy.deepcopy(network)
+                    new_network._set_meta({'mtu': iface.mtu})
+                    vif['network'] = new_network
+
+            except nova.exception.InstanceNotFound:
+                pass
+
             self.vif_driver.plug(instance, vif)
 
     def _unplug_vifs(self, instance, network_info, ignore_errors):
```

I'm still testing/thinking about it a little.

PR #2089, opened 2023-01-20
RDTIBCC-4253: Enable the setting of custom routes

mboruta1:
Certain deployments may require the ability to set custom routes, e.g. use a different MTU for traffic outbound to a given CIDR).

**Describe your changes**
See commit message

**Testing performed**
Tested on a local cluster both the addition and deletion of custom routes.

**Additional context**
N/A

### DIFF
--- a/ansible/playbooks/roles/common/templates/bird/bird.conf.j2
+++ b/ansible/playbooks/roles/common/templates/bird/bird.conf.j2
@@ -11,7 +11,14 @@ filter to_tor {
 
 filter to_kernel {
   krt_prefsrc = {{ interfaces['service']['ip'] }};
+{% if 'default_route_params' in networking %}
+  if net = 0.0.0.0/0 then {
+    {{ networking['default_route_params'] }}
+    accept;
+  }
+{% else %}
   if net = 0.0.0.0/0 then accept;
+{% endif %}
   reject;
 }
 
--- a/ansible/playbooks/roles/common/tasks/configure-bgp.yml
+++ b/ansible/playbooks/roles/common/tasks/configure-bgp.yml
@@ -57,3 +57,41 @@
   changed_when: false
   register: result
   until: result.rc == 0
+
+- name: add a custom route
+  block:
+    - name: create custom route systemd unit
+      template:
+        src: bird/bcc-custom-route.service.j2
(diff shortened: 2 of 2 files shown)


tj90241 (approved):

ansible/playbooks/roles/common/templates/bird/bird.conf.j2
```diff
@@ -11,7 +11,14 @@ filter to_tor {
...
 filter to_kernel {
   krt_prefsrc = {{ interfaces['service']['ip'] }};
+{% if 'default_route_params' in networking %}
+  if net = 0.0.0.0/0 then {
+    {{ networking['default_route_params'] }}
+    accept;
+  }
+{% else %}
   if net = 0.0.0.0/0 then accept;
+{% endif %}
```
comay, line 21: This is fine although I think the following is a bit clearer?

```
filter to_kernel {
  krt_prefsrc = {{ interfaces['service']['ip'] }};
  if net = 0.0.0.0/0 then {
{% if 'default_route_params' in networking %}
      {{ networking['default_route_params'] }}
{% endif %}
    accept;
  }
}
```

mboruta1: I agree that how you wrote it is clearer (I was thinking about it myself :D), but since I am messing around with bird and every change to the config file results in a reload of the service and temporary blip in network connectivity, I wanted to keep the config file of existing clusters that do not specify custom network params exactly the same. As it stands now this change will be a no-op for all clusters except those configured with non-default route parameters.

Do you think this reasoning is over-cautious?

comay: Ah, understood and makes sense. Perhaps next time we make a global change, we'll remember to do a bit of clean-up. ;-)

comay (approved):

ansible/playbooks/roles/common/templates/bird/bird.conf.j2
mihalis68: I like the caution for a first go, personally, and then a later reunification, so to speak :)

tj90241 (approved):

PR #2091, opened 2023-01-24
Add 32768 to reserved ports list

tj90241:
haproxy uses port 32768 to commmunicate details about the
QoS stick tables between hosts. If we don't reserve it,
the kernel may use it for something else and then HAproxy
will fail to start.


mboruta1 (approved):

tj90241:
In my case, when I was rebooting a headnode, `etcd` started up before `haproxy`. `etcd` bound and listened on `<local_ip>:32768`, which inhibited `haproxy` from starting subsequently. I restarted `etcd` to entice to pick another port, at which point `haproxy` was then able to start. But this makes sure that we never find ourselves in this state, again.

PR #2093, opened 2023-01-26
Shoehorn ceph-deploy in when using Jammy

tj90241:
ceph-deploy is deprecated (it has been, for awhile) and
thus no longer ships with Jammy. But as it's just a python
module, carting it from the last release that it shipped
with (Focal) is fine and lets us carry on.


tj90241:
rebase

jingqgeng (approved):

PR #2094, opened 2023-01-27
rally: Pin SQLalchemy dependency

tj90241:
SQLAlchemy 2.0.0 does not work with `rally` in it's current
form. `rally` appears to try calling `begin()` on a
connection when it already exists and has a transaction
pending as part of the `rally db ensure` command. This now
raises an exception and causes the build/CI to fail.

We really need to not use arbitrary versions of packages
in the builds, but for now pin SQLalchemy as to unbreak the
build.


psipika (approved):
👍

comay (approved):

PR #2096, opened 2023-01-31
nova-head: Drop filter before restarting scheduler

tj90241:
There is a possible race condition where the `nova.conf`
has the additional filters we drop in specified in the
configuration (which then fires notification to immediately
restart `nova-scheduler`), but we have not dropped the
module yet. This can raise an exception and prevent an
initial Chef run of `bcpc::nova-head` from completing.

Thus, reorder things within the recipe.

And while we're at it, we might as well move the Ussuri
patches up to where the filters are dropped, so that the
first time `nova-*` is restarted, the code changes take
effect (instead of being deferred until the end of the
initial client run).


mboruta1 (dismissed):
Nice! One small request, can you move the `bcpc_proxysql_user` and `bcpc_proxysql_reload` resources down as well? Think it is clearer to continue having all DB-related tasks/resources in one place.

tj90241:
yes, will do that now.

mboruta1 (approved):

psipika (commented):
purdy 👋

PR #2097, opened 2023-01-31
rabbitmq: Remove deprecated `node_health_check`, fix broker restarts

tj90241:
Also, add `set -o pipefail` - how did this work, before?

### DIFF
--- a/chef/cookbooks/bcpc/recipes/rabbitmq.rb
+++ b/chef/cookbooks/bcpc/recipes/rabbitmq.rb
@@ -85,6 +85,8 @@ begin
   # add this node to the existing rabbitmq cluster if one exists
   unless init_rmq?
     members = rmqnodes(exclude: node['hostname'])
+    username = config['rabbit']['username']
+    password = config['rabbit']['password']
 
     hosts = members.collect do |m|
       "rabbit@#{m['hostname']}"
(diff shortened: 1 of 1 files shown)


tj90241:
I have *no* idea how this was working before with RabbitMQ 3.8.x (particularly the issue pertaining to broker restarts).

Found it when testing the Jammy builds, which are horrifically broken out of the gate without these fixes.

chef/cookbooks/bcpc/recipes/rabbitmq.rb
```diff
@@ -94,12 +96,17 @@
...
+        unset http_proxy
+        unset https_proxy
 
         # try to find a healthy cluster member
         #
+        member=''
+
         for h in #{hosts}; do
-          if rabbitmqctl node_health_check -n ${h}; then
+          status=$(curl -su '#{username}:#{password}' "http://${h}:55672/api/health/checks/alarms")
```
mboruta1, line 108: I looked at https://www.rabbitmq.com/monitoring.html#health-checks and see several increasingly detailed methods of checking the health of a node using `rabbitmq-diagnostics`, but no mention of calling the API directly. Looking at [this](https://github.com/rabbitmq/rabbitmq-management/pull/854/files#diff-3540eb594ac4e5743dab75b829eeb5b75c4bb81abc1d6455e3c5da028fd10208R170) PR I gather that what you are doing here is equivalent to `rabbitmq-diagnostics -q alarms`. Just wondering:
- Why call the API directly instead of using `rabbitmq-diagnostics`?
- Why not run all stage 3 checks (`rabbitmq-diagnostics -q check_running && rabbitmq-diagnostics -q check_local_alarms`), or run the stage 5 check (`rabbitmq-diagnostics -q check_virtual_hosts`)?

tj90241: Because we are seeking candidate nodes which are suitable for us to use to perform a `join_cluster` action against. `rabbitmq-diagnostics ..` will give us the results of *our* cluster, which may be a single free-standing node, and which may have a totally different status than the cluster/instance of the node to which we are joining.

Discovery of the cluster is really something suited to Consul, but we're trying to do it in Chef in a bash script... this is just patching what we have now without addressing it properly, admittedly.

Aside: I am fine with adding more of the deeper health checks if you think they would be of value, but I do think they need to remain HTTP-based in order to probe externally for the reason mentioned above.

mboruta1 (approved):

PR #2098, opened 2023-02-01
RDTIBCC-4377: Update rally & rally-openstack to latest versions

mboruta1:

**Describe your changes**
Bump versions of rally and rally-openstack to latest.

**Testing performed**
Internal CI tests pass.

**Additional context**
N/A

### DIFF
--- a/chef/cookbooks/bcpc/attributes/rally.rb
+++ b/chef/cookbooks/bcpc/attributes/rally.rb
@@ -3,8 +3,8 @@
 ###############################################################################
 
 default['bcpc']['rally']['enabled'] = false
-default['bcpc']['rally']['rally']['version'] = '3.2.0'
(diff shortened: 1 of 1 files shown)


mboruta1:
> j/w if you have any opposition to using the `rally` package included with the OS in the future to help get rid of the venvs. Jammy ships with 3.3.0, for reference. Focal is a bit older at 2.1.0.
> 
> Alternatively, we could just pull in all the rally requirements via apt and then just have a venv with a rally overlay?
> 
> Anyways, this is passing CI so good for now. :)

Using the pip packaging allows us to easily fetch the latest of both rally and rally-openstack, but neither of these project have had a new release for 1+ years so not the greatest benefit. Still, I would like to keep the flexibility in case either projects become active again. Maybe pulling in reqs via apt would be a good middle ground, something to look into :).

PR #2100, opened 2023-02-02
Use Cinc Infra Client 16.16.13 rather than the older Chef version

comay:
A couple of comments on these changes:

`ansible/playbooks/roles/chef-node/files/cinc/cinc-wrapper` - the effective change is near the end of the file where the `timeout` parameter for `Mixlib::ShellOut.new()` is doubled as I was running into build failures where the first Cinc client `headnode` run was taking over an hour.

`ansible/playbooks/roles/chef-node/files/ohai/cpu.rb` - this updates the version of OHAI's `cpu.rb` to reflect the changes in the newer Cinc client.

I also added some (default) retries to some of the `knife` interactions with the Chef server since I was consistently seeing sporadic failures with some of these operations. I haven't determined the underlying cause of these failures (particularly since they're effectively being run over the loopback interface) but this appears to alleviate the issues I was seeing.


psipika (approved):
Awesomeness 👍

tj90241 (changes requested):
There are mixed versions in the PR currently (15.6.3, 15.6.10).

tj90241 (approved):
Nevermind, that's OHAI.

psipika (dismissed):
🤙

tj90241 (approved):

PR #2101, opened 2023-02-03
cinder: Major recipe refactoring

tj90241:
Until now, we witnessed sporadic CI failures, which were
usually due to some issues with `cinder`. As it turned out,
the CI failures were the canary in the coal mine that
pointed out severe issues with respect to the ordering and
inter-dependencies of resources within the `cinder` recipe.

This commit performs significant overhaul to the recipe
in order to avoid issues with cluster bootstrapping by
reordering resources such that configuration is laid down
before services are started.

The proper ordering of resources means that we can also
eliminate a few service bounces and 'immediate' restarts
that were formerly in place to try and hammer everything
together.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/cinder.rb
+++ b/chef/cookbooks/bcpc/recipes/cinder.rb
@@ -121,6 +121,11 @@ begin
 end
 # create cinder volume services and endpoints ends
 
+# Install *only* the base Cinder scaffolding that creates the role account
+# and provides functionality for initializing the database. This gives us
+# a window to configure the services and initialize the database prior to
+# installing the packages which provide unit files (and thus start) the
+# actual Cinder services (cinder-api, cinder-scheduler, cinder-volume).
 package %w(
   cinder-common
   python3-cinder
@@ -213,8 +218,8 @@ template '/etc/cinder/cinder.conf' do
   )
 
   notifies :restart, 'service[cinder-api]', :delayed
-  notifies :restart, 'service[cinder-volume]', :delayed
   notifies :restart, 'service[cinder-scheduler]', :delayed
+  notifies :restart, 'service[cinder-volume]', :delayed
 end
 
 # Ensure the database user is present on ProxySQL
--- a/chef/cookbooks/bcpc/recipes/cinder.rb
+++ b/chef/cookbooks/bcpc/recipes/cinder.rb
@@ -121,35 +121,16 @@ begin
 end
 # create cinder volume services and endpoints ends
 
-# cinder package installation and service definition
-package ['cinder-scheduler', 'cinder-volume'] do
-  action :upgrade
-end
+package %w(
+  cinder-common
+  python3-cinder
+)
 
-service 'cinder-api' do
-  service_name 'apache2'
-end
-
-service 'cinder-volume' do
-  retries 10
-  retry_delay 5
-end
-
-service 'cinder-scheduler' do
-  retries 10
-  retry_delay 5
-end
-
-# create policy.d dir for policy overrides
-directory '/etc/cinder/policy.d' do
-  action :create
-end
-
+# create client.*cinder Ceph users and keyrings
 directory '/etc/ceph' do
   action :create
 end
 
-# create client.*cinder Ceph users and keyrings
 cinder_config.ceph_clients.each do |client|
   template "/etc/ceph/ceph.client.#{client['client']}.keyring" do
     source 'cinder/ceph.client.cinder.keyring.erb'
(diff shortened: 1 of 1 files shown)


tj90241:
This should be the nail in the coffin for intermittent `rally` failures.

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -121,35 +121,16 @@
...
-  retries 10
-  retry_delay 5
-end
-
-# create policy.d dir for policy overrides
-directory '/etc/cinder/policy.d' do
-  action :create
-end
+package %w(
+  cinder-common
```
mboruta1, line 125: Just wondering why this package is being installed; dont see it anywhere in the recipe before the change. Maybe it was installed implicitly before and now we want to install it explicitly?

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -229,16 +271,18 @@
...
 
+# configure cinder service starts
+package 'cinder-api'
+
 execute 'disable old cinder config' do
   command 'a2disconf cinder-wsgi'
   only_if 'a2query -c cinder-wsgi'
 end
 
-file '/etc/apache2/conf-available/cinder-wsgi.conf' do
```
mboruta1, line 237: This is no longer applicable?

mboruta1 (commented):
Nice! Have minor questions, otherwise I do not see any obvious issues with this.

chef/cookbooks/bcpc/recipes/cinder.rb
tj90241: Yes -- `cinder-common` is a package ordinarily pulled in by any "higher-level" package which provides a service, such as `cinder-api` or `cinder-scheduler`. A bit on this:

`cinder-common`'s `postinst` script is what deploys the initial version of `/etc/cinder/cinder.conf`, creates the `cinder` user/group in `/etc/passwd`, etc. It does not ship with any `systemd` units, though. So, by installing it, we put ourselves in a position to configure everything and initialize the database before the services (`cinder-scheduler`, etc. start).

The recipe prior to PR was for all intents and purposes a big fat race condition that was being resolved by hammering services with restart invocations while config was being laid down and initialization performed. It led to bizarre situations... such as the database not being fully initialized with all tables needed before a `cinder` service would start.

chef/cookbooks/bcpc/recipes/cinder.rb
tj90241: This file is deployed by the Cinder packaging. We shouldn't delete it, as we never deployed it and an `apt-get dist-upgrade` or similar will simply redeploy it. The correct way to manage this file, as was already being done, is simply to `a2disconf cinder-wsgi`.

tj90241: (actually, I wonder if we even need the `cinder-api` package at all... something to investigate in the future)

chef/cookbooks/bcpc/recipes/cinder.rb
comay: Thanks for that explanation about the `postinst` script - that helps understand the rationale.

tj90241:
rebase

tj90241:
Just did a CI run and both the build and rally passed.

comay:
> Just did a CI run and both the build and rally passed.

Almost done reviewing this.

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -174,6 +155,68 @@
...
+# add AccessList filter and update cinder entry_points.txt
+if zone_config.enabled?
+  cookbook_file '/usr/lib/python3/dist-packages/cinder/scheduler/filters/access_filter.py' do
+    source 'cinder/access_filter.py'
+    notifies :run, 'execute[py3compile-cinder]', :immediately
+  end
+
+  execute 'py3compile-cinder' do
+    action :nothing
+    command 'py3compile -p python3-cinder'
```
comay, line 172: How did this work before? ;-)

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -174,6 +155,68 @@
...
+    backends: cinder_config.backends,
+    config: config,
+    headnodes: headnodes(all: true),
+    rmqnodes: rmqnodes(all: true),
+    scheduler_default_filters: cinder_config.filters
+  )
+
+  notifies :restart, 'service[cinder-api]', :delayed
+  notifies :restart, 'service[cinder-volume]', :delayed
+  notifies :restart, 'service[cinder-scheduler]', :delayed
```
comay, line 217: So what's the rationale here and on line 197 to switch to `delayed`? I'm all for reducing the number of restarts but now they're deferred to the end of the Chef run?

chef/cookbooks/bcpc/recipes/cinder.rb
tj90241: Python still detects when bytecode needs to be compiled/invalidated -- it just can't flush the work back to the disk when it does so (because the process is running as user `cinder`). So, the bytecode remains in memory and is recompiled each time the process is restarted, effectively.

chef/cookbooks/bcpc/recipes/cinder.rb
tj90241: They're deferred to the end of a Chef run when _subsequent_ changes to Cinder configuration are made, yes.

During cluster bootstrapping, the configuration is put into place before the services start and that's what makes it tick.

comay: Also, although there's likely no practical reason to order the services here in any particular way, I thought the order from a readability perspective should be the lowest level (volume manipulation) to the highest level (API) or put another way, in dependent order such as

```
notifies :restart, 'service[cinder-volume]', ...
notifies :restart, 'service[cinder-scheduler]', ...
notifies :restart, 'service[cinder-api]', ...
```

tj90241: eh... when the cluster goes up, the API is stood up and then the backends... so why would this not be that way, too? (and the official install guide seems to favor neither and suggests bouncing both at the same time)

Keep in mind that in production clusters this is all active/active anywho, so the order of the restarts is a wash in that case.

chef/cookbooks/bcpc/recipes/cinder.rb
mboruta1: Understood, thank you for the explanation

chef/cookbooks/bcpc/recipes/cinder.rb
mboruta1: Indeed, good to know! Thanks for the explanation. Maybe a comment indicating the above would be useful to include as this is not trivial to understand?

tj90241: Sure thing, added a comment in front of the package resource for `cinder-common`.

mboruta1: Nice, thanks!

mboruta1 (approved):
Looks good on my end.

PR #2111, opened 2023-03-09
nova: Fix CPU/RAM scheduler weighers

tj90241:
"Normalize the weights instead of using raw values"
introduced a change whereby all weigher classes' outputs
are normalized to the range [0, 1].

However, the CPU/RAM weighers were not adjusted to use a
percent utilization instead of an absolute value as part
of this change, so the normalization here heavily biases
(stacks) allocations on certain classes of hypervisors
over others initially in a heterogeneous environment.

e.g., suppose we have two hypervisors, one with 512GB RAM
and one with 2TB RAM. If we consider the RAMWeigher in its
current form, it will weigh the 2TB RAM hypervisor >= the
512GB RAM hypervisor until there are 1.5TB of instances on
the former (assume we are looking at RAM in isolation).
Only _then_ will the weigher actually begin giving equal
weights to encourage spreading.

The module-level comment suggests the weigher intends to
spread evenly, but does not define what "evenly" is. It
seems logical that we would want to preferentially inter-
leave across both hypervisors in a biased fashion, rather
than stacking until the resource are at equilbrium and then
spreading after.

This commit makes that change.

### DIFF
--- a/chef/cookbooks/bcpc/templates/default/nova/nova.conf.erb
+++ b/chef/cookbooks/bcpc/templates/default/nova/nova.conf.erb
@@ -161,6 +161,7 @@ workers = <%= node['bcpc']['openstack']['services']['workers'] %>
 available_filters = <%= available_filter %>
 <% end %>
 enabled_filters = <%= @enabled_filters.join(',') %>
+weight_classes = nova.scheduler.weights.affinity.ServerGroupSoftAffinityWeigher,nova.scheduler.weights.affinity.ServerGroupSoftAntiAffinityWeigher,nova.scheduler.weights.bcpc_cpu.BCPCCPUWeigher,nova.scheduler.weights.bcpc_ram.BCPCRAMWeigher,nova.scheduler.weights.compute.BuildFailureWeigher,nova.scheduler.weights.cross_cell.CrossCellWeigher,nova.scheduler.weights.disk.DiskWeigher,nova.scheduler.weights.io_ops.IoOpsWeigher,nova.scheduler.weights.metrics.MetricsWeigher,nova.scheduler.weights.pci.PCIWeigher
 
 [vnc]
 enabled = true
--- a/chef/cookbooks/bcpc/files/default/nova/bcpc_cpu.py
+++ b/chef/cookbooks/bcpc/files/default/nova/bcpc_cpu.py
@@ -1,4 +1,5 @@
 # Copyright (c) 2016, Red Hat Inc.
+# Copyright (c) 2023 Bloomberg
 # All Rights Reserved.
 #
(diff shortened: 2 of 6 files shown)


tj90241:
* Dumped all_weighers using logging and verified it matches the list now present in `nova.conf`. Removed CPUWeigher and RAMWeigher, added BCPC variants.
* Added debug logging to the BCPC variants to ensure they were active and that the upstream equivalents were not after these changes applied.
* Manually altered amount of RAM visible to the kernel on a test hypervisor to a very small amount and subsequently observed that the scheduler would not pick said hypervisor before, but did after these changes applied. This shows that the scheduler will not simply bypass a host now if and when its absolute quantity of resources is dwarfed by other hypervisors.

chef/cookbooks/bcpc/templates/default/nova/nova.conf.erb
```diff
@@ -161,6 +161,7 @@ workers = <%= node['bcpc']['openstack']['services']['workers'] %>
 available_filters = <%= available_filter %>
 <% end %>
 enabled_filters = <%= @enabled_filters.join(',') %>
+weight_classes = nova.scheduler.weights.affinity.ServerGroupSoftAffinityWeigher,nova.scheduler.weights.affinity.ServerGroupSoftAntiAffinityWeigher,nova.scheduler.weights.bcpc_cpu.BCPCCPUWeigher,nova.scheduler.weights.bcpc_ram.BCPCRAMWeigher,nova.scheduler.weights.compute.BuildFailureWeigher,nova.scheduler.weights.cross_cell.CrossCellWeigher,nova.scheduler.weights.disk.DiskWeigher,nova.scheduler.weights.io_ops.IoOpsWeigher,nova.scheduler.weights.metrics.MetricsWeigher,nova.scheduler.weights.pci.PCIWeigher
```
mboruta1, line 164: Is this list basically the default list with the CPU and RAM schedulers switched out?

tj90241: [nova.conf](https://docs.openstack.org/nova/ussuri/configuration/config.html#filter_scheduler.weight_classes) defaults to `nova.scheduler.weights.all_weighers`, so essentially expanded that (below) and did a switcher-oo... yes.

The `nova.scheduler.weights.all_weighers` should expand to this, which I dumped with logging and copy/pasted into `nova.conf`:
```
./scheduler/weights/__init__.py:    return HostWeightHandler().get_all_classes()
```

mboruta1: Understood, makes sense!

comay (approved):
Great work, Tyler - thanks for addressing my concerns.

PR #2118, opened 2023-04-06
RDTIBCC-4492: ceph: Push a map update when OSD stops

tj90241:
"fast" OSD shutdown, which is just an abrupt stop of the
OSD instead of a "clean" stop, is a default feature of new
Cephs. The mechanism bizzarely provides no notice ot the
mon that the OSD is being stopped, and thus relies on
timeouts, etc. to tear things down by default.

This, implemented as is, results in seconds of I/O latency
when OSDs stop: particularly on large clusters, which have
the clog flooded with notices pertaining to unexpected
death of an OSD.

In order to resolve the latency issues associated with
this, we instead notify the mon that we're shutting down
(with haste) so that it's not quite so brutal. Though, we
still abruptly stop.

### DIFF
--- a/chef/cookbooks/bcpc/templates/default/ceph/ceph.conf.erb
+++ b/chef/cookbooks/bcpc/templates/default/ceph/ceph.conf.erb
@@ -89,6 +89,22 @@ bluefs buffered io = true
 # have landed on the new release and tested mClock sufficiently.
 osd op queue = wpq
 
+# "fast" OSD shutdown, which is just an abrupt stop of the OSD instead of a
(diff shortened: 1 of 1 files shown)


tj90241:
With 15.2.16 in the default settings (off), I stopped a set of 4 OSDs on a 32-OSD cluster with noout set. clog spam was observed -- ~2000 message spammed to clog. I then started the OSDs, ran `ceph tell 'osd.*' injectargs --osd_fast_shutdown_notify_mon=false` and stopped them again... at which point no clog notices about unexpectedly failed OSDs were observed.

PR #2123, opened 2023-04-14
RDTIBCC-4526 Cinder deferred volume deletion

jingqgeng:
Enable deferred deletion in Cinder  so volumes are deleted asynchronously after a request is issued. This helps lift unnecessary stress to Cinder and Ceph when a lot of volumes are deleted the same time.


psipika (approved):
👍

tj90241:
Just mentioning it to see what you think about it... deferred volume deletion is only implemented for the RBD driver, not for LightOS. So in the future, one driver will do deferred deletions while other(s) will not. I personally think this is fine if it provides a benefit to RBD -- WDYT?

tj90241 (approved):
Based on the [blueprint](https://specs.openstack.org/openstack/cinder-specs/specs/stein/support-deferred-deletion-in-rbd.html), this seems great!
(though, you have the release commit in this PR and will need to be removed first)

mzhang741 (approved):

chef/cookbooks/bcpc/attributes/cinder.rb
```diff
@@ -11,6 +11,8 @@
 default['bcpc']['cinder']['workers'] = nil
 default['bcpc']['cinder']['allow_az_fallback'] = true
 default['bcpc']['cinder']['backend_native_threads_pool_size'] = nil
+default['bcpc']['cinder']['deferred_deletion_purge_interval'] = 60
```
comay, line 14: I believe this is the default value and as such, I think our convention is to use `nil` to indicate that we're not overriding it here.

jingqgeng: > 

You mean `backend_native_thread_pool_size`? It is overwritten in the props.

mboruta1: Is cinder's default for `deferred_deletion_purge_interval` 60 Jing?

comay: Correct but in this particular case, you're setting `deferred_deletion_purge_interval` to `60` which is it's "compiled in" default so there's no need to actually render it in `/etc/cinder/cinder.conf`. If you set it to `nil` here, it will use that value anyway and if one wants to customize it via the props, they can do so.

tj90241: (in that case, we should be using `osd_pool_default_pg_num` as the default with the above thinking it seems?)

comay: I don't have the source handy but in Octopus, is the default 8 or 32 for `pg[p]_num`?

tj90241: `osd_pool_default_pg_num` was 8 for Mimic, and is 32 for Octopus.

I'm suggesting that we should have used `nil` in the Ceph attributes as well with this line of thinking, instead of hard-coding the default `pg_num` value ourselves. If we had, then the default would not have to be maintained as it is in that case.

What you've proposed is a good idea (I think, anyways!) -- but it is not consistency applied across our tree... and that it should not be a PR item until it is.

comay: That's fine - it was more of a nit from me than anything else. @jingqgeng What you have is fine although using `nil` here would also.

jingqgeng: > That's fine - it was more of a nit from me than anything else. @jingqgeng What you have is fine although using `nil` here would also work.

Sure, I'm happy to change it to `nil`. Just one sec

comay (approved):

PR #2124, opened 2023-04-14
ceph: Remove defunct settings

tj90241:
Lots of settings in our `ceph.conf` either no longer exist
or are no longer needed. Do a little cleanup.


tj90241:
Commits have explanations (as needed).

Changes were tested against Quincy for regressions. I had initially tried removing the `rocksdb...` one, but there was a big performance regression when touching it so leaving it as-is for now.

mboruta1:
@jingqgeng Can you please review this? Settings should be familiar to you considering you looked through them all a few week back :)

jingqgeng:
> @jingqgeng Can you please review this? Settings should be familiar to you considering you looked through them all a few week back :)

Sure, I'll get to this on Monday.

tj90241:
It's not urgent, doesn't even need to make the next release. Just been meaning to get rid of some of these dead options.

chef/cookbooks/bcpc/attributes/ceph.rb
```diff
@@ -104,13 +98,6 @@
   'max_background_compactions=4',
 ]
 
-default['bcpc']['ceph']['bluestore_cache_size_ssd'] = 10737418240
```
jingqgeng, line 107: Are we sure we want to remove this? How will it affect the cache of OSDs? According to this [page](https://docs.ceph.com/en/quincy/rados/configuration/bluestore-config-ref/#confval-bluestore_cache_size), since we do not set `bluestore_cache_size`, either `bluestore_cache_size_hdd` or `bluestore_cache_size_ssd` will be used. I'm not sure what happens if neither is set?

jingqgeng (commented):
Do we want to merge this in while we're still on Octopus, or do we want to merge this in after we go to Quincy? I'm not too sure which version I should look for these configs, and mostly looked in Octopus.

chef/cookbooks/bcpc/attributes/ceph.rb
tj90241: I think the key bit is this:
_BlueStore can be configured to automatically resize its caches when TCMalloc is configured as the memory allocator and the bluestore_cache_autotune setting is enabled_

(and it is default enabled for us)

Note that only under *MANUAL* cache sizing is this mentioned:
_The amount of memory consumed by each OSD for BlueStore caches is determined by the bluestore_cache_size configuration option_

Basically the important setting for us is `osd_memory_target`. This is the total memory footprint that the OSD tries to hover around. The autotuner tries to keep the BlueStore data/meta/other caches all try to fit within this boundary.

Right now, we have `bluestore_cache_size` > `osd_memory_target` which makes no sense...

chef/cookbooks/bcpc/attributes/ceph.rb
```diff
@@ -74,18 +74,12 @@
 # Set tcmalloc max total thread cache
 default['bcpc']['ceph']['tcmalloc_max_total_thread_cache_bytes'] = '128MB'
 
-# Set the max open fds at the OS level
-default['bcpc']['ceph']['max_open_files'] = 2048
```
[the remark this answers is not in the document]
tj90241: LevelDB is long deprecated in favor of RocksDB.

jingqgeng (dismissed):
Nice work!!

mboruta1 (approved):
Nice!

jingqgeng:
Do we need to rebase this branch against the new `development` branch?

chef/cookbooks/bcpc/templates/default/ceph/ceph.conf.erb
```diff
@@ -60,16 +60,12 @@ osd scrub chunk max = <%= @node['bcpc']['ceph']['osd_scrub_chunk_max'] %>
 osd scrub max scrubs = <%= @node['bcpc']['ceph']['osd_max_scrubs'] %>
 
 # settings to throttle OSD recovery parameters
-osd mon report interval min = <%= @node['bcpc']['ceph']['osd_mon_report_interval_min'] %>
```
jingqgeng, line 30: This config still exists, but has been renamed to osd_mon_report_interval. See the documentation [here](https://docs.ceph.com/en/octopus/releases/mimic/?highlight=osd_mon_report_interval#upgrade-compatibility-notes)

tj90241: I found your commit for this and picked it onto this branch, thanks!

jingqgeng (approved):

PR #2128, opened 2023-04-25
Allow for multi-OS Packer box builds

tj90241:
As we prepare to transition into Ubuntu Jammy and split the
management of compute and storage infrastructure, it is
helpful to provide developers with the ability to build
clusters with varying compositions of operating system
releases (e.g., a Focal-fueled OpenStack cluster conjoined
with a Jammy-fueled Ceph cluster).

This PR provides that workflow.


tj90241:
I've updated our internal documentation and tested the BCC-on-BCC builds with this.

I still need to make changes to the CI so that it has the updated `variables.json`, but other than that... this should be good to go.

jingqgeng (approved):
Awesome!!

PR #2129, opened 2023-04-26
Default to Jammy builds for BVE storage

tj90241:


tj90241:
This just builds upon https://github.com/bloomberg/chef-bcpc/pull/2128

Now, when checking out a BVE, any dedicated storage topologies (those with storage[head]node groups) build with Jammy/Quincy. Standard 1h1w, 3h3w still build with Focal/Ussuri/Octopus for now.

jingqgeng (approved):

PR #2130, opened 2023-04-28
RDTIBCCOPT-129 Increase Update Resources Interval

jingqgeng:

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/chef/cookbooks/bcpc/attributes/nova.rb
+++ b/chef/cookbooks/bcpc/attributes/nova.rb
@@ -64,6 +64,9 @@ default['bcpc']['nova']['compute']['limits']['nofile']['hard'] = 4096
 # frequency of syncing power states between hypervisor and database
 default['bcpc']['nova']['sync_power_state_interval'] = 600
 
+# frequency of updating compute resources
+default['bcpc']['nova']['update_resources_interval'] = 3600
+
 # automatically restart guests that were running when hypervisor was rebooted
 default['bcpc']['nova']['resume_guests_state_on_host_boot'] = false
 
--- a/chef/cookbooks/bcpc/templates/default/nova/nova-compute.conf.erb
+++ b/chef/cookbooks/bcpc/templates/default/nova/nova-compute.conf.erb
@@ -4,6 +4,7 @@
 [DEFAULT]
 compute_driver = libvirt.LibvirtDriver
 enable_new_services = false
+update_resources_interval = <%= node['bcpc']['nova']['update_resources_interval'] %>
 
 [libvirt]
 virt_type = <%= @virt_type %>


tj90241:
This is very interesting, because a lot of the `ceph df` calls that the worknodes make are excessive (and redundant) in our context/case...

The only thing I think we have to watch out for is that the vCPU used and memory used stats are not reliant upon this. Suppose a hypervisor comes back from service: it will report 0 vCPU/0 RAM used. Thus, in this state, the scheduler is very likely to send it lots of VMs. If the vCPU/RAM resources are only updated 1/hr after this change, it is possible the scheduler will overload that hypervisor if it is not dynamically tracking resource use outside of this mechanism.

I have no idea about what `update_resources_interval` updates though without looking, so this could be a moot point. I would be kind of surprised if vCPU/memory stats were only sent on an interval.

chef/cookbooks/bcpc/attributes/nova.rb
```diff
@@ -64,6 +64,9 @@
 # frequency of syncing power states between hypervisor and database
 default['bcpc']['nova']['sync_power_state_interval'] = 600
 
+# frequency of updating compute resources
+default['bcpc']['nova']['update_resources_interval'] = 3600
```
comay, line 68: My only concern is are there any other bits of hypervisor data that we're not going to be providing updates to Nova for such a long time (an hour)?

jingqgeng:
> This is very interesting, because a lot of the `ceph df` calls that the worknodes make are excessive (and redundant) in our context/case...
> 
> The only thing I think we have to watch out for is that the vCPU used and memory used stats are not reliant upon this. Suppose a hypervisor comes back from service: it will report 0 vCPU/0 RAM used. Thus, in this state, the scheduler is very likely to send it lots of VMs. If the vCPU/RAM resources are only updated 1/hr after this change, it is possible the scheduler will overload that hypervisor if it is not dynamically tracking resource use outside of this mechanism.
> 
> I have no idea about what `update_resources_interval` updates though without looking, so this could be a moot point. I would be kind of surprised if vCPU/memory stats were only sent on an interval.

That's a good point, but I think it'll be fine. TL;DR: Besides the periodic update I changed here, the resources are also updated whenever there's an change in the instances in a different code path. Below I'll try my best to present the function call chains. 

In general, all the resources are updated in a function called `nova/virt/libvirt/driver/update_provider_tree()`.  
Specifically, for `ceph df`, it's invoked like the following:
[update_provider_tree](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/virt/libvirt/driver.py#L7628)
 --> [_get_local_gb_info](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/virt/libvirt/driver.py#L7675)
     --> [get_pool_info](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/virt/libvirt/storage/rbd_utils.py#L373)
         --> [ceph df](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/virt/libvirt/storage/rbd_utils.py#L383). 
         
This `update_provider_tree` is called inside of `nova/compute/resource_tracker()` whenever we build, migrate, and move instances. For example, when an instance is built and run:
 [build_and_run_instance](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/compute/manager.py#L2126)
 --> [_do_build_and_run_instance](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/compute/manager.py#L2217)
--> [_build_and_run_instance](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/compute/manager.py#L2357)
--> [resource_tracker.instance_claim](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/compute/resource_tracker.py#L116)
--> [_update](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/compute/resource_tracker.py#L1222)
--> [_update_to_placement](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/compute/resource_tracker.py#L1222)
--> [update_provider_tree](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/virt/libvirt/driver.py#L7628)... [ceph df](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/virt/libvirt/storage/rbd_utils.py#L383)
 
My change only affects when `update_available_resource` is called with `CONF.update_resources_interval`, the chain happens like this:
[update_available_resource](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/compute/manager.py#L9878)
--> [_update_available_resource_for_node](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/compute/manager.py#L9840)
--> [resource_tracker.update_available_resource](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/compute/resource_tracker.py#L847)
--> [libvirt_driver.get_available_resource](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/virt/libvirt/driver.py#L8334)
 --> [_get_local_gb_info](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/virt/libvirt/driver.py#L7675)
--> [get_pool_info](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/virt/libvirt/storage/rbd_utils.py#L373)
 --> [ceph df](https://github.com/openstack/nova/blob/3fe8880d3759cbd7b19d75dcf235dfd5c511be13/nova/virt/libvirt/storage/rbd_utils.py#L383).

chef/cookbooks/bcpc/attributes/nova.rb
jingqgeng: Please see my comment below.

tj90241 (approved):
In that case, awesome! 🚀 

Thanks for looking into that to be sure.

comay (approved):

PR #2135, opened 2023-05-16
[RDTIBCCOPT-135] Clean up deprecated nova configs

jingqgeng:
Found the following deprecated configs:

- vcpu_pin_set (we set it to nil so clean it up for now) ==> 
  - [compute] cpu_dedicated_set
  - [compute] cpu_shared_set
- scheduler_host_subset_size ==> host_subset_size: **NOTE: scheduler_host_subset_size was set only in attribute and not in the templates so it was not actually used**
- send_arp_for_ha: deprecated since 16.0.0 and we're running 21.2.4
- osapi_compute_link_prefix ==> compute_link_prefix
- dhcp_domain ==> moved from [default] to [api]
- [quota] max_age/until_refresh ==> not used since 17.0.0 and will be [removed in xena](https://github.com/openstack/nova/blob/b3fdd7ccf01bafb68e37a457f703b79119dbfa86/releasenotes/notes/remove-quota-options-0e407c56ea993f5a.yaml#L6)

Reference: https://docs.openstack.org/nova/ussuri/configuration/config.html

### DIFF
--- a/chef/cookbooks/bcpc/templates/default/nova/nova.conf.erb
+++ b/chef/cookbooks/bcpc/templates/default/nova/nova.conf.erb
@@ -13,22 +13,18 @@ cert = /etc/nova/ssl-bcpc.pem
 my_ip = <%= node['service_ip'] %>
 enable_new_services = false
 cpu_allocation_ratio = <%= node['bcpc']['nova']['cpu_allocation_ratio'] %>
-<% if node['bcpc']['nova']['vcpu_pin_set'] %>
-vcpu_pin_set =  <%= node['bcpc']['nova']['vcpu_pin_set'] %>
-<% end %>
 force_config_drive = true
 ram_allocation_ratio = <%= node['bcpc']['nova']['ram_allocation_ratio'] %>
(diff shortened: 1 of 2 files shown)


chef/cookbooks/bcpc/templates/default/nova/nova.conf.erb
```diff
@@ -161,6 +155,7 @@ workers = <%= node['bcpc']['openstack']['services']['workers'] %>
 available_filters = <%= available_filter %>
 <% end %>
 enabled_filters = <%= @enabled_filters.join(',') %>
+host_subset_size = <%= node['bcpc']['nova']['scheduler']['host_subset_size'] %>
```
tj90241, line 158: This was a great find - I think it's going to help with the big clusters a bit.

tj90241 (approved):

chef/cookbooks/bcpc/templates/default/nova/nova.conf.erb
jingqgeng: Yes! This will help mitigate the problem when a hypervisor comes online it always gets picked by nova

PR #2136, opened 2023-05-16
configure-sshd: Converge libvirt/physical config

tj90241:
Previously, we configure virtualbox/libvirt/bare metal
slightly differently. This commit converges the
configuration of libvirt/bare metal, and also supports
the case where ansible_host != the service IP.

### DIFF
--- a/ansible/playbooks/roles/common/templates/ssh/sshd_config.j2
+++ b/ansible/playbooks/roles/common/templates/ssh/sshd_config.j2
@@ -18,13 +18,13 @@ Protocol 2
 # What ports, IPs and protocols we listen for
 Port 22
 {% if ansible_virtualization_type == 'virtualbox' %}
-ListenAddress {{ virtualbox['nat_ip'] }}
-{% elif libvirt_kvm_guest %}
-ListenAddress {{ ansible_eth0.ipv4.address }}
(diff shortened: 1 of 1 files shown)


mboruta1 (approved):

ansible/playbooks/roles/common/templates/ssh/sshd_config.j2
```diff
@@ -19,13 +19,13 @@ Protocol 2
 Port 22
 {% if ansible_virtualization_type == 'virtualbox' %}
 ListenAddress {{ virtualbox['nat_ip'] }}
-{% elif libvirt_kvm_guest %}
-ListenAddress {{ ansible_eth0.ipv4.address }}
-{% endif %}
-ListenAddress {{ interfaces['service']['ip'] }}
-{% for transit in transit_interfaces -%}
-ListenAddress {{ transit['ip'] | ipaddr('address') }}
+{% else %}
```
comay, line 22: Doesn't this mean that with VirtualBox now the only addresses listed is the NAT IP? In other works, there's no transit or service IPs listed.

tj90241: yup, fixed!

comay (approved):

PR #2139, opened 2023-05-19
Use `--system-site-packages` in Rally for more consistent Python environment

comay:
RDTIBCC-4569

Rally now installs cleanly again and an internal CI/CD Rally run appears successful.

The approach I took was to start with the minimum virtual environment created with `--system-site-packages` and then added any updates. Through trial and error and testing, the additional Python packages that I needed to install using `pip(1)` were `pyOpenSSL` (but I pinned it to less than the current major version) and Tyler's earlier pin of `SQLAlchemy` (which needed to be higher that the system version but no higher than before).


mboruta1 (approved):
Nice!

PR #2142, opened 2023-05-25
RDTIBCCOPT-79 Apache httpd tuning

jingqgeng:
Made the following changes:
- Increase `ServerLimit`, the hard limit of the number of processes created for a module, to 32, which aligns with to the change I made to increase the number of openstack service workers. 
- Increase `MaxRequestWorkers`, the max number of concurrent connections Apache will serve, to 800: this is calculated per ServerLimit (32) * ThreadsPerChild (25) = 800. See the documentation [here](https://httpd.apache.org/docs/2.4/mod/mpm_common.html#maxrequestworkers).
- Set `GracefulShutdownTimeout` to 900. This should prevent threads-leak after graceful reboots. See the "Graceful process termination and Scoreboard usage" in this [page](https://httpd.apache.org/docs/2.4/mod/event.html#page-header). I set this to a large number to avoid killing threads when they're still busy for legit reasons. 

Note: `ServerLimit` and `MaxRequestWorkers` set the HARD LIMIT for processes and threads respectively. The actual number of threads is still dynamically managed by Apache, which means we will have anywhere from 25 (set by `min_spare_threads`) to 75 (set by `max_spare_threads`) idle threads ready to serve new requests.


mboruta1 (approved):
Nice!

PR #2144, opened 2023-05-31
Upgrade to Calico 3.26

comay:
RDTIBCC-4354

This PR upgrades Calico from `3.22.5` to `3.26.0` and reduces the number of patched versions of files we carry down. To that goal, I also took care of removing one additional such file that we needed for Bionic but upgrading the version of `etcd3gw` that we use in that one particular case from `0.2.6` to `1.0.2`.

Testing continues but sanity testing `1h1w` & `3h3w` builds via CI/CD produced clean builds including for Rally.


chef/cookbooks/bcpc/attributes/etcd3gw.rb
```diff
@@ -3,6 +3,6 @@
 ###############################################################################
 
 default['bcpc']['etcd3gw']['remote_file'] = {
-  'file' => 'etcd3gw-0.2.6.tar.gz',
-  'checksum' => '4a765a382ae18ebd4fccbc1388d1ac5226db0540bccd1f081052600df89145fd',
+  'file' => 'etcd3gw-1.0.2.tar.gz',
+  'checksum' => '8a1482c6993cb3032ea9bee76c78f18d438730f006d772df6bcc8b330fa2740e',
```
tj90241, line 7: This is only used for Bionic IIRC - does it _need_ to be changed as part of this commit? Have we tested changing it on Bionic, or can we be rest assured that it won't break Bionic somehow?

tj90241: Looks like it might be needed? I dislike the fact we have to change this without assurance... but so be it, maybe:
https://github.com/projectcalico/calico/blob/232174be3b759e1a781e125e61daf395cb10c360/networking-calico/requirements.txt#L8

tj90241:
On the Yoga branch, we need to drop the `config.py` patch as this has fixes contained within I think. That's not here, but V/W/X/Y all need changes now.... erg. @npustchi @chaowang987

chef/cookbooks/bcpc/attributes/etcd3gw.rb
comay: I'm testing Bionic as well in all this and I've also looked at the differences between the two `etcd3gw` versions and the risk should be minimal.

comay:
> On the Yoga branch, we need to drop the `config.py` patch as this has fixes contained within I think. That's not here, but V/W/X/Y all need changes now.... erg. @npustchi @chaowang987

Yes, that's additional cleanup I would be happy to take on once this makes it in.

tj90241 (approved):
Would like to see if we can cut a release and get it on the train this week. Let's chat Monday.

psipika:
> > On the Yoga branch, we need to drop the `config.py` patch as this has fixes contained within I think. That's not here, but V/W/X/Y all need changes now.... erg. @npustchi @chaowang987
> 
> Yes, that's additional cleanup I would be happy to take on once this makes it in.

Please keep me in the loop and I'll take a look to review.

psipika (approved):
LGTM 👍

PR #2147, opened 2023-06-05
Adjust patched `ceph-deploy` file to use official upstream change

comay:
While looking at the patches we were carrying with an eye towards Yoga, I noticed that the `ceph-deploy` issue had been fixed upstream and also our version of the file incorrectly has an Apache license added to it while the underlying component is actually [MIT licensed](https://github.com/ceph/ceph-deploy/blob/master/LICENSE). This PR is to use the upstream patch for Python 3.8 and to reflect the correct license. And since this is effectively the upstreamed file, I added it to the `flake8` exception list and left it as-is.

### DIFF
--- a/.github/scripts/run_linters.sh
+++ b/.github/scripts/run_linters.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -x
 
-# Copyright 2022, Bloomberg Finance L.P.
+# Copyright 2023, Bloomberg Finance L.P.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


tj90241 (changes requested):
Unless there's a benefit, I do not think we should be changing Ceph components such as this (which are already deprecated). We use the one packaged from Canonical as is... is something with it not working as expected?

comay:
> Unless there's a benefit, I do not think we should be changing Ceph components such as this (which are already deprecated). We use the one packaged from Canonical as is... is something with it not working as expected?

Correct, we're using the Canonical package with one file patched for Python 3.8 compatibility - I'm just using the official fix from `ceph/ceph-deploy` to the one that Jing came up with. It's a very minor cleanup to align better with the (final) `ceph-deploy` release that on GitHub.

tj90241:
Was there anything _functionally_ problematic with the code, though?

Most of the changes within this PR are linting-related -- the irony being that _said changes now made the linter we use upset_ ... and so we have to add an exclusion to our linting configuration that was formerly not present!

I could see the changes isolate to drop the `distro` module maybe... but I don't agree pulling in any changes that work against our linter when the logic itself is not changing (unless I'm mistaken).

comay:
The logic is slightly different in how it’s accounting for the deprecation in Python 3.8 but otherwise it’s doing the same sort of decision making. And yes, it’s unfortunate to add it to the exception list but I see that as the norm for upstream code that we’re patching. Even if it was flake8-clean, I think adding it to the exception list makes sense since it is upstream code.

This isn’t big win in any way just a desire to align this with the other code we patch (minimally).

comay:
Closing this as I have a different approach in mind.

PR #2171, opened 2023-06-26
Enable cinder image cache for non-ceph backends

jingqgeng:
With the image cache enabled, cinder will download images from glance when the image is used for the first time. After that, cinder will create a smallest possible volume based on the image, and keep this information in the db. When another volume is created with the same image, Cinder will query the db, find out there's a volume containing this image, and simply create a larger volume by extending that volume. See more details [here in cinder documentation](https://docs.openstack.org/cinder/latest/admin/image-volume-cache.html) and a case study [here](https://www.purestorage.com/content/dam/purestorage/pdf/whitepapers/Pure_Storage_Glance_Image_Cache_OpenStack_wp.pdf)

For backends other than Ceph, this will drastically improve the volume creating time.


jingqgeng:
Close for now. Will push this change again when lightbits is ready.

PR #2172, opened 2023-06-26
Small adjustment to the SolarFlare scope to improve performance and reduce TCP retransmissions

desrod:
After a battery of tests with/without `sfboot` from the `sfutil` package with no significant change in the TCP retransmission scope, I pivoted to using a different TCP congestion algorithm, which markedly changed the TCP retransmission rate in a positive way. This PR represents the 2-line change needed to implement that.


desrod:
Closing this PR for rework in a way that doesn't affect the TCP congestion algorithm in a negative way.

PR #2173, opened 2023-06-29
[RDTIBCCOPT-161] Bug-fix for Cinder deferred volume deletion 

jingqgeng:


tj90241:
I did a diff against the Yoga build:
```
--- /usr/lib/python3/dist-packages/cinder/volume/drivers/rbd.py	2023-01-26 12:02:51.000000000 +0000
+++ rbd.py	2023-06-29 19:57:43.348529288 +0000
@@ -412,11 +412,7 @@
             for vol in self.RBDProxy().trash_list(client.ioctx):
                 try:
                     self.RBDProxy().trash_remove(client.ioctx, vol.get('id'))
-                except OSError as e:
-                    # NOTE(arne_wiebalck): trash_remove raises EPERM in case
-                    # the volume's deferral time has not expired yet, so we
-                    # want to explicitly handle this "normal" situation.
-                    # All other exceptions, e.g. ImageBusy, are not re-raised
+                    # ALL other exceptions, e.g. ImageBusy, are not re-raised
                     # so that the periodic purge retries on the next iteration
                     # and leaves ERRORs in the logs in case the deletion fails
                     # repeatedly.
@@ -432,6 +428,13 @@
                     LOG.info("Deleted %s from trash for backend '%s'",
                              vol.get('name'),
                              self._backend_name)
+                    _, parent, parent_snap = self._get_clone_info(self.rbd.Image(client.ioctx,
+                                                                                 vol.get('name')),
+                                                                  vol.get('name'))
+                    if parent:
+                      LOG.info("Volume '%s' is a clone so cleaning references", vol.get('name'))
+                      parent_snap = typing.cast(str, parent_snap)
+                      self._delete_clone_parent_refs(client, parent, parent_snap)
 
     def _start_periodic_tasks(self) -> None:
         if self.configuration.enable_deferred_deletion:
@@ -1340,8 +1343,9 @@
                     return
 
                 # If it is a clone, walk back up the parent chain deleting
-                # references.
-                if parent:
+                # references. If deferred deletion is enabled, delete references
+                # when the volume is purged.
+                if parent and not self.configuration.enable_deferred_deletion:
                     LOG.debug("volume is a clone so cleaning references")
                     parent_snap = typing.cast(str, parent_snap)
                     self._delete_clone_parent_refs(client, parent, parent_snap)
```

The removal of the `except OSError as e:` looks unintentional -- the next line of code past the comment block is `if (e.errno == errno.EPERM):`... ?

jingqgeng:
> I did a diff against the Yoga build:
> 
> ```
> --- /usr/lib/python3/dist-packages/cinder/volume/drivers/rbd.py	2023-01-26 12:02:51.000000000 +0000
> +++ rbd.py	2023-06-29 19:57:43.348529288 +0000
> @@ -412,11 +412,7 @@
>              for vol in self.RBDProxy().trash_list(client.ioctx):
>                  try:
>                      self.RBDProxy().trash_remove(client.ioctx, vol.get('id'))
> -                except OSError as e:
> -                    # NOTE(arne_wiebalck): trash_remove raises EPERM in case
> -                    # the volume's deferral time has not expired yet, so we
> -                    # want to explicitly handle this "normal" situation.
> -                    # All other exceptions, e.g. ImageBusy, are not re-raised
> +                    # ALL other exceptions, e.g. ImageBusy, are not re-raised
>                      # so that the periodic purge retries on the next iteration
>                      # and leaves ERRORs in the logs in case the deletion fails
>                      # repeatedly.
> @@ -432,6 +428,13 @@
>                      LOG.info("Deleted %s from trash for backend '%s'",
>                               vol.get('name'),
>                               self._backend_name)
> +                    _, parent, parent_snap = self._get_clone_info(self.rbd.Image(client.ioctx,
> +                                                                                 vol.get('name')),
> +                                                                  vol.get('name'))
> +                    if parent:
> +                      LOG.info("Volume '%s' is a clone so cleaning references", vol.get('name'))
> +                      parent_snap = typing.cast(str, parent_snap)
> +                      self._delete_clone_parent_refs(client, parent, parent_snap)
>  
>      def _start_periodic_tasks(self) -> None:
>          if self.configuration.enable_deferred_deletion:
> @@ -1340,8 +1343,9 @@
>                      return
>  
>                  # If it is a clone, walk back up the parent chain deleting
> -                # references.
> -                if parent:
> +                # references. If deferred deletion is enabled, delete references
> +                # when the volume is purged.
> +                if parent and not self.configuration.enable_deferred_deletion:
>                      LOG.debug("volume is a clone so cleaning references")
>                      parent_snap = typing.cast(str, parent_snap)
>                      self._delete_clone_parent_refs(client, parent, parent_snap)
> ```
> 
> The removal of the `except OSError as e:` looks unintentional -- the next line of code past the comment block is `if (e.errno == errno.EPERM):`... ?

Indeed... I moved around some code and accidently deleted a few lines. Good catch!

comay:
Jing, could you please post a link to the upstream bug report? It would also be good to include a URL in the recipe that is delivering the patched file.

jingqgeng:
> Jing, could you please post a link to the upstream bug report? It would also be good to include a URL in the recipe that is delivering the patched file.

Sure. There're 3 similar bug reports:
https://bugs.launchpad.net/cinder/+bug/1957804
https://bugs.launchpad.net/cinder/+bug/1826115
https://bugs.launchpad.net/cinder/+bug/2012622

I believe they are the same issue I'm trying to fix here.

jingqgeng:
For easy review, here a diff of the change in rbd.py (compared to that in yoga):
```
diff --git a/tmp/gpw2_rbd.py b/chef/cookbooks/bcpc/files/default/cinder/rbd.py
index 8db8a81..5ad2390 100644
--- a/tmp/gpw2_rbd.py
+++ b/chef/cookbooks/bcpc/files/default/cinder/rbd.py
@@ -411,7 +411,21 @@ class RBDDriver(driver.CloneableImageVD, driver.MigrateVD,
         with RADOSClient(self) as client:
             for vol in self.RBDProxy().trash_list(client.ioctx):
                 try:
-                    self.RBDProxy().trash_remove(client.ioctx, vol.get('id'))
+                    # NOTE(jinggeng): clone info can only be retrieved from rbd.Image, and
+                    # generating rbd.Image object from trashed volume will cause an error of
+                    # "RBD image not found". In order to retrieve the clone information,
+                    # the volume needs to be first restored from trash, and then deleted.
+                    # Although RBDProxy().remove is slow, the restored volumes are deleted
+                    # asynchronously so it's still an improvement.
+                    vol_name = utils.convert_str(vol.get('name'))
+                    self.RBDProxy().trash_restore(client.ioctx,
+                                                  vol.get('id'),
+                                                  vol_name)
+                    _, parent, parent_snap = self._get_clone_info(self.rbd.Image(client.ioctx,
+                                                                                 vol_name),
+                                                                  vol_name)
+                    self.RBDProxy().remove(client.ioctx,
+                                           vol_name)
                 except OSError as e:
                     # NOTE(arne_wiebalck): trash_remove raises EPERM in case
                     # the volume's deferral time has not expired yet, so we
@@ -432,6 +446,10 @@ class RBDDriver(driver.CloneableImageVD, driver.MigrateVD,
                     LOG.info("Deleted %s from trash for backend '%s'",
                              vol.get('name'),
                              self._backend_name)
+                    if parent:
+                        LOG.info("Volume '%s' is a clone so cleaning references", vol.get('name'))
+                        parent_snap = typing.cast(str, parent_snap)
+                        self._delete_clone_parent_refs(client, parent, parent_snap)
 
     def _start_periodic_tasks(self) -> None:
         if self.configuration.enable_deferred_deletion:
@@ -1340,8 +1358,9 @@ class RBDDriver(driver.CloneableImageVD, driver.MigrateVD,
                     return
 
                 # If it is a clone, walk back up the parent chain deleting
-                # references.
-                if parent:
+                # references. If deferred deletion is enabled, delete references
+                # when the volume is purged.
+                if parent and not self.configuration.enable_deferred_deletion:
                     LOG.debug("volume is a clone so cleaning references")
                     parent_snap = typing.cast(str, parent_snap)
                     self._delete_clone_parent_refs(client, parent, parent_snap)

```

tj90241:
I understand you're un-trashing a volume in order to get it's clone information. But can/what if the clone is trashed by a scheduled RBD operation outside of the scope of `cinder`, and if so what happens as the `parent` and `parent_snap` information are irrecoverable in that case? This is a controllable factor in our environment, but I am guessing that upstream will question you on this topic as they do not like to assume?

```
[operations@ts-yoga-test-cloud-r001n02-h ~]$ openstack volume create --size 100 --type ceph tyler-ceph
[operations@ts-yoga-test-cloud-r001n02-h ~]$ openstack volume create --size 100 --type ceph --source ab1e2fdd-b433-4235-8b7e-79fcda6e723f tyler-ceph-clone
# => creates clone volume 9710ab26-c8ca-4616-ac9c-573579fa2b12 (tyler-ceph-clone) from parent ab1e2fdd-b433-4235-8b7e-79fcda6e723f (tyler-ceph) 

[operations@ts-yoga-test-cloud-r001n02-h ~]$ openstack volume delete  9710ab26-c8ca-4616-ac9c-573579fa2b12
[operations@ts-yoga-test-cloud-r001n02-h ~]$ openstack volume show  9710ab26-c8ca-4616-ac9c-573579fa2b12
# => volume goes to error_deleting state as we expect

# Meanwhile... the operator may have scheduled a trash purge, meaning we lose the reference:
root@ts-yoga-test-cloud-r001n05-h:~# rbd --pool volumes trash list
c3ab4eee45bb6 volume-9710ab26-c8ca-4616-ac9c-573579fa2b12
root@ts-yoga-test-cloud-r001n05-h:~# rbd --pool volumes trash purge
Removing images: 100% complete...done.
root@ts-yoga-test-cloud-r001n05-h:~# rbd --pool volumes trash list
root@ts-yoga-test-cloud-r001n05-h:~# 
```

Have you thought about flattening the child volume @ deletion time in order to remove the clone reference? It would probably be slower, but seems safer as it avoids the situation in which the clone reference can be lost by an out of band operation.

And although it's a 1/1,000,000 chance - what happens when untrashing  the volume and Ceph explodes before remove is called?

tj90241:
In terms of removing ourselves from the current situation -- I think we should be able to scan if any volume is in `error_deleting` state. If it is not (likely case?), then it can be said that no children were attempted to have been deferred-deleted prior to the parent being deleted.

This would leave the case where the parent is deleted before the child, but in that case I see that the volume is renamed to `volume-<uuid>.deleted` in the pool and not moved into the trash, so it should also be recognizable?

jingqgeng:
> deleted




> I understand you're un-trashing a volume in order to get it's clone information. But can/what if the clone is purged* by a scheduled RBD operation outside of the scope of `cinder`, and if so what happens as the `parent` and `parent_snap` information are irrecoverable in that case? This is a controllable factor in our environment, but I am guessing that upstream will question you on this topic as they do not like to assume?
> 
> ```
> [operations@ts-yoga-test-cloud-r001n02-h ~]$ openstack volume create --size 100 --type ceph tyler-ceph
> [operations@ts-yoga-test-cloud-r001n02-h ~]$ openstack volume create --size 100 --type ceph --source ab1e2fdd-b433-4235-8b7e-79fcda6e723f tyler-ceph-clone
> # => creates clone volume 9710ab26-c8ca-4616-ac9c-573579fa2b12 (tyler-ceph-clone) from parent ab1e2fdd-b433-4235-8b7e-79fcda6e723f (tyler-ceph) 
> 
> [operations@ts-yoga-test-cloud-r001n02-h ~]$ openstack volume delete  9710ab26-c8ca-4616-ac9c-573579fa2b12
> [operations@ts-yoga-test-cloud-r001n02-h ~]$ openstack volume show  9710ab26-c8ca-4616-ac9c-573579fa2b12
> # => volume goes to error_deleting state as we expect
> 
> # Meanwhile... the operator may have scheduled a trash purge, meaning we lose the reference:
> root@ts-yoga-test-cloud-r001n05-h:~# rbd --pool volumes trash list
> c3ab4eee45bb6 volume-9710ab26-c8ca-4616-ac9c-573579fa2b12
> root@ts-yoga-test-cloud-r001n05-h:~# rbd --pool volumes trash purge
> Removing images: 100% complete...done.
> root@ts-yoga-test-cloud-r001n05-h:~# rbd --pool volumes trash list
> root@ts-yoga-test-cloud-r001n05-h:~# 
> ```
> 
> Have you thought about flattening the child volume @ deletion time in order to remove the clone reference? (EDIT: or just bypassing deferred deletion outright in such cases?) It would probably be slower, but seems safer as it avoids the situation in which the clone reference can be lost by an out of band operation.
> 
> And although it's a 1/1,000,000 chance - what happens when untrashing the volume and Ceph explodes before remove is called?

If trash is purged outside of Cinder, then when Cinder runs this periodic task, the purged volume just won't appear in this list. So a volume is a clone,  the parent (temporary snapshot) carry around that dead reference and cannot be deleted in Cinder. I'm not sure how flatten a volume at deletion time will help with this issue? From my understanding, flatten is like a deep copy that moves everything into a clone. But I don't think that automatically cleans up the references in the parents? 

If un-trashing failed, an exception will be thrown out and caught, and the parent snapshot will get into the same zombie state.

jingqgeng:
> In terms of removing ourselves from the current situation -- I think we should be able to scan if any volume is in `error_deleting` state. If it is not (likely case?), then it can be said that no children were attempted to have been deferred-deleted prior to the parent being deleted.
> 
> This would leave the case where the parent is deleted before the child, but in that case I see that the volume is renamed to `volume-<uuid>.deleted` in the pool and not moved into the trash, so it should also be recognizable?

If a parent is deleted before its children, Cinder will keep it as a silent volume and delete it when all the children are gone: https://github.com/openstack/cinder/blob/stable/yoga/cinder/volume/drivers/rbd.py#L1352

jingqgeng:
The proposal https://review.opendev.org/c/openstack/cinder/+/900111 was rejected by the upstream and they have another fix that's merged: https://review.opendev.org/c/openstack/cinder/+/835384

PR #2179, opened 2023-07-05
Add cinder QoS policy support

mzhang741:
Add chef automation to create or delete a cinder qos policy with the specified parameters 

**Describe your changes**
Add a new section in cinder.rb to create or delete a qos policy based on the attributes defined in chef_environment

**Testing performed**

- qos enabled -> successfully create a qos policy and associate it with the volume type defined in chef_environment

- qos disabled -> qos policy is disassociated from the volume type and deleted

- chef-client cinder recipe twice when qos is disabled -> qos is deleted after the first run and second run completed successfully without error 

- chef-client cinder recipe twice when qos is enabled -> both runs completed successfully and only one qos is created

- failed to execute the command to create a qos policy -> raise an exception and exit with error 

- run cinder recipe second time after failing to create a qos policy -> run finished with a policy being created and associated with the volume type

- failed to execute the command to associate the qos policy -> raise an exception and exit with error after the policy is created

- run cinder recipe second time after failing to associate the policy to the volume type -> run completes successfully, no new policy is created and current policy is associated with the volume type

- failed to execute the command to disassociate a qos policy -> raise an exception and exit with error 

- run cinder recipe second time after failing to disassociate the policy -> disassociate and delete the policy successfully. Run completed without errors. 

- failed to execute the command to delete a qos policy -> raise an exception and exit with error after disassociate the policy

- run cinder recipe second time after failing to delete the policy -> run complete successfully with qos policy being deleted.

### DIFF
--- a/ansible/group_vars/all/chef_environment.yml.example
+++ b/ansible/group_vars/all/chef_environment.yml.example
@@ -27,6 +27,28 @@ chef_environment:
         ceph:
           pool:
             size: 2
+        qos:
+          enabled: true
+          volume_types:
+            - name: ceph
+              limits:
+                # for fixed iops per volume
+                read_iops_sec: 1000000
+                write_iops_sec: 1000000
+                total_iops_sec:
+                # for burst iops per volume
+                read_iops_sec_max:
+                write_iops_sec_max:
+                # for fixed bandwidth per volume
+                read_bytes_sec:
+                write_bytes_sec:
+                total_bytes_sec:
+                # for burst bandwidth per volume
+                read_bytes_sec_max:
+                write_bytes_sec_max:
+                total_bytes_sec_max:
+                # for burst bucket size:
+                size_iops_sec:
       cloud:
         domain: "{{ cloud_domain }}"
         fqdn: "{{ cloud_fqdn }}"
--- a/ansible/group_vars/all/chef_environment.yml.example
+++ b/ansible/group_vars/all/chef_environment.yml.example
@@ -33,22 +33,19 @@ chef_environment:
             - name: ceph
               limits:
                 # for fixed iops per volume
-                read_iops_sec: 1000000
-                write_iops_sec: 1000000
-                total_iops_sec:
+                read_iops_sec: 2000
+                write_iops_sec: 1000
                 # for burst iops per volume
-                read_iops_sec_max:
-                write_iops_sec_max:
+                read_iops_sec_max: 4000
+                write_iops_sec_max: 2000
                 # for fixed bandwidth per volume
-                read_bytes_sec:
-                write_bytes_sec:
-                total_bytes_sec:
+                read_bytes_sec: 1000000000
+                write_bytes_sec: 500000000
                 # for burst bandwidth per volume
-                read_bytes_sec_max:
-                write_bytes_sec_max:
-                total_bytes_sec_max:
+                read_bytes_sec_max: 2000000000
+                write_bytes_sec_max: 1000000000
                 # for burst bucket size:
-                size_iops_sec:
+                size_iops_sec: 4096
       cloud:
         domain: "{{ cloud_domain }}"
         fqdn: "{{ cloud_fqdn }}"
--- a/chef/cookbooks/bcpc/recipes/cinder.rb
+++ b/chef/cookbooks/bcpc/recipes/cinder.rb
@@ -371,3 +371,77 @@ cinder_config.backends.each do |backend|
     not_if { node.run_state['os_vol_type_props'].dig(backend_name, 'volume_backend_name') == backend_name }
   end
 end
+
+# cinder qos 
+node['bcpc']['cinder']['qos']['volume_types'].each do |volume_type|
+
+  # create a qos
+  execute "create qos" do
+    # construct openstack command based on the qos attributes
+    qos_name = "#{volume_type['name']}-qos"
+    qos_create_opts = []
+
+    # Iterate over each key and value in the limits
+    # construct openstack argument
+    # add it to qos_create_opts
+    volume_type.fetch('limits', []).each do |key, value|
+      qos_create_opts.push("--property #{key}=#{value}")
+    end
+
+    # execute the openstack command
+    environment os_adminrc
+    retries 3
+    command <<-DOC
+      openstack volume qos create \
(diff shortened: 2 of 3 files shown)


chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -371,3 +371,67 @@
...
+      # associate the qos policy to the volume type
+      cmd = "openstack volume qos associate \
+        #{qos_name} \
+        #{volume_type['name']}"
+      cmd_out = shell_out(cmd, env: os_adminrc)
+
+      raise 'unable to associate the qos policy' if cmd_out.error?
+    end
+
+    only_if { node['bcpc']['cinder']['qos']['enabled'] }
```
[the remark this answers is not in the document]
mzhang741: i think its ok to be merged in since the code itself is only gonna take effect if we set the override values in chef_environment.yml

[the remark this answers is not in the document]
jingqgeng: Clarified offline that we don't need `do not merge`

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -371,3 +371,67 @@
...
+  end
+
+  ruby_block 'clean up the qos policy' do
+    qos_name = "#{volume_type['name']}-qos"
```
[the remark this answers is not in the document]
mzhang741: The user decides to set what policy on which volume type in chef_environment. For volume_types that are not specified there, we don't create any policies.

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -371,3 +371,67 @@
...
+        cmd = "openstack volume qos create \
+          #{qos_create_opts.join(' ')} \
+          #{qos_name}"
+        cmd_out = shell_out(cmd, env: os_adminrc)
+
+        raise 'unable to create a qos policy' if cmd_out.error?
+      end
+
+      # associate the qos policy to the volume type
+      cmd = "openstack volume qos associate \
```
mboruta1, line 401: Will this attempt to associate the same policy multiple times to the given volume type? If QoS is enabled and a QoS policy has been created and associated, subsequent runs will still always execute `openstack volume qos show #{qos_name}` and `openstack volume qos associate #{qos_name} #{volume_type['name']}`, right?

If so, I think it is cleaner to just assume that if a policy with the given name exists, it is also associated. Users should not be modifying these policies outside of automation regardless.

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -371,3 +371,67 @@
...
+        #{volume_type['name']}"
+      cmd_out = shell_out(cmd, env: os_adminrc)
+
+      raise 'unable to associate the qos policy' if cmd_out.error?
+    end
+
+    only_if { node['bcpc']['cinder']['qos']['enabled'] }
+  end
+
+  ruby_block 'clean up the qos policy' do
```
mboruta1, line 412: Why not have the name of the block mirror `create a qos policy for the volume type #{volume_type['name']}`?

mzhang741: make sense

chef/cookbooks/bcpc/recipes/cinder.rb
mzhang741: agree. Even if the associate step fails, we could disable the qos and the automation would clean up the qos policy that is unassociated.

mboruta1: Nice, thank you!

mboruta1 (dismissed):
Nice!

comay (commented):
It would be nice if a commented-out example could be added to

ansible/group_vars/all/overrides.yml.example

referencing, for example, the default "ceph" volume type.

ansible/group_vars/all/chef_environment.yml.example
```diff
@@ -27,6 +27,36 @@ chef_environment:
...
+                read_bytes_sec:
+                write_bytes_sec:
+                total_bytes_sec:
+                # for burst bandwidth per volume
+                read_bytes_sec_max:
+                write_bytes_sec_max:
+                total_bytes_sec_max:
+                # for burst bucket size:
+                size_iops_sec:
+            - name: dev-lightos
```
jingqgeng, line 52: Not sure how we feel about mentioning the specific backend names other than `ceph`. Maybe we can use `cinder-backend-0`, `cinder-backend-1`, etc? Or having 1 example for `ceph` is enough?

mzhang741: good point

ansible/group_vars/all/chef_environment.yml.example
```diff
@@ -27,6 +27,28 @@ chef_environment:
...
             size: 2
+        qos:
+          enabled: true
+          volume_types:
+            - name: ceph
+              limits:
+                # for fixed iops per volume
+                read_iops_sec: 1000000
+                write_iops_sec: 1000000
+                total_iops_sec:
```
mboruta1, line 38: Please fill in values for the remaining options, example config should ideally not have syntax errors.

ansible/group_vars/all/chef_environment.yml.example
comay: I think just having a entry for `ceph` is suitable.

ansible/group_vars/all/chef_environment.yml.example
comay: Agreed.

mboruta1 (approved):
Nice work!

ansible/group_vars/all/chef_environment.yml.example
```diff
@@ -27,6 +27,25 @@ chef_environment:
         ceph:
           pool:
             size: 2
+        qos:
```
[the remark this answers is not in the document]
mzhang741: So there's only one qos section, but we can define different policies for different volume types. For each volume type, we just create an entry under the volume_types and specify the limits, like iops, in that entry.

npustchi (approved):
LGTM

PR #2181, opened 2023-07-07
nova: Fix availability-zone scheduling in Yoga

tj90241:
As the name suggests, query_placement_for_availability_zone
queries placement for availability zone filtering (instead
of doing it within Nova, as in prior releases). This is
default on as of Xena.

Unfortunately, we either have not configured this
correctly, or it's not working in larger deployments, as
I observed the following:

This works, of course:
openstack server create --min 1 --max 1 \\
    --network dev --image cirros --flavor dev-generic1.tiny \\
    --wait put-me-anywhere

But this only works with this PR applied:
openstack server create --min 1 --max 1 \\
    --network dev --image cirros --flavor dev-generic1.tiny \\
    --availability-zone dev-1 --wait put-me-in-az-1


tj90241:
Ironically, this seems to work fine in BVE without the flag. But not so much in larger deployments...?

npustchi (approved):
LGTM

PR #2183, opened 2023-07-16
Quincy fine-tuning

tj90241:


tj90241:
Still seeing some regressions even with this, I think...

chef/cookbooks/bcpc/templates/default/ceph/ceph.conf.erb
```diff
@@ -67,12 +67,13 @@ osd max backfills  = <%= @node['bcpc']['ceph']['osd_max_backfills'] %>
...
-# backs bluestore.  Use bitmap allocators at the expense of memory consumption
-# to avoid degrading I/O performance in such cases.
-bluefs_allocator = bitmap
-bluestore_allocator = bitmap
+# In Quincy 17.2.5, the bitmap (and by extension, hybrid when it falls back to
+# bitmap internally) allocators can have tremendous p99_* latency degredation
+# when they have to scan L0 tables. The AVL allocator uses a bit more memory,
+# but is much more consistent in the worst case since the bluestore_avl_alloc_
+# ff_max_* options were added.
+bluefs_allocator = avl
```
jingqgeng, line 75: May not directly related to this regression - have we considered/tried hybrid?

tj90241: Hybrid is effectively AVL, but fallback to bitmap for certain cases.

I tried AVL, but it segfaulted when loading an OSD with production data.

Though, I am re-considering the second commit in this PR... it may have an impact...

PR #2189, opened 2023-07-21
Quincy fixes

tj90241:
Some stability/performance fixes with the Ceph in 8.4.x.

### DIFF
--- a/chef/cookbooks/bcpc/attributes/ceph.rb
+++ b/chef/cookbooks/bcpc/attributes/ceph.rb
@@ -89,16 +89,7 @@ default['bcpc']['ceph']['osd_memory_target'] = 8589934592
 default['bcpc']['ceph']['mon_osd_down_out_interval'] = 300
 
 # BlueStore tuning
-default['bcpc']['ceph']['bluestore_rocksdb_options'] = [
-  'compression=kNoCompression',
-  'max_write_buffer_number=4',
-  'min_write_buffer_number_to_merge=1',
-  'recycle_log_file_num=4',
-  'write_buffer_size=268435456',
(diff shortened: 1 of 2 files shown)


tj90241:
The individual commits have explanations, FYI.

chef/cookbooks/bcpc/attributes/ceph.rb
```diff
@@ -85,21 +85,9 @@
 default['bcpc']['ceph']['osd_deep_scrub_interval'] = 2592000
 default['bcpc']['ceph']['osd_scrub_max_interval'] = 604800
 default['bcpc']['ceph']['osd_scrub_sleep'] = 0.05
-default['bcpc']['ceph']['osd_memory_target'] = 6442450944
+default['bcpc']['ceph']['osd_memory_target'] = 8589934592
 default['bcpc']['ceph']['mon_osd_down_out_interval'] = 300
 
-# BlueStore tuning
-default['bcpc']['ceph']['bluestore_rocksdb_options'] = [
-  'compression=kNoCompression',
```
mboruta1, line 93: Nit: would advise to just make this an empty array instead of taking everything out; we will likely want to tweak something in this space in the future.

mboruta1 (dismissed):
Nice, all for using default values unless we have documentation + figures to back up a change.

chef/cookbooks/bcpc/attributes/ceph.rb
tj90241: Funny you mention it... there is a follow-up to this where we'll have to weigh-in whether or not to tweak it (much more significantly) in the future.

okay, sounds good.

mboruta1 (approved):

PR #2192, opened 2023-07-26
Adding alternate backends support.

npustchi:
**Changes**
Support multiple backends generically. Operators can add storage backends with specific properties of the backend.

**Test**
Tested on CI/CD.

### DIFF
--- a/.github/scripts/run_linters.sh
+++ b/.github/scripts/run_linters.sh
@@ -23,6 +23,7 @@ function main {
     find . -name "*.sh" -print0 | xargs -0 -t shellcheck;
     find . -name "*.sh" -print0 | xargs -0 -t bashate -e E006;
     find . -name "*.py" \
+        ! -path "./chef/cookbooks/bcpc/files/default/cinder/lightos.py" \
         ! -path \
             "./chef/cookbooks/bcpc/files/default/neutron/external_net_db.py" \
         ! -path "./chef/cookbooks/bcpc/files/default/neutron/model_query.py" \
--- a/chef/cookbooks/bcpc/libraries/zone_config.rb
+++ b/chef/cookbooks/bcpc/libraries/zone_config.rb
@@ -1,7 +1,7 @@
 # Cookbook:: bcpc
 # Library:: zone_config
 #
-# Copyright:: 2020 Bloomberg Finance L.P.
+# Copyright:: 2023 Bloomberg Finance L.P.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -26,6 +26,14 @@ class ZoneConfig
   end
 
   attr_reader :cinder_config, :node, :nova_config, :nova_compute_config
+
+  def alternate_backends
+    @node['bcpc']['cinder']['alternate_backends']['backends']
+  end
+
+  def alternate_backends_enabled?
+    @node['bcpc']['cinder']['alternate_backends']['enabled']
+  end
 
   def databag(id:)
     @data_bag_item.call(@region, id)
@@ -89,6 +97,43 @@ class CinderConfig
     end
 
     backends
+  end
+
+  def alternate_backends
+    alternate_backends = []
+
+    if !@zone_config.enabled? || !@zone_config.alternate_backends_enabled?
+      return alternate_backends
+    end
+
+    zones = @zone_config.zone_attr()
+    databag = @zone_config.databag(id: 'zones')
+    general_backend_configs = Hash[
+      @zone_config.alternate_backends()
+                  .map { |backend| [backend[:name].to_s, backend] }]
+    zones.each do |zone|
+      next if zone['cinder']['alternate_backends'].nil?
+      backends = zone['cinder']['alternate_backends']
+      backend_databags = Hash[
+      databag[zone['zone']]['alternate-backends']
+                         .map { |b| [b[:name].to_s, b] }]
+      backends.each do |backend|
+        next unless general_backend_configs[backend['name']]['enabled']
+        alternate_backends.append({
+          'backend_name' => backend['backend_name'],
+          'volume_driver' => general_backend_configs[backend['name']]['volume_driver'],
+          'properties' => {}.merge(
+          general_backend_configs[backend['name']]['properties'])
+              .merge(backend_databags[backend['name']]['properties'])
+              .merge(backend['properties'] || {}),
+          'private' => backend['private'],
+          'volume_type_properties' => {}.merge(backend['volume_type_properties'])
+            .map { |k, v| "--property #{k}=#{v}" }.join(' '),
+        })
+      end
+    end
+
+    alternate_backends
   end
 
   def ceph_clients
--- a/chef/cookbooks/bcpc/recipes/cinder.rb
+++ b/chef/cookbooks/bcpc/recipes/cinder.rb
@@ -19,6 +19,7 @@ region = node['bcpc']['cloud']['region']
 config = data_bag_item(region, 'config')
 zone_config = ZoneConfig.new(node, region, method(:data_bag_item))
 cinder_config = zone_config.cinder_config
+alternate_backends = cinder_config.alternate_backends
 
 mysqladmin = mysqladmin()
 psqladmin = psqladmin()
@@ -165,6 +166,20 @@ directory '/etc/cinder/policy.d' do
   action :create
 end
 
+# add lightos active/active driver
+if zone_config.alternate_backends_enabled?
+  cookbook_file '/usr/lib/python3/dist-packages/cinder/volume/drivers/lightos.py' do
+    source 'cinder/lightos.py'
+    notifies :run, 'execute[py3compile-cinder]', :immediately
+    notifies :restart, 'service[cinder-volume]', :delayed
+  end
+
+  execute 'py3compile-cinder' do
+    action :nothing
+    command 'py3compile -p python3-cinder'
+  end
+end
+
 # add AccessList filter and update cinder entry_points.txt
 if zone_config.enabled?
   cookbook_file '/usr/lib/python3/dist-packages/cinder/scheduler/filters/access_filter.py' do
@@ -213,7 +228,9 @@ template '/etc/cinder/cinder.conf' do
     backends: cinder_config.backends,
     config: config,
     headnodes: headnodes(all: true),
+    alternate_backends_enabled:  zone_config.alternate_backends_enabled?,
     rmqnodes: rmqnodes(all: true),
+    alternate_backends: cinder_config.alternate_backends,
     scheduler_default_filters: cinder_config.filters
   )
 
@@ -342,8 +359,7 @@ end
 
 cinder_config.backends.each do |backend|
   backend_name = backend['name']
-  create_args = []
-  create_args.append(backend_name)
+  create_args = [].append(backend_name)
 
   if backend['private']
     create_args.append('--private')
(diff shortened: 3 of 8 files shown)


chef/cookbooks/bcpc/libraries/zone_config.rb
```diff
@@ -91,6 +95,30 @@ def backends
...
+      backend_config = @zone_config.alternate_backend()
+      zones.each do |zone|
+        next if zone['cinder']['alternate_backend'].nil?
+        backend = zone['cinder']['alternate_backend']
+        alternate_backends.append({
+          'name' => backend['name'],
+          'private' => backend['private'],
+          'project_key' => backend_config['project_key'],
+          'project_name' => backend['project_name'],
+          'jwt_key' => backend_config['jwt_key'],
```
tj90241, line 113: Only a specific Cinder backend uses a JWT key...

chef/cookbooks/bcpc/templates/default/cinder/cinder.conf.erb
```diff
@@ -80,3 +80,15 @@ rbd_store_chunk_size = 4
...
 <% end %>
+<% if @alternate_backend['enabled'] %>
+<% @alternate_backends.each do |backend| %>
+[<%= backend['name'] %>]
+volume_driver = <%= @alternate_backend['volume_driver'] %>
+volume_backend_name = <%= backend['name'] %>
+<% @alternate_backend['properties'].each do |property_name, property_value| %>
+<%= property_name %> = <%= property_value.is_a?(Array) ? property_value.join(',') : property_value %>
+<% end %>
+<%= "#{backend['jwt_key']} = #{backend['jwt_value']}" %>
```
tj90241, line 91: Same comment about JWT key here. Is there as reason this could not be a property?

.github/scripts/run_linters.sh
```diff
@@ -28,6 +28,7 @@ function main {
         ! -path "./chef/cookbooks/bcpc/files/default/neutron/model_query.py" \
         ! -path "./chef/cookbooks/bcpc/files/default/nova/guest.py" \
         ! -path "./chef/cookbooks/bcpc/files/default/nova/migration.py" \
+        ! -path "./chef/cookbooks/bcpc/files/default/cinder/lightos.py" \
```
tj90241, line 31: nit - please maintain alphabetical ordering.

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -165,6 +166,20 @@
   action :create
 end
 
+# add lightos active/active driver
+if alternate_backend['enabled']
+  cookbook_file '/usr/lib/python3/dist-packages/cinder/volume/drivers/lightos.py' do
```
tj90241, line 171: This file should be patched regardless of whether or not alternate backends are enabled.

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -165,6 +166,20 @@
...
+    source 'cinder/lightos.py'
+    notifies :run, 'execute[py3compile-cinder]', :immediately
+    notifies :restart, 'service[cinder-volume]', :delayed
```
tj90241, line 174: If this restart is delayed, then `cinder-volume` will not restart until after the entire `chef-run` completes.
This means, though, that the LightOS driver will not be patched for active/active support until after the Chef run completes.

I can see this possibly being a problem when a cluster is first-built -- we always start `cinder.conf` with active/active and expect it to come up as part of the first Chef run. This may need to be `:immediately` rather than delayed for this reason - WDYT/have you tested this case?

chef/cookbooks/bcpc/templates/default/cinder/cinder.conf.erb
```diff
@@ -17,7 +17,7 @@ my_ip = <%= node['service_ip'] %>
 transport_url = rabbit://<%= @rmqnodes.map {|n| "#{@config['rabbit']['username']}:#{@config['rabbit']['password']}@#{n['service_ip']}:5672" }.join(',') %>
 glance_api_servers = <%= "https://#{node['bcpc']['cloud']['fqdn']}:9292" %>
 glance_api_insecure = true
-enabled_backends = <%= @backends.map {|b| b['name']}.join(',') %>
+enabled_backends = <%= @alternate_backend['enabled'] ? @backends.map {|b| b['name']}.concat(@alternate_backends.map {|b| b['name']}).join(',') : @backends.map {|b| b['name']}.join(',') %>
```
tj90241, line 20: Probably a tad cleaner to _always_ `join(',')` the Ceph backend with either:
* `@alternate_backends.map {|b| b['name']}`
* empty list

tj90241 (dismissed):
Getting closer...

If we are going to be generic about the additional backend to support, I think we kind of have to truly be generic -- re: not mandating a JWT for additional backends.

.github/scripts/run_linters.sh
npustchi: there is no alphabetical ordering.

chef/cookbooks/bcpc/libraries/zone_config.rb
tj90241: A JWT in and of itself is not, but what if someone wants to configure e.g., a non-LightOS backend? Other backends do not use JWTs -- so what would be supplied for `jwt_key` in that case?

npustchi: the json web token of that backend. every backend should have a secret key,

ansible/playbooks/roles/common/defaults/main/chef.yml
```diff
@@ -16,6 +16,17 @@ chef_environment:
         domain: "{{ cloud_domain }}"
         fqdn: "{{ cloud_fqdn }}"
         vip: "{{ cloud_ip }}"
+      cinder:
+        alternate_backend:
+          enabled: true
+          volume_driver: cinder.volume.drivers.driver.driverName
```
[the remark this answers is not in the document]
npustchi: the default should be false, but otherwise expectation is whom ever adding a backend will replace it with correct dirver.

.github/scripts/run_linters.sh
npustchi: ok, :)

chef/cookbooks/bcpc/recipes/cinder.rb
npustchi: Why do you need to patch a driver if you are not adding related backend ?

ansible/playbooks/roles/common/defaults/main/chef.yml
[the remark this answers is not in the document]
tj90241: if the default is `false`, that works.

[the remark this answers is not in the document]
comay: Navid, I think my suggestion was to not include this here but **_maybe_** provide it as an example in [ansible/group_vars/all/chef_environment.yml.example](https://github.com/bloomberg/chef-bcpc/blob/main/ansible/group_vars/all/chef_environment.yml.example). The idea would be `chef-bcpc` would only provide a sparse definition of the form:

```
chef_environment:
 .
 .
  override_attributes:
    bcpc:
      .
      .
      cinder:
        alternate_backend:
          enabled: false
          properties:
        .
        .
```

but then the remaining properties can be added in via the environment file by adding entries to the `properties` list of the form `X = Y`.

Is that not possible

ansible/playbooks/roles/common/defaults/main/chef.yml
```diff
@@ -16,6 +16,17 @@ chef_environment:
...
+          jwt_key: jwt
+          project_key: project_name
+          properties:
+            api_address:
+              - 10.10.10.10
+            api_port: 443
+            compression: true
```
comay, line 29: All of these are too driver-specific IMO for here.

chef/cookbooks/bcpc/libraries/zone_config.rb
comay: As I mentioned above, this is too driver-specific - it should be part of the overrides that take place which is why `chef-bcpc` should only define a mechanism to take a generic list of properties (or perhaps two lists, one for parameters pertaining to the `volume type` and one pertaining to `cinder.conf`.

comay: And if it's really the case that every backend does have a secret key, then the parameter should be more generically defined such as `secret` or `secret_key`.

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -372,6 +389,41 @@
   end
 end
 
+if zone_config.alternate_backends_enabled?
+  alternate_backends.each do |backend|
+    backend_name = backend['backend_name']
+    backend_properties = backend['volume_type_properties']
+    create_args = [].append(backend_name)
```
comay, line 396: For consistency sake, could you change this to the two lines as seen in #362-363 - or alternatively, change that one to the single line implementation?

chef/cookbooks/bcpc/libraries/zone_config.rb
```diff
@@ -91,6 +99,41 @@ def backends
...
+      zones = @zone_config.zone_attr()
+      databag = @zone_config.databag(id: 'zones')
+      general_backend_configs = Hash[
+        @zone_config.alternate_backends()
+                    .map { |backend| [backend[:name].to_s, backend] }]
+      zones.each do |zone|
+        next if zone['cinder']['alternate_backends'].nil?
+        backends = zone['cinder']['alternate_backends']
+        backend_databags = Hash[
+        databag[zone['zone']]['alternate-backends']
```
comay, line 115: Shouldn't the `alternate-backends` here be `alternate_backends`?

chef/cookbooks/bcpc/libraries/zone_config.rb
```diff
@@ -91,6 +99,41 @@ def backends
...
 
+  def alternate_backends
+    alternate_backends = []
+
+    if @zone_config.enabled? && @zone_config.alternate_backends_enabled?
+      zones = @zone_config.zone_attr()
+      databag = @zone_config.databag(id: 'zones')
+      general_backend_configs = Hash[
+        @zone_config.alternate_backends()
+                    .map { |backend| [backend[:name].to_s, backend] }]
```
comay, line 110: Nit - I believe the closing `]` should be on its own line, aligned with the opening `general_backend_configs = Hash[`.

chef/cookbooks/bcpc/libraries/zone_config.rb
```diff
@@ -91,6 +99,41 @@ def backends
...
+      zones.each do |zone|
+        next if zone['cinder']['alternate_backends'].nil?
+        backends = zone['cinder']['alternate_backends']
+        backend_databags = Hash[
```
comay, line 114: Same comment as above regarding reliability.

chef/cookbooks/bcpc/libraries/zone_config.rb
```diff
@@ -91,6 +99,41 @@ def backends
...
+        databag[zone['zone']]['alternate-backends']
+                           .map { |b| [b[:name].to_s, b] }]
+        backends.each do |backend|
+          next if general_backend_configs[backend['name']]['enabled']
```
comay, line 118: Shouldn't this be

```
        next if general_backend_configs[backend['name']]['enabled']
```

?

chef/cookbooks/bcpc/libraries/zone_config.rb
```diff
@@ -91,6 +99,41 @@ def backends
     backends
   end
 
+  def alternate_backends
+    alternate_backends = []
+
+    if @zone_config.enabled? && @zone_config.alternate_backends_enabled?
```
comay, line 105: You can lose a level of indentation by changing this to

```
    if !@zone_config.enabled? || !@zone_config.alternate_backends_enabled?
      alternate_backends
    end
```

chef/cookbooks/bcpc/recipes/cinder.rb
npustchi: changed both to one line.

chef/cookbooks/bcpc/libraries/zone_config.rb
npustchi: no, line 115 follows databag convention, when line 113 follows environment.

chef/cookbooks/bcpc/libraries/zone_config.rb
npustchi: it should be unless, 'instead' of 'if'.

chef/cookbooks/bcpc/libraries/zone_config.rb
npustchi: I prefer to keep this pattern, I think is more readable when someone reads the code, for the first time. I try to keep the structure simple.

npustchi:
> Getting closer...
> 
> If we are going to be generic about the additional backend to support, I think we kind of have to truly be generic -- re: not mandating a JWT for additional backends.

this has been changed.

chef/cookbooks/bcpc/libraries/zone_config.rb
comay: It's a bit confusing though and there's nothing preventing using `alternate_backends`, correct? I mean we already have `chef_databags` and some of the properties one might define will contain a `_` as well.

npustchi: chef_databags keys are all separated with "-" not with "_". specifically zone data structure uses the same pattern I used. We have some password_* which I also used the same pattern here.

comay (approved):

comay (approved):

PR #2197, opened 2023-08-15
Allow Jammy-based worknodes to converge and work

tj90241:
We've laid all the groundwork to allow Jammy-based
hypervisors to converge and operate successfully - just
need some additional taps of the wrench within this
PR to bring things full circle.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/nova-compute.rb
+++ b/chef/cookbooks/bcpc/recipes/nova-compute.rb
@@ -39,6 +39,7 @@ package %w(
   nova-api-metadata
   ovmf
   pm-utils
+  qemu-block-extra
   sysfsutils
 ) do
   options '--no-install-recommends'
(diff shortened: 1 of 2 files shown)


chef/cookbooks/bcpc/recipes/nova-compute.rb
```diff
@@ -39,6 +39,7 @@
   nova-api-metadata
   ovmf
   pm-utils
+  qemu-block-extra
```
comay, line 42: Can you provide a bit more details on the need for this package?

tj90241: The `qemu-block-extra` package provides shared libraries for block storage backends, such as RBD (`usr/lib/x86_64-linux-gnu/qemu/block-rbd.so`).
We formerly pulled in `qemu-block-extra` for free as a hard requirement of another package (`nova-compute`?); in Jammy, this hard dependency was broken and the package now has to be installed explicitly if one requires it.

comay: Thanks for the explanation.

comay (approved):

PR #2202, opened 2023-09-06
RDTIBCC-4741: Add workaround for Lunar kernels

tj90241:
Calico (Felix) 3.26 BPF objects only work with `bpftool`
linked against libbpf < 1.0. This is because the BPF ELF
files contain sections that libbpf >= 1.0 dropped support
for. As a result, Felix cannot load the eBPF that it
requires in order to operate.

Unfortunately, Canonical statically links libbpf into
`bpftool`, so there is no tomfoolery with dynamic linkers
to be had.

Thus, until Tigera addresses the deprecated/legacy map
sections in their BPF file, just continue to use `bpftool`
from the "mainline" Jammy kernels.


tj90241:
I did open a Tigera case for this, by the way. Once the BPF objects packaged with `calico-helix`  stop making use of deprecated sections, we can just drop this block of code.

For now, I still have not enabled Lunar kernels in BVE/Thermite builds, so this is effectively a no-op until more testing is performed to ensure that there are no other regressions.

comay (approved):

PR #2208, opened 2023-09-27
Remove all references to Bionic

tj90241:
As of chef-bcpc 8.4.0, support for Bionic is formally
deprecated and unsupported. Remove all configuration
specific to it.

### DIFF
--- a/ansible/playbooks/roles/cinder/meta/main.yml
+++ b/ansible/playbooks/roles/cinder/meta/main.yml
@@ -9,7 +9,7 @@ galaxy_info:
   platforms:
   - name: Ubuntu
     versions:
-    - bionic
+    - focal
 
 dependencies:
   - role: haproxy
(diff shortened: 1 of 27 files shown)


psipika (changes requested):
Shellcheck is failing with a minor thingy:

```
In ./virtual/network/provisioner.sh line 25:
distribution_codename=$(lsb_release -sc)
^-------------------^ SC2034 (warning): distribution_codename appears unused. Verify use (or export if used externally).
```

tj90241:
Note that in `ansible/playbooks/roles/common/defaults/main/base.yml` it still refers to Ubuntu 18.04. It seems the version of `cine-client` we're using is so old that 20.04 was not a thing when it was compiled and hence there is no 20.04 package. Though I suppose it's fitting, as our version of `chef-server` is from 16.04... sigh.

psipika (approved):
🚀

PR #2211, opened 2023-09-29
RDTIBCC-4728: Fix keystone healthcheck failures

tj90241:
heat.conf declares the heat_domain_admin to be in "Default"
domain, but the user is created by Chef in the "heat"
domain in the Heat recipe.

Change the configuration to reflect this, which ultimately
prevents Heat from trying to lookup a user that does not
exist (thus tripping alarms).


psipika (approved):
🥵

tj90241:
See: https://github.com/bloomberg/chef-bcpc/blob/eee73d2229dd4e2c103a142a2a9aa0279e4d1e9c/chef/cookbooks/bcpc/recipes/heat.rb#L94

PR #2215, opened 2023-10-24
add decoding databag items.

npustchi:
RDTIBCC-4850

This change will enabled decoding of data bag items encoded in base64.

Tested with a build.

### DIFF
--- a/chef/cookbooks/bcpc/libraries/zone_config.rb
+++ b/chef/cookbooks/bcpc/libraries/zone_config.rb
@@ -15,6 +15,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require 'base64'
+
 class ZoneConfig
   def initialize(node, region, data_bag_item)
     @cinder_config = CinderConfig.new(self)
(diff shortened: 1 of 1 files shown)


chef/cookbooks/bcpc/libraries/zone_config.rb
```diff
@@ -124,7 +126,8 @@ def alternate_backends
           'volume_driver' => general_backend_configs[backend['name']]['volume_driver'],
           'properties' => {}.merge(
           general_backend_configs[backend['name']]['properties'])
-              .merge(backend_databags[backend['name']]['properties'])
+              .merge(Hash[backend_databags[backend['name']]['properties']
```
comay, line 129: So does this imply that all properties in an alternate backend need to be Base64 encoded?

Also, curious why you're doing the decoding here rather than in `cinder.rb` since the other Base64 decoding occurs in the recipes themselves?

npustchi: all properties in chef databags.
We keep properties generic in alternate backed, so all properties are populated into a map of properties. At that point all properties, chef databags, environment and overrides are merged into a map of properties. We can not distinguish those without losing alternate backend being generic.

comay (approved):

PR #2217, opened 2023-10-30
integrated Haproxy configuration into haproxy role

chaowang987:
**[RDTIBCC-4779](https://jira.prod.bloomberg.com/browse/RDTIBCC-4779) - HAProxy role integration in ansible**

**Describe your changes**
Individual roles for configuring haproxy clients have been moved to haproxy role. Now reload haproxy should be only ran once after finishing iteration of all of the haproxy configuration.

**Testing performed**
+ ran the individual playbook on thermite build and reload proxy was only run once after placing configuration for each service
```
ansible-playbook -v \
                -i inventory.yml playbooks/site.yml \
                -t configure-haproxy-clients -f 1 --limit headnodes
```
+ ran make all on thermite build and the automation completed without issue.


Makefile
```diff
@@ -122,63 +122,11 @@ configure-haproxy :
...
-	configure-placement-haproxy \
-	configure-nova-haproxy \
-	configure-cinder-haproxy \
-	configure-heat-haproxy \
-	configure-watcher-haproxy
-
-configure-keystone-haproxy :
+configure-haproxy-clients :
 
 	ansible-playbook -v \
```
npustchi, line 127: since we integrated the haproxy role, using configure-haproxy tag for clients would be fine. You can remove this.

ansible/playbooks/roles/haproxy/tasks/main.yml
```diff
@@ -10,3 +10,7 @@
 - import_tasks: configure-haproxy.yml
   become: true
   tags: [never,configure-haproxy]
+
+- import_tasks: configure-haproxy-clients.yml
+  become: true
+  tags: [never,configure-haproxy-clients]
```
npustchi, line 16: you can replace this tag with configure-haproxy

npustchi (dismissed):
LGTM, just please address the comments.

ansible/playbooks/roles/haproxy/tasks/main.yml
chaowang987: Tyler mentioned for example, you want to rebuild individual service like nova, you might want to have a tag to do that specifically so you dont need to run the whole configure-haproxy. if you dont think that is necessary, I'll just remove it.

tj90241: Yeah, for context... I think that Navid originally wrote it the way that this PR is headed in, but maybe it was @comay or me that requested it be broken out Into individual roles? Anyways, since @npustchi wrote the code I thought he should at least have some say here.

npustchi: I think having a single tag is fine. As long as there is no change in haproxy, and just a nova config change. It will only reload haproxy and it will not affect it. And if there is, it is safer to have a single tag.

npustchi (approved):

PR #2221, opened 2023-11-06
Enable rbd_thin_provisioning for Glance

zhanz1:
# WHY
When uploading images to Glance, the image files may be sparse files, meaning that there exist a large amount of "holes" (null bytes) in the file. For [example](https://blueprints.launchpad.net/glance-store/+spec/handle-sparse-image), the 8GB base CentOS 7 cloud image consists of 1GB of data and 7GB of holes (null bytes). This means that glance_store, without enabling `rbd_thin_provisioning`, is writing all of the data along with the null bytes, which implies that null bytes (holes) are sent over the network and consuming spaces in the storage backend. 

Currently, RBD already supports [sparse files](https://www.oreilly.com/library/view/unix-filesystems-evolution/9780471456759/chap03-sec017.html#:~:text=A%20hole%20is%20a%20gap,hole%2C%20zeroes%20will%20be%20returned.), meaning that if glance only write the [actual data themselves with offsets](https://review.opendev.org/c/openstack/glance_store/+/744282) (without the null bytes), Ceph is able to interpret that and output the null byte sequence when accessed by others (less to upload, but no changes to users who access the image). 

The `rbd_thin_provisioning` therefore takes an advantage of this feature by [not writing the chunks](https://opendev.org/openstack/glance_store/commit/201d85b4eab9ffa4aa0aacf6c21e03a771173da4#diff-9f65ef44d483dd1a0ec1fafdef9cf9fba466e756) (in image file) that are null bytes (holes). In this case, using the 8GB base CentOS 7 cloud image example again, glance_store, with `rbd_thin_provisioning` enabled, will not send that 7GB of null bytes, but only the 1GB data with proper offsets to storage backend. This greatly reduce the upload time as well as the storage consumption. 

So far there is no drawbacks expected, as long as the storage backend supports sparse file. This can be verified by using the uploaded image (with `rbd_thin_provisioning` enabled) to create an instance and check if it is working. Please check the `TEST` section. 

Regarding more details and more test results, please check out [here](https://specs.openstack.org/openstack/glance-specs/specs/victoria/approved/glance_store/handle-sparse-image.html). 

# WHAT
Enable `rbd_thin_provisioning` for Openstack Glance service. 

# Potential Negative Effects
The key difference here is that with `rbd_thin_provisioning` enabled, `glance_store` will sending data (files) in a different format (sparse file). If the storage backend does not support this format (read and write), then file would be corrupted and cannot be used. However, since the current storage backend supports processing sparse file, there should be no concerns regarding this issue. 

# TEST
## Validation
After building the cloud using `make all`, the config is enabled in `glance-api.conf`:
```
...
[glance_store]
...
rbd_thin_provisioning = true
...
```
Then create the following image for testing(using `cirros` and `Ubuntu 22.04`):
```
$ openstack image create cirros_test_thin <additional_args_for_creating_img>
<Img info>
$ openstack image create Ubuntu_test_thin <additional_args_for_creating_img>
<Img info>
```
And create a server to test if the image is usable:
```
$ openstack server create test-instance --image cirros_test_thin --wait <additional_args_for_creating_server>
<VM Info with {IP}>
$ ping {IP}
PING {IP} ({IP}) 56(84) bytes of data.
...
--- {IP} ping statistics ---
5 packets transmitted, 5 received, 0% packet loss, time 4006ms

$ openstack server create test-instance --image Ubuntu_test_thin --wait <additional_args_for_creating_server>
<VM Info with {IP}>
$ ping {IP}
PING {IP} ({IP}) 56(84) bytes of data.
...
--- {IP} ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2011ms
```
Download the image again using glance, and there is no difference between the uploaded img and downloaded img
```
$ glance image-download --file ./Ubuntu_thin_download.img <img_id> --progress
[=============================>] 100%
$ ls
Ubuntu_test.img  Ubuntu_thin_download.img 
$ diff Ubuntu_test.img Ubuntu_thin_download.img 
$ echo $?
0
```
## Performance Improvements
Using Glance's log to check the approximate runtime for uploading image (from the steps of uploading image using rbd to done checking image is uploaded correctly). Four runs of `openstack image create` are performed with thin provisioning enabled and disabled:
```
Enabled:
17:46:47.437 - 17:46:58.948 = 11.511s
17:47:09.933 - 17:47:14.697 = 4.641s
17:47:52.025 - 17:47:55.432 = 3.407s
17:47:30.629 - 17:47:36.448 = 5.819s

Disabled:
17:35:37.376 - 17:35:51.028 = 13.652s
17:36:19.128 - 17:36:33.054 = 13.926s
17:37:09.964 - 17:37:27.125 = 17.161s
17:39:26.175 - 17:39:43.200 = 17.025s
```
Similar test has been conducted with Ubuntu:
```
Enabled:
15:00:36.222 - 15:09:18.265 = 522.043s
15:42:47.980 - 15:51:26.420 = 518.44s
16:10:54.596 - 16:19:46.066 = 531.47s

Disabled:
16:21:55.811 - 16:40:17.718 = 1101.907s
17:45:15.110 - 18:02:51.909 = 1056.799s
18:33:28.601 - 18:52:42.905 = 1154.304s
```
The improvement in performance is similar to the results showing [here](https://specs.openstack.org/openstack/glance-specs/specs/victoria/approved/glance_store/handle-sparse-image.html).


jingqgeng:
Wow, there's some huge difference after the first upload! I wonder why that's the case...did you use the same image repeatedly? Can you maybe try images of larger sizes and see what happens?

tj90241:
@jingqgeng My guess is virtualization and "cold start" overhead may have come into play? Would be curious to see if there is any difference with this feature disabled vs. first and second runs, assuming the "disabled" runs here were not run on a totally clean build.

> We know from experience that seemingly simple tweaks can have unintended consequences so should cover as many bases as possible.
Indeed... 😄

zhanz1:
> Wow, there's some huge difference after the first upload! I wonder why that's the case...did you use the same image repeatedly? Can you maybe try images of larger sizes and see what happens?

Updated the comment with tests using Ubuntu. The same image is used repeatedly and it is deleted before each upload.

> We need a bit more detail here, specifically:
> 
> * What exactly does rbd_thin_provisioning do? Why should we expect image uploads to be faster? What other operations should we expect to be faster? Slower?
> * What are the potential drawbacks for enabling this feature?
> 
> We know from experience that seemingly simple tweaks can have unintended consequences so should cover as many bases as possible.

I've linked relevant pages that are related to this change and hopefully they should be able to clarify some confusions. 

>  Would be curious to see if there is any difference with this feature disabled vs. first and second runs, assuming the "disabled" runs here were not run on a totally clean build.

We've spoke about this and I will test on it :).

mboruta1 (approved):
Looks good to me. Would note that it looks like we will benefit from the write optimization detailed in [spec](https://specs.openstack.org/openstack/glance-specs/specs/victoria/approved/glance_store/handle-sparse-image.html), but not from the read due to lack of a staging store.

tj90241 (approved):

PR #2231, opened 2023-11-15
Insert hack to fix suspected race condition

tj90241:
During the HAproxy step. Unfortunately, it looks like
there is now another issue impacting 3h3w builds that
arises during the `bcpc::glance` recipe.

Add a large sleep in before Chef'ing to:
  a) Confirm that there actually is some kind of race-
 condition residing about in the code and

  b) Fix the very bad success rate of builds that is now
 consuming developer time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ all : \
 	configure-web-server \
 	configure-common-node \
 	configure-haproxy \
+	apply-duct-tape \
 	run-chef-client \
 	configure-ceph \
 	add-cloud-images \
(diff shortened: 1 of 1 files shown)


mihalis68 (approved):

chaowang987 (approved):

comay:
Just curious, did this race start after #2217 was merged?

The work-around is fine although I'd prefer it was localized to the end of the `configure-haproxy` target with a comment indicating it's to work-around a race condition with `bcpc::glance`.

PR #2236, opened 2023-12-13
Bump apparmor

mihalis68:
apparmor/libapparmor files are gone from kernel.org. Bump to next nearest minor version. Files now download ok. No further testing just yet.


mihalis68:
wrong base

PR #2240, opened 2023-12-18
Set internal project & user id for image-volume cache

jingqgeng:
I'd like to enable Cinder [image-volume cache](https://docs.openstack.org/cinder/latest/admin/image-volume-cache.html) for alternate backends, and it is required to set `cinder_internal_tenant_project_id` and `cinder_internal_tenant_user_id` in the default section of cinder configurations.

**Testing performed**
Tested on a physical test cluster that has alternate Cinder backends enabled.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/cinder.rb
+++ b/chef/cookbooks/bcpc/recipes/cinder.rb
@@ -270,6 +270,7 @@ template '/etc/cinder/cinder.conf' do
     rmqnodes: rmqnodes(all: true),
     alternate_backends: cinder_config.alternate_backends,
     scheduler_default_filters: cinder_config.filters,
+    init_cloud: init_cloud?,
     cinder_internal_tenant_project_id: lazy { cinder_internal_tenant_project_id },
     cinder_internal_tenant_user_id: lazy { cinder_internal_tenant_user_id }
   )
--- a/chef/cookbooks/bcpc/recipes/cinder.rb
+++ b/chef/cookbooks/bcpc/recipes/cinder.rb
@@ -210,6 +210,64 @@ if zone_config.enabled?
   end
 end
 
+if zone_config.alternate_backends_enabled?
+  internal_project_id = ''
+  internal_user_id = ''
+  # configure Cinder internal project and user id
+  internal_project = node['bcpc']['cinder']['alternate_backends']['cinder_internal_tenant_project'] || ''
+#  unless internal_project.to_s.strip.empty?
+  ruby_block 'collect Cinder internal project uuid' do
+    block do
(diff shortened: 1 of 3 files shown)


mboruta1 (dismissed):
Nice!

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -210,6 +210,44 @@
...
+      os_command_out = shell_out(os_command, env: os_adminrc)
+      if os_command_out.error?
+        raise "Unable to find the project id for project name #{internal_project}"
+      end
+      internal_project_id = os_command_out.stdout.chomp()
+      unless internal_project_id.empty?
+        cinder_internal_tenant_project_id = internal_project_id
+      end
+    end
+    not_if internal_project.empty?
```
tj90241, line 229: Do you need to guard this with `and init_cloud?` or something similar?

What if the `internal_project` is not `nil` (i.e., the user wants this feature), and they are building a new cluster? The projects other than `admin` are not created until after `bcpc::cinder` runs.

tj90241: Alternatively, maybe it's easier to only allow this to be configured for the `admin` project instead of trying to work out how to do that, if that's how you intend to use this feature?

jingqgeng: I do intend to only use `admin`, even though it is not a requirement. Since accessing & managing the image-volume-cache is a kind of a super-power so it should be only accessible to `admin`. Maybe I should be more explicit and check the value set for `internal_project` here?

tj90241: Either sounds good - just trying to avoid a future issue if we decide to use a different tenancy.

jingqgeng: That was a good point!

chef/cookbooks/bcpc/templates/default/cinder/cinder.conf.erb
```diff
@@ -22,6 +22,10 @@ enable_v3_api = true
 <% if @scheduler_default_filters.any? %>
 scheduler_default_filters = <%= @scheduler_default_filters.join(',') %>
 <% end %>
+<% if @alternate_backends_enabled%>
```
tj90241, line 25: You also need the check here, right? Otherwise, the first Chef run will define `cinder_internal_tenant_project_id` as empty when `init_cloud?` is true, but `@alternate_backends_enabled` will be true too?

tj90241 (approved):
🚀

comay (approved):

PR #2244, opened 2023-12-29
Upgrade to Calico 3.27.0 - RDTIBCC-4947

comay:
I've tested the upgrade under both a `3h3w` build as well as through a Jenkins build including a full Rally run. All builds were successful and the tests passed.

### DIFF
--- a/ansible/playbooks/roles/common/defaults/main/base.yml
+++ b/ansible/playbooks/roles/common/defaults/main/base.yml
@@ -184,8 +184,8 @@ assets_files:
     environment: "{{ internet_proxy | default({}) }}"
 
   - name: calicoctl
-    url: https://github.com/projectcalico/calico/releases/download/v3.26.1/calicoctl-linux-amd64
-    checksum: sha256:c8f61c1c8e2504410adaff4a7255c65785fe7805eebfd63340ccd3c472aa42cf
+    url: https://github.com/projectcalico/calico/releases/download/v3.27.0/calicoctl-linux-amd64
+    checksum: sha256:46e79ae146b3dd90998f56511cf5d6db64deb97cb784235caf1f99e0672d66e4
     filename: calicoctl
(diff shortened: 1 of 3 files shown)


tj90241:
What testing was performed?

tj90241:
This hack should now be removed:
https://github.com/bloomberg/chef-bcpc/commit/c65a83040ed104a754c8351e534de1095d45e289#diff-608b0148f775df0cd6f1a88372206290443a13677cb8c5db5a981fbdea256a0eR189

The fixes for the BPF objects are contained within 3.27.0 per our request.

comay:
> This hack should now be removed: [c65a830#diff-608b0148f775df0cd6f1a88372206290443a13677cb8c5db5a981fbdea256a0eR189](https://github.com/bloomberg/chef-bcpc/commit/c65a83040ed104a754c8351e534de1095d45e289#diff-608b0148f775df0cd6f1a88372206290443a13677cb8c5db5a981fbdea256a0eR189)
> 
> The fixes for the BPF objects are contained within 3.27.0 per our request.

Fixed.

comay:
> What (functional) testing was performed? The virtual clusters do not have policy pushed to them...?

As I typically do, I created a number of instances and verified that the policy set by the default security groups is enacted by Calico. I will definitely test this on a semi-production system prior to merging, once the packages are available internally (PR already approved but not yet merged or deployed for that). In any case, I wasn't going to merge this until it's available internally.

Also, I think perhaps it makes sense to provide some sample policy that mirrors what the security groups provides (such as ICMP and/or SSH) that can be used to model Calico in production better but I think that's the subject of a separate PR.

tj90241 (approved):
As long as representative policy set is tested, ok -- that's my primary concern here... given the nature of Calico 3.27 was to optimize many of the former algorithms related to Felix policy, there could be a regression in the policy rendering that would not be seen in just a vanilla build (as you mention).

PR #2245, opened 2023-12-29
Nova databases should be built before services come online - RDTIBCC-4950

comay:
Recently, I've seen build failures where a failure occurs in bcpc::nova-head when it brings the Nova services online via

```
service 'nova-api'
service 'nova-scheduler'
service 'nova-conductor'
service 'nova-novncproxy' do
  service_name 'apache2'
```

The issue appears to be that `nova-scheduler` tries to open repeatedly the database back-end specified in `/etc/nova/nova.conf` file before the database/tables have been created. This race appears obvious looking at the code - we should not have Chef define these services until all of the relevant files have been laid down.

I've tested the upgrade under both a 3h3w build as well as through a Jenkins build including a full Rally run. All builds were successful and the tests passed.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/nova-head.rb
+++ b/chef/cookbooks/bcpc/recipes/nova-head.rb
@@ -135,13 +135,6 @@ package %w(
   options '--no-install-recommends'
 end
 
-service 'nova-api'
-service 'nova-scheduler'
-service 'nova-conductor'
-service 'nova-novncproxy' do
-  service_name 'apache2'
-end
-
 # create policy.d dir for policy overrides
 directory '/etc/nova/policy.d' do
   action :create
@@ -337,10 +330,10 @@ template '/etc/nova/nova.conf' do
   )
 
   notifies :run, 'execute[update cell1]', :immediately
-  notifies :restart, 'service[nova-api]', :immediately
-  notifies :restart, 'service[nova-scheduler]', :immediately
-  notifies :restart, 'service[nova-conductor]', :immediately
-  notifies :restart, 'service[nova-novncproxy]', :immediately
+  notifies :restart, 'service[nova-api]', :delayed
+  notifies :restart, 'service[nova-scheduler]', :delayed
+  notifies :restart, 'service[nova-conductor]', :delayed
+  notifies :restart, 'service[nova-novncproxy]', :delayed
 end
 
 execute 'update cell1' do
@@ -414,6 +407,11 @@ if extended_apis['enabled']
     end
   end
 end
+
+service 'nova-api'
+service 'nova-conductor'
+service 'nova-novncproxy'
+service 'nova-scheduler'
 
 execute 'wait for nova to come online' do
   environment os_adminrc


tj90241:
I'm confused... I don't doubt that this PR may fix things, but is it doing what it purports to?

I struggle to understand what was even going on before 🙃 :
```
service 'nova-api'
service 'nova-scheduler'
service 'nova-conductor'
service 'nova-novncproxy' do
  service_name 'apache2'
end
```

This somehow ties stuff to `apache2`??? Anyways...

The Nova services are still going to be started, [here](https://github.com/bloomberg/chef-bcpc/pull/2245/files#diff-a5fb30382887b13e45a6c0980045c07c216d7255e8f3bb0c9cc13229c54baea9L128) before the database is built after this PR is merged as a result of the `postinst` scripts contained within e.g. `nova-api`. The only way to stop them from starting is to install `nova-common` and setup the database with that, as `nova-common` does not have any `postinst` scripts.

Maybe the changes you have made here just result in us kicking Nova in the shins just a little less while we're bootstrapping -- enough to tide things over?

Regardless, another aside: this _seems_ like it _could_ potentially present trouble wrt the delayed restarts given how we patch a bunch of Nova's code to since adding `extended_apis` -- specifically, depending on the configuration, this PR will result in Nova running for an appreciable amount of time with the `extended-apis` config enabled in `api-paste.ini` and some files like `policy.py` patched with changes for the new code, but the service unbounced -- it's kind of in a half-deployed state?

Was your testing with the `extended_api` turned on? It's default off in `chef-bcpc`.

I think the correct way to probably address this issue is something similar to [the treatment formerly given to Cinder](https://github.com/bloomberg/chef-bcpc/commit/2b7261dd674508e8e24e60b03cc742e00146d39a) -- it is a little more involved than what's going on here.

jingqgeng:
Regarding `extended_apis`, if we apply this change, will the extended apis become temporarily unavailable while this recipe runs?

comay:
> I'm confused... I don't doubt that this PR may fix things, but is it doing what it purports to?

It aims to fix the race that I observed where the nascent `nova-scheduler` `systemd` service fails, which causes Chef to exit prematurely.

> I struggle to understand what was even going on before 🙃 :

I hear you - it's a bit of a mess.

> 
> ```
> service 'nova-api'
> service 'nova-scheduler'
> service 'nova-conductor'
> service 'nova-novncproxy' do
>   service_name 'apache2'
> end
> ```
> 
> This somehow ties stuff to `apache2`??? Anyways...

So there's a couple of issues with this block, besides its placement in the file. None of these services are Apache-based. What I think happened is as part of the Train upgrade (`0d1cb4230e7f54b7517f819aab1934fba2d49095`), the following change was made

```
@@ -223,11 +148,9 @@ package %w(
 end

 service 'nova-api'
-service 'nova-consoleauth'
 service 'nova-scheduler'
 service 'nova-conductor'
-service 'nova-novncproxy'
-service 'placement-api' do
+service 'nova-novncproxy' do
   service_name 'apache2'
 end
 service 'haproxy-nova' do
 ```
 
and so the `service_name apache2` (which was previously *only* applied to `placement-api` but appears in the file to be universal because of the way the block is structured) ends up now being associated incorrectly with `nova-novncproxy`. So as part of fixing the race, the moved block removes this bit of debris.
 
> The Nova services are still going to be started, [here](https://github.com/bloomberg/chef-bcpc/pull/2245/files#diff-a5fb30382887b13e45a6c0980045c07c216d7255e8f3bb0c9cc13229c54baea9L128) before the database is built after this PR is merged as a result of the `postinst` scripts contained within e.g. `nova-api`. The only way to stop them from starting is to install `nova-common` and setup the database with that, as `nova-common` does not have any `postinst` scripts.

Indeed and while the services will complain and perhaps fail from a `systemd` perspective, what actually appears to cause Chef to quit is the fact that Chef's `service` resource looks at the `systemd` status and if it's failed, it ends the Chef run.

Because it's a race, I believe that in "working" builds the `systemd` status of these services (but in particular, `nova-scheduler`) is still OK but in "failed" builds, the `systemd` status has already turned to failed and so the Chef `service` resource fails as well.
 
> Maybe the changes you have made here just result in us kicking Nova in the shins just a little less while we're bootstrapping -- enough to tide things over?

Yes, that's part of it - basically, it works around what I consider to be broken Debian/Ubuntu packages where a service that cannot reasonably be enabled/started out of the box is delivered as such.

> Regardless, another aside: this _seems_ like it _could_ potentially present trouble wrt the delayed restarts given how we patch a bunch of Nova's code to since adding `extended_apis` -- specifically, depending on the configuration, this PR will result in Nova running for an appreciable amount of time with the `extended-apis` config enabled in `api-paste.ini` and some files like `policy.py` patched with changes for the new code, but the service unbounced -- it's kind of in a half-deployed state?

I suspect there's already a race here If the Python interpreter ends up loading a new or updated Python file but we haven't yet bounced the service. The one thing that perhaps can minimize that (different) race is we collect all of the "code patching" at the end and if any one of those triggers a file update, then we immediately restart (all?) of the Nova services.
 
> Was your testing with the `extended_api` turned on? It's default off in `chef-bcpc`.

Yes, I had enabled it.
 
> I think the correct way to probably address this issue is something similar to [the treatment formerly given to Cinder](https://github.com/bloomberg/chef-bcpc/commit/2b7261dd674508e8e24e60b03cc742e00146d39a) -- it is a little more involved than what's going on here.

Understood and if you feel this shouldn't go in before that more correct fix, that's OK. Apparently I'm the only one running into this particular race - it happens all the time for me and I'm not sure why now (I noticed it while moving my Vagrant-based lab instance over to a Cascade Lake system as part of testing the CPU baseline work).

tj90241:
> > I'm confused... I don't doubt that this PR may fix things, but is it doing what it purports to?
> 
> It aims to fix the race that I observed where the nascent `nova-scheduler` `systemd` service fails, which causes Chef to exit prematurely.

...

> Understood and if you feel this shouldn't go in before that more correct fix, that's OK. Apparently I'm the only one running into this particular race - it happens all the time for me and I'm not sure why now (I noticed it while moving my Vagrant-based lab instance over to a Cascade Lake system as part of testing the CPU baseline work).

I'm all for fixing it, but what I am trying to point out is that the race you are fixing can _still_ happen as the `package` stanzas trigger `postinst` scripts that run (not to mention, the issues with the extended API possibly being patched all in the midst of this). Mostly, I'm afraid that while it may fix a build of type XYZ, it could just as easily break something else.

I've made some additional suggestions to this PR that should avoid it outright, I think -- not really too much different than what was proposed here, which should prevent any service start until everything is laid down on the disk and the DBs, etc. are all setup:
https://github.com/bloomberg/chef-bcpc/commit/53cb8edddba1f30dbc92e202f2caeb9108b06ba2

comay:
> Regarding `extended_apis`, if we apply this change, will the extended apis become temporarily unavailable while this recipe runs?

I don't believe it will but in any case, I've already started restructuring things along the lines of what Tyler did for Cinder.

comay:
> I'm all for fixing it, but what I am trying to point out is that the race you are fixing can _still_ happen as the `package` stanzas trigger `postinst` scripts that run and start the Nova services already (not to mention, the issues with the extended API possibly being patched all in the midst of this). Mostly, I'm afraid that while it may fix a build of type XYZ, it could just as easily break something else.
> 
> I've made some additional suggestions to this PR that should avoid it outright, I think -- not really too much different than what was proposed here, which should prevent any service start until everything is laid down on the disk and the DBs, etc. are all setup: [7b66889](https://github.com/bloomberg/chef-bcpc/commit/7b66889ba6ca0327bf92a60fabd206d4be2c7235)

Thanks! I already started down this path yesterday and so I'll compare/check the changes I've made with the above commit and then refresh this PR accordingly.

tj90241:
We also needed to do this for Neutron -- it was being held together with a `systemd` hack I added: https://github.com/bloomberg/chef-bcpc/commit/096ab09d275a7cfdc0365eb2a1b987c0c694b4e7

comay:
Closing in lieu of #2248.

PR #2246, opened 2024-01-11
provide stages within make all for easier retrying

mihalis68:
This is no change to behavior when you issue make all. However it adds optional intermediate parts of all such as prechef runchef and postchef so that retrying is a little easier. If the stages before Chefing nodes completed, one can retry the chef stage with make runchef and then make postchef.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,11 @@ storageheadnodes = \
 stubnodes = $$(ansible stubnodes -i ${inventory} --list | tail -n +2 | wc -l)
 
 all : \
+	prechef \
+	runchef \
+	postchef
+
+prechef : \
 	sync-assets \
 	configure-operator \
 	configure-node \
@@ -23,8 +28,12 @@ all : \
 	configure-chef-nodes \
 	configure-web-server \
 	configure-common-node \
-	configure-haproxy \
-	run-chef-client \
+	configure-haproxy
+
+runchef : \
+	run-chef-client
+
+postchef : \
 	configure-ceph \
 	add-cloud-images \
 	register-compute-nodes \


comay:
@mihalis68 I'll take a look at this - I like the idea although I'm wondering about precise boundaries so will think back to my failed builds to see if they correspond.

tj90241:
Another idea -  leverage Make's strengths and stamp files to determine the dependencies to run:

```
all: idk-maybe-run-chef
.PHONY: clean idk-maybe-run-chef run-ansible-part-1 run-ansible-part-2

clean:
	@rm -rf .stamps

.stamps:
	mkdir -p .stamps

run-ansible-part-1 .stamps/ansible_part1: .stamps
	@echo now running ansible part 1
	@touch .stamps/ansible_part1

run-ansible-part-2 .stamps/ansible_part2: .stamps/ansible_part1
	@echo now running ansible part 2
	@touch .stamps/ansible_part2

# simulate a chef-client failure
idk-maybe-run-chef: .stamps/ansible_part2
	@echo now running chef
	false
```

e.g.:
```
tstachecki@JH3JX9X4JK play % make clean all
mkdir -p .stamps
now running ansible part 1
now running ansible part 2
now running chef
false
make: *** [idk-maybe-run-chef] Error 1
tstachecki@JH3JX9X4JK play % make all
now running chef
false
make: *** [idk-maybe-run-chef] Error 1
```

One thing I'm not a fan of is duplicating the stamp filename in the corresponding `touch`, but "$@" expands to the target that caused the run to run, which in the case of `make run-ansible-part-1 ` would create a file named `run-ansible-part-1 ` in the directory when touching $@.

mihalis68:
this throws away one thing I wanted which was easy to type shorter make commands. Are you really going to make me type make run-chef-client?
make all is short
make runchef also
make three-delimited-words less so

comay:
Ah, I didn't realize you were specifically looking at shortening what needed to be typed as well - I thought it was strictly around grouping things. I tried to follow the existing nomenclature of `<verb>-<object>` but we could also put in some shorter aliases like `run-chef` for `run-chef-client` (I figure the latter was originally chosen since it was specific about what part of Chef was being run).

mihalis68:
Brevity and easy to remember are both goals here for me. I'd take run-chef

PR #2248, opened 2024-01-16
RDTIBCC-4950: Build Nova/Neutron databases before onlining services

tj90241:
To prevent spurious build failures, always build databases
before starting the associated services.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/neutron-head.rb
+++ b/chef/cookbooks/bcpc/recipes/neutron-head.rb
@@ -129,14 +129,12 @@ end
 #
 # create network service and endpoints ends
 
-# neutron package installation and service definition starts
+# neutron base installation starts
 #
-package 'neutron-server'
-
-package 'calico-control' do
-  action :upgrade
-  notifies :restart, 'service[neutron-server]', :delayed
-end
+package %w(
+  neutron-common
+  python3-neutron
+)
 
 # install patches for both neutron and neutron-lib
 # https://bugs.launchpad.net/neutron/+bug/1918145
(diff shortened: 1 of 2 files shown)


chef/cookbooks/bcpc/recipes/neutron-head.rb
```diff
@@ -306,7 +307,6 @@
   action :run
 end
 
-# create networks starts
```
comay, line 309: I believe line 471 or so should be removed as well then?

chef/cookbooks/bcpc/recipes/nova-head.rb
```diff
@@ -124,23 +124,11 @@
...
-# nova package installation and service definition
+# install nova base packages
 package %w(
-  ceph-common
-  nova-api
-  nova-conductor
-  nova-novncproxy
-  nova-scheduler
-) do
-  options '--no-install-recommends'
```
jingqgeng, line 135: Why is `optoins '--no-install-recommends'` removed in this commit?

comay: Good catch, Jing - it seems like it's missing from both `package` declarations.

I also wonder why we weren't do this with Neutron - I thought this had been added across the board.

tj90241: This has been the default for awhile now:
https://github.com/bloomberg/chef-bcpc/blob/f68c5ed204253eb795b0bdb746493ad3d6d1d4c1/virtual/packer/packer_box_provision.sh#L55

Though I am realizing the physical machines are not getting this configuration due to a regression elsewhere. I will include this in the Jammy builds -- in the meantime, I think we should proceed in removing it in cases like this.

chef/cookbooks/bcpc/recipes/neutron-head.rb
tj90241: Not following... this comment was reintroduced on another line? (L#287)

comay: Ah, sorry - I missed it being moved to line 287 so I thought you were deleting it.

comay (approved):

PR #2255, opened 2024-01-30
Clean up unneeded users, login shells

desrod:
This addresses the issues detailed in RDTIBCC1959 and RDTIBCC1966, related to ensuring that unnecessary user accounts are not included post-automation, and those that are, but do not require interactive logins, have their shells neutered to `nologin`.

### DIFF
--- a/ansible/playbooks/roles/common/tasks/configure-system.yml
+++ b/ansible/playbooks/roles/common/tasks/configure-system.yml
@@ -135,44 +135,39 @@
   user:
     name: "{{ item }}"
     state: absent
-    remove: no
-  loop: "{{ remove_users }}"
-  vars:
-    remove_users:
-      - 'backup'
-      - 'games'
-      - 'gnats'
-      - 'irc'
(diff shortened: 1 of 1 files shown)


ansible/playbooks/roles/common/tasks/configure-system.yml
```diff
@@ -130,6 +130,50 @@
     - udisks2
     - upower
 
+# Remove unnecessary user accounts (vs. nologin shells)
+- name: Remove specified user accounts
+  user:
+    name: "{{ item }}"
+    state: absent
+    remove: no
+  loop: "{{ remove_users }}"
```
tj90241, line 139:
```
  with_items:
    - 'backup'
    ...
```

or `loop:` and the list under it -- you don't need to create a variable for this.

tj90241:
lgtm, assuming you launched a VM or something afterwards to validate?

tj90241 (approved):

PR #2260, opened 2024-02-02
redo use of run_once to be passed in where appropriate

mihalis68:
enable-compute-service, configure-host-aggregates, configure-licenses talk solely to openstack via the CLI

register-compute-nodes talks solely to openstack via nova-manage

They only need to run once. Explicitly tell ansible this to avoid runs across subsequent head nodes.


comay (approved):
Please remember to use the `Squash and merge` option when merging and collapse the comments to just the the title and sign-off.

PR #2262, opened 2024-02-07
Update chef automation for creating cinder volume qos

mzhang741:
### Describe your changes
Update the automation to apply QoS for cinder volumes. Default qos limits are defined in the `chef_environment.yml` file, and we can add override values for each volume type in the `overrides.yml` file. To enable QoS for a storage backend, e.g. lightos, we just need to add `qos_enabled=true` tag under the cinder/lightos section in chef environment. Then a qos policy will be created and associated with each volume type for that storage backend. Once the `qos_enabled` flag is set to `false`, the corresponding qos policies will be disassociated and deleted. 

#### For Ceph-backed cinder volumes

1. Enable qos with default qos limits -> a qos with default values is created and associated for each volume type; default qos values are listed under each ceph volume type in cinder.conf on headnodes
2. Enable qos with override values -> qos created for each volume type, but with different qos values; qos override values are listed under the corresponding ceph volume type in cinder.conf on headnodes
3. Set qos_enabled=false, but override values are provided ->  all qos for ceph are deleted; qos values are not shown in cinder.conf file
4. Set qos_enabled=false when there are qos in place -> all qos for ceph are deleted; qos values are not shown in cinder.conf file
5. Set qos_enabled=false when there are no qos in place -> no action and cinder.conf remains the same
6. No qos_enabled tag when there isn't qos in place -> no action and cinder.conf remains the same
7. No qos_enabled tag when there is qos in place ->  all qos for ceph are deleted; qos values are not shown in cinder.conf file

#### For LB-backed cinder volumes

repeat the same tests for LB-backed volumes

1. Enable qos with default qos limits -> a qos with default values is created and associated for each volume type; default qos values are listed under each lightos volume type in cinder.conf on headnodes
2. Enable qos with override values -> qos created for each volume type, but with different qos values; qos override values are listed under the corresponding lightos volume type in cinder.conf on headnodes
3. Set qos_enabled=false, but override values are provided ->  all qos for lightOS are deleted; qos values are not shown in cinder.conf file
4. Set qos_enabled=false when there are qos in place -> all qos for lightOS are deleted; qos values are not shown in cinder.conf file
5. Set qos_enabled=false when there are no qos in place -> no action and cinder.conf remains the same
6. No qos_enabled tag when there isn't qos in place -> no action and cinder.conf remains the same
7. No qos_enabled tag when there is qos in place ->  all qos for lightOS are deleted; qos values are not shown in cinder.conf file


ansible/group_vars/all/overrides.yml.example
```diff
@@ -151,13 +151,21 @@ zones:
           type: ceph
           pool:
             name: volumes
+          qos:
+            read_bytes_sec: 750000001
```
tj90241, line 155: 750000000 + 1 byte? oddly specific number

zhanz1: Had the same question b4 but that's for the example file xD

chef/cookbooks/bcpc/recipes/cinder.rb
```diff
@@ -472,68 +522,55 @@
...
-    qos_name = "#{volume_type['name']}-qos"
-    block do
-      Chef::Resource::RubyBlock.send(:include, Chef::Mixin::ShellOut)
-
-      # create a qos policy
-      qos_create_opts = []
-
-      volume_type.fetch('limits', []).each do |key, value|
-        qos_create_opts.push("--property #{key}=#{value}")
+    ruby_block "create a qos policy for the volume type #{backend_name}" do
```
tj90241, line 526: Chef still provides the ability to use functions - given that that are 50+ lines here that I think are the same as above (if I'm reading correctly), I would probably suggest we pop them into a function to dedup some of the code.

tj90241:
lgtm - just take a look at
https://github.com/bloomberg/chef-bcpc/blob/main/chef/cookbooks/bcpc/libraries/utils.rb

as an idea to see if some of the QoS creation can be put into a function and called.

ansible/group_vars/all/overrides.yml.example
mzhang741: Yeah, it just an example to show that we can add override values this way. Any suggestion on what value to use or how to make it more clear?

chef/cookbooks/bcpc/recipes/cinder.rb
mzhang741: makes sense

ansible/group_vars/all/overrides.yml.example
tj90241: As it's the example, I don't think you need to change it. I missed that part as well and thought it was intended configuration.

mboruta1:
Code looks good to me; just wondering what type of testing was performed? Some test cases I can think of:

- Enable QoS policy on all backends (ceph and alternate) and verify
- Change QoS policy on all backends + verify
- Delete QoS policy on all backends + verify
- Repeat the above 3 tests with but while specifying custom policy for each backend

chef/cookbooks/bcpc/libraries/zone_config.rb
```diff
@@ -121,24 +134,43 @@ def alternate_backends
...
+        end
+
+        alternate_backends.append(alternate_backend_to_add)
       end
     end
 
     alternate_backends
   end
 
+  # Combine default qos limits with override values
```
npustchi, line 165: this can be a one line ternary operator, will be more readable.

chef/cookbooks/bcpc/libraries/zone_config.rb
```diff
@@ -121,24 +134,43 @@ def alternate_backends
...
           'volume_type_properties' => {}.merge(backend['volume_type_properties'])
-            .map { |k, v| "--property #{k}=#{v}" }.join(' '),
-        })
+                                        .map { |k, v| "--property #{k}=#{v}" }
+                                        .join(' '),
+        }
+
+        qos_enabled = general_backend_configs[backend['name']]['qos_enabled']
+        alternate_backend_to_add['qos_enabled'] = qos_enabled
+        if qos_enabled
```
npustchi, line 153: can you use ternary operator here ?

npustchi (approved):
LGTM, please test the null cases for both backends and qos.

chef/cookbooks/bcpc/libraries/zone_config.rb
mzhang741: make sense

mboruta1:
> Code looks good to me; just wondering what type of testing was performed? Some test cases I can think of:
> 
> * Enable QoS policy on all backends (ceph and alternate) and verify
> * Change QoS policy on all backends + verify
> * Delete QoS policy on all backends + verify
> * Repeat the above 3 tests with but while specifying custom policy for each backend

Apologies, somehow missed the testing procedures you detailed above. Looks good to me, +1 regarding Tyler's comment about using functions to cut down on the size of the recipe.

ansible/group_vars/all/chef_environment.yml.example
```diff
@@ -27,11 +27,13 @@ chef_environment:
         ceph:
           pool:
             size: 2
+          qos_enabled: false
         alternate_backends:
           enabled: true
           backends:
             - name: lightos
               enabled: true
+              qos_enabled: true
```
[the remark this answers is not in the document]
npustchi: we can do that, we just need to update the code as well.
though I referred line 30 and 36 to avoid mishaps between ceph and alternate backends.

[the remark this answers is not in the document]
mboruta1: The reason why I asked regarding `enabled` and `name` is because I believe it wouldn't make any sense to rename them with a `backend_` prefix; I feel the same way regarding `qos_enabled`. To put it another way, each entry in the list under 'backends' pertains to one backend; this is obvious and the fields defining said backend should remain generic. To do otherwise would be akin to creating the following structure:

```
struct {
  string backend_name,
  bool backend_enabled,
  bool backend_qos_enabled,
} backend_st
```

Its already clear from the name of the structure what the fields are all relating to; a backend.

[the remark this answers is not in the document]
comay: I tend to agree - avoid replicating the prefix that's under a group with the same name.

chef/cookbooks/bcpc/libraries/zone_config.rb
comay: Will it be more readable though? Maybe I'm misunderstanding which lines are in question.

npustchi: yes, It does for simple cases, when ternary operator  used appropriately it can make it easier to read and understand though not for complex conditions or when it is overused.

ansible/group_vars/all/chef_environment.yml.example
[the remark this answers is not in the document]
npustchi: line 30 refers to ceph, and line 36 is generic backends. having the same variable name will make bugs in the future.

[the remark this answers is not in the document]
mboruta1: As opposed to `name` or `enabled`? :) Or the countless other places in the our yaml files where generic names are used?

chef/cookbooks/bcpc/libraries/zone_config.rb
mzhang741: The tenary operator version would be as follows:

```
qos_limits = overrides.nil ? @zone_config.qos_default_limits : 
      @zone_config.qos_default_limits.merge(overrides)
```
current version:

```
qos_default_limits = @zone_config.qos_default_limits
if overrides.nil?
  return qos_default_limits
end
qos_default_limits.merge(overrides)
```

chef/cookbooks/bcpc/libraries/zone_config.rb
mzhang741: I understand using tenary operator would have fewer lines, but I think current version would be more readable. Any thoughts?

```
alternate_backend_to_add['qos'] = qos_enabled ? get_combined_qos_limits(backend['qos']) : nil
```

mboruta1: The latter is more readable in my view.

In general I would avoid using one-time-use temp variables and just break up long lines as needed:

```
if qos_enabled
   alternate_backend_to_add['qos'] = get_combined_qos_limits(backend['qos'])
end
```

chef/cookbooks/bcpc/libraries/zone_config.rb
mboruta1: Think in this case the ternary operator is more readable.

chef/cookbooks/bcpc/libraries/utils.rb
```diff
@@ -430,3 +430,41 @@ def etcd_advertised_name(member = node)
   hosts_to_cnames = node['bcpc']['etcd']['host_to_cnames']
   hosts_to_cnames.fetch(member['hostname'], member['service_ip'])
 end
+
+def create_qos_for_volume(backend)
```
mboruta1, line 434: I would suggest renaming this to `add_qos_to_volume_type(backend)` and updating the delete function to `remove_qos_from_volume_type(backend)`. Think this more clearly reflects what we are doing in these functions.

chef/cookbooks/bcpc/libraries/zone_config.rb
```diff
@@ -89,13 +93,22 @@ def backends
...
+      backend_to_add = {
         'name' => backend['name'],
         'private' => backend['private'],
         'client' => zone['ceph']['client'],
         'pool' => backend['pool']['name'],
         'libvirt_secret' => databag[zone['zone']]['libvirt']['secret'],
-      })
+      }
+
+      qos_enabled = @zone_config.node['bcpc']['cinder']['ceph']['qos_enabled']
```
mboruta1, line 105: Can you please add a comment (sentence or two) describing what this code does? May be trivial for some to understand, but not to me :).

tj90241 (approved):
lgtm less Mike's comments - thank you for looking into the functions and moving some code there, its a big improvement I think.

chef/cookbooks/bcpc/libraries/zone_config.rb
npustchi: why do you need this qos_enabled variable ?

chef/cookbooks/bcpc/libraries/utils.rb
mzhang741: make sense

chef/cookbooks/bcpc/libraries/zone_config.rb
mzhang741: okk

chef/cookbooks/bcpc/libraries/zone_config.rb
mzhang741: yeah it's just a local variable to hold the value of `general_backend_configs[backend['name']]['qos_enabled']`. I can get rid of it

npustchi: I dont think it is necessary, since you can just use the general_backend_configs, your call.

npustchi (approved):

chef/cookbooks/bcpc/libraries/zone_config.rb
mzhang741: agree

mboruta1 (approved):
Nice work!

PR #2264, opened 2024-02-08
Reinstate bpftool workaround for Felix

tj90241:
This was slated to be fixed in Calico 3.27.0, but the
BPF objects still appear to have not been linked against
a new enough version of libbpf. Thus, ensure that Felix
continues using a version of bpftool that's compatible
with it.

This is slightly less heinous than the former attempt
at a workaround and should allow us to remove it easier
in the future once it is fixed.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/calico-felix.rb
+++ b/chef/cookbooks/bcpc/recipes/calico-felix.rb
@@ -17,6 +17,50 @@
 
 include_recipe 'bcpc::calico-apt'
 
+# 22.04 + HWE-edge kernel (now kernel 6.x) provides a bpftool which is
+# linked against a newer version of libbpf that is incapable of loading
+# BPF objects that Felix needs. To workaround this, install the non-HWE
+# version of bpftool and have Felix use that one -- i.e., use the same
+# bpftool that would be installed had we not selected the HWE kernel.
+if Integer(node['kernel']['release'].split('.')[0]) == 6
+  linux_tools_version = ''
+
+  ruby_block 'Query the installed version of linux-tools-generic' do
+    block do
+      Chef::Resource::RubyBlock.send(:include, Chef::Mixin::ShellOut)
+      apt_query = "apt-cache show linux-tools-generic | awk -F ':' '/Version:/ {gsub(/ /, \"\", $2) ; print $2; exit}'"
+
+      # The version in the path and the package are different...
+      # The package has versions like x.y.z.m.n but the path to the
+      # tool is found at x.y.z-m instead... so regex it is:
+      linux_tools_version = shell_out(apt_query).stdout.chomp()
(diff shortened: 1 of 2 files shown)


tj90241:
Realized I forgot that this would have to be only done for non-storage nodes probably?

tj90241:
tl;dr:

When host is running kernel 6.x (22.04 + HWE-edge), then we want `calico-helix` to use a different version of `bpftool`. Tested by Cheffing this onto 20.04 nodes running 5.x and 22.04 worknodes running 6.x to verify that it only activates under the intended case.

chef/cookbooks/bcpc/templates/default/calico/override.conf.erb
```diff
@@ -0,0 +1,2 @@
+[Service]
+Environment="PATH=/usr/lib/linux-tools/<%= @linux_tools_version %>-generic:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```
comay, line 2: Nit - I believe in all of the cases we support (Focal and Jammy) that both `/bin` and `/sbin` are symlinks to their respective cousins so they can be removed out of the `PATH`.

chef/cookbooks/bcpc/recipes/calico-felix.rb
```diff
@@ -17,6 +17,50 @@
...
+# BPF objects that Felix needs. To workaround this, install the non-HWE
+# version of bpftool and have Felix use that one -- i.e., use the same
+# bpftool that would be installed had we not selected the HWE kernel.
+if Integer(node['kernel']['release'].split('.')[0]) == 6
+  linux_tools_version = ''
+
+  ruby_block 'Query the installed version of linux-tools-generic' do
+    block do
+      Chef::Resource::RubyBlock.send(:include, Chef::Mixin::ShellOut)
+      apt_query = "apt-cache show linux-tools-generic | awk -F ':' '/Version:/ { gsub(/ /, \"\", $2) ; print $2; exit}'"
```
comay, line 31: Nit - you have a space after then opening `{` but not before the closing `}`.

comay (dismissed):
See two very minor nits I noticed.

chef/cookbooks/bcpc/templates/default/calico/override.conf.erb
tj90241: I just copy/pasted $PATH from `/etc/environment` on a clean Ubuntu install. I don't think we should derive from whatever that specifies.

chef/cookbooks/bcpc/recipes/calico-felix.rb
tj90241: Removed the space before `gsub`.

comay (approved):

PR #2267, opened 2024-02-12
RDTIBCC-5070: nova-compute: Backport a libvirt fix

tj90241:
When live-migrating or starting VMs that birthed in a
Focal-based version of Nova, they will refuse to start
anymore in Jammy until their XML is altered. This commit
includes a backport of an upstream commit which makes the
pertinent changes to the XML file as needed.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/nova-compute.rb
+++ b/chef/cookbooks/bcpc/recipes/nova-compute.rb
@@ -271,8 +271,17 @@ end
 
 # Ensure that the MTU is not changed during a live-migration
 # https://bugs.launchpad.net/nova/+bug/1984009
+#
+# Also backport a bug stemming from a libvirt change:
+# https://bugs.launchpad.net/nova/+bug/1978489
 cookbook_file '/usr/lib/python3/dist-packages/nova/virt/libvirt/migration.py' do
   source 'nova/migration.py'
+  notifies :run, 'execute[py3compile-nova]', :immediately
+  notifies :restart, 'service[nova-compute]', :delayed
+end
+
+cookbook_file '/usr/lib/python3/dist-packages/nova/virt/libvirt/driver.py' do
+  source "nova/driver_#{node['platform_version']}.py"
   notifies :run, 'execute[py3compile-nova]', :immediately
   notifies :restart, 'service[nova-compute]', :delayed
 end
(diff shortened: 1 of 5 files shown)


tj90241:
https://review.opendev.org/c/openstack/nova/+/898554/2/nova/virt/libvirt/driver.py
for the changes to driver.py

chef/cookbooks/bcpc/recipes/nova-compute.rb
```diff
@@ -271,12 +271,21 @@
...
+# Also backport a bug stemming from a libvirt change:
+# https://bugs.launchpad.net/nova/+bug/1978489
 cookbook_file '/usr/lib/python3/dist-packages/nova/virt/libvirt/migration.py' do
   source 'nova/migration.py'
   notifies :run, 'execute[py3compile-nova]', :immediately
   notifies :restart, 'service[nova-compute]', :delayed
 end
 
+cookbook_file '/usr/lib/python3/dist-packages/nova/virt/libvirt/driver.py' do
+  source "nova/driver_#{node['platform_version']}.py"
```
psipika, line 284: 👏

psipika (dismissed):
Nice catch.

chef/cookbooks/bcpc/recipes/nova-compute.rb
tj90241: 🏃

tj90241: (the differences are sizable enough that I think it's not worth trying to poke with a stick at this point -- as long as things land in Jammy fine, that's all that should matter)

tj90241:
While this fixes Jammy, it introduces a bug on Focal.

The Jammy hypervisors cannot accept instances with the shares from Focal due to the fact that they use cgroupsv2, which only supports a maximum of 10,000 share. By default, Focal/nova assigns 1,000 shares per vCPU. So a domain which ran file in Focal fails to load on Jammy for > 10vCPU when not modified.

OTOH, adding this patch into Focal strips the shares from the domain entirely. Thus, if a Focal hypervisor to migrate an instance from itself to another Focal hypervisor, its shares will be stripped. Where this becomes problematic is when you consider the fact that the other domains on that destination Focal hypervisor may not yet have had their shares stripped. This will result in the other instances having 1,000s of more CPU time than the instance which was migrated and had its shares stripped.

Yet another sticky aspect is that it is the source hypervisor which is doing the migration operations and rewriting the domain XML. Thus it would seem that the source hypervisor needs to develop an awareness of the version of the hypervisor to which it is migrating if we are to proceed with this change?

psipika (approved):
🚀

PR #2268, opened 2024-02-15
calico-work: Disable and mask HAproxy

tj90241:
calico-compute seems to be indirectly pulling in HAproxy,
and it is started by default. This is not a service we
need running on worknodes with Calico, so disable it.

### DIFF
--- a/chef/cookbooks/bcpc/recipes/calico-work.rb
+++ b/chef/cookbooks/bcpc/recipes/calico-work.rb
@@ -1,7 +1,7 @@
 # Cookbook:: bcpc
 # Recipe:: calico-work
 #
-# Copyright:: 2023 Bloomberg Finance L.P.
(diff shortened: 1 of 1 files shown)


chef/cookbooks/bcpc/recipes/calico-work.rb
```diff
@@ -27,6 +27,12 @@
 
 service 'calico-dhcp-agent'
 
+# calico-compute pulls in HAproxy and starts it by default
+# This is more of a Neutron thing not needed by Calico, so disable it
+systemd_unit 'haproxy' do
+  action [:stop, :disable, :mask]
```
comay, line 33: Can this be folded into the section below so they're all consistent how we "disable" unneeded services?

comay (commented):
Also, don't need to fix but the Copyright year can be updated.

comay (approved):

PR #2271, opened 2024-02-16
Use 5-tuple (proto, addr, port) hashing for ECMP

tj90241:
Restrict to Jammy compute nodes for now, as we need to
ensure this is safe to apply to storagenodes at runtime
if we are to roll it there.


tj90241:
The gate might actually be needed. There seems to be severe issues with the Jammy/5.15 kernel.

PR #2273, opened 2024-02-17
Workaround nftables performance regression

tj90241:
calico-felix performs autodetection as to whether or not
to use nftables instead of iptables. On Focal nodes,
nftables was not installed and thus Calico used iptables
on Focal.

For Jammy, nftables is now installed by default and this
auto- detection results in nftables being used. For Ceph
workloads, performance with nftables was comparable to
iptables and as nftables is the forward-looking solution,
we went with that approach. However, the story is very
much different for some compute workloads at the moment,
and quite stark regressions in nftables have been seen.

For this reason, default to using iptables for anything
other than strictly Ceph nodes until its performance is
comparable.


tj90241:
Took two VMs, live-migrated them back and forth between two pairs of Jammy nodes. One pair has `nftables`-based policy evaluation, the other uses `iptables`-based policy evaluation. In addition to the visible performance differences, you run something like `perf top` on the hypervisor while it's running, you'll see a bunch of time being spent in a series of `nftables` symbols vs. a relatively fewer amount of the `iptables` ones.

Two VMs doing an `iperf3` session across a VXLAN overlay (60 seconds averaged):
```
Jammy nftables:
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-60.00  sec  31.8 GBytes  4.55 Gbits/sec    0             sender
[  5]   0.00-60.04  sec  31.8 GBytes  4.55 Gbits/sec                  receiver

Jammy iptables:
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-60.00  sec  39.4 GBytes  5.64 Gbits/sec    0             sender
[  5]   0.00-60.05  sec  39.4 GBytes  5.64 Gbits/sec                  receiver

Focal iptables:
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-60.00  sec  44.7 GBytes  6.40 Gbits/sec    0             sender
[  5]   0.00-60.04  sec  44.7 GBytes  6.39 Gbits/sec                  receiver
```

= >20%... the result does fluctuate by +/- 100Mbps or so, but is still repeatedly decisive. And, unfortunately, this result itself is still behind Focal even still...

tj90241:
Though in the straight TCP tests not using a VXLAN overlay like this, Jammy is now topping Focal by a large margin with this and other recent changes to the NIC drivers:
```
Jammy iptables:
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-60.00  sec   132 GBytes  18.8 Gbits/sec  2111             sender
[  5]   0.00-60.04  sec   132 GBytes  18.8 Gbits/sec                  receiver

Focal iptables:
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-60.00  sec   122 GBytes  17.5 Gbits/sec  1340             sender
[  5]   0.00-60.04  sec   122 GBytes  17.4 Gbits/sec                  receiver
```

zhanz1:
I also looked into this config and asked DC, but I skipped it because we were not aware that Jammy has `nftables` by default (and thus `Auto` won't be just picking `iptables` like it is in Focal). Need to recheck my list of configs against Jammy next week :O

chef/cookbooks/bcpc/templates/default/calico/override.conf.erb
```diff
@@ -1,2 +1,11 @@
...
 Environment="PATH=/usr/lib/linux-tools/<%= @linux_tools_version %>-generic:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+# Significant regressions in performance are noted for some workloads when
+# nftables is used vs. iptables. This does not seem to have impacted Ceph
+# workloads much, but does impact some hypervisor workloads considerably.
+#
+# NOTE: In order for this setting to change, hosts *MUST* be rebooted. Do
+# not attempt to change this dynamically without considerable testing.
+<% unless @is_ceph_node or @is_focal_node %>
+Environment="FELIX_IPTABLESBACKEND=Legacy"
+<% end %>
```
zhanz1, line 11: Wonder what's the reason that we want to put this in the environment variable instead of changing the Felix config file (just curious bc I submitted a change for another config)? 🤔

tj90241: ah, I guess I didn't think to look if it was a config file setting - I'll move it there.

zhanz1: https://docs.tigera.io/calico/latest/reference/felix/configuration => `IptablesBackend`

zhanz1:
LGTM but will w8 for another pair of eyes on this :)

chef/cookbooks/bcpc/templates/default/calico/felix.cfg.erb
```diff
@@ -51,3 +51,13 @@ AllowVXLANPacketsFromWorkloads = true
 
 # Set to true to allow IPIP encapsulated traffic from workloads.
 AllowIPIPPacketsFromWorkloads = true
+<% unless @is_ceph_node or @is_focal_node %>
```
comay, line 54: I'm a bit uneasy with using different implementations on nodes with different roles. If `nftables` is going to be used on a subset of nodes, have we at least verified the existing `iptables` commands can be used there as documented on https://www.nftables.org/projects/nftables/index.html?

Alternatively since you said the performance is compatable, perhaps we should stick with the legacy `iptables` until we can deploy it across all roles.
```
nft syntax differs from {ip,ip6,eb,arp}tables. Moreover, there is a backward compatibility layer that allows you run iptables/ip6tables, using the same syntax, over the nftables infrastructure.
```

tj90241: Of course, it's not ideal... a 30% performance hit in a VXLAN test a big carving of meat off the bone. If the decision is to accept the performance hit to keep things unified, I have no concerns closing this PR.

The tools work all the same AFAICT - you can run `iptables` on a Ceph node and run `iptables-legacy` on a Jammy compute node and get similar results/functionality out of them.

mboruta1: I believe DC was suggesting keeping everything at iptables and thus avoiding any performance hit. This is what I would advocate for; keep using `iptables` and investigate `nftables` performance regressions later this year. nftables was made available in 3.13 so not a new project, perhaps there is something we could tune on our end to improve the numbers.

tj90241: I was mostly sympathizing with:
> I'm a bit uneasy with using different implementations on nodes with different roles.

Totally agree in this case - I wish we had known this in advance of the Ceph work and would love to see parity here.

mboruta1 (approved):
Nice find!

chef/cookbooks/bcpc/templates/default/calico/felix.cfg.erb
comay: > Understood, and we do not want to revert to iptables for storage nodes because it would require a full reboot of all our storage nodes. Oh well :(, it is what it is.

I don't think reverting back to `iptables` necessitates a reboot or specifically, I'm not advocating that be done. Rather by setting things to use `iptables`, on the _next_ reboot we'll lazily switch.

To be clear, my concern around consistency isn't so much around the internal implementation although I think that's not to be discounted. It's more around what the cloud operator needs to do when dropping down into the commands to look at sets, chains, tables, etc. Supposedly there is some sort of compatibility layer but clearly it does require a second CLI implementation to access it which means it's not obviously compatible to me. If we can make things such that the operation can continue to commands called `iptables` and `ipset` but has different implementations in the kernel that are optimal for the role, that would be perfect. If the command names are different or their output is different, that's unfortunate (although given I see `/etc/alternatives` are in use, perhaps one can set things accordingly).

But I worry if I'm going to need to remember which type of node I'm on in order to decide to run `iptables` or `iptables-legacy`, etc.

Finally, I thought Tyler said the performance was comparable between the two implementations on the Ceph-side - that's one reason to just stick with the older one for all nodes now until there's a reason to move.

mihalis68: Coming to this conversation, but it seems no the performance is not comparable (going off data presented above): nftables is worse with vxlan over udp (hyps) and better with tcp, but we almost randomly landed on the ones that are better on the various node types. Achieving consistency would come at a performance hit to one or the other.

tj90241: Are we okay merging this now that the above issue is being worked on in https://github.com/bloomberg/chef-bcpc/pull/2298?

comay: I suppose so although it really would be nice to make sure they're part of the same release.

PR #2276, opened 2024-02-22
RDTIBCCOPT-152: mysql: Change error log location and verbosity

mboruta1:
By default `mysql` 5.7 logs error messages to stderr with a verbosity of Error, Warning, and Note. This behavior is over-written by `mysqld_safe`, the script called by `/etc/init.d/mysql` that executes `mysql`, to instead route stderr to a file in the mysql data directory.

Said error log file is not rotated via the logrotate configuration file bundled with Percona XtraDB Cluster 5.7 because this method of logging errors was deemed 'obsolete', with syslog being recommended.

This commit routes mysql error messages to syslog by default, and limits the verbosity to only Error and Warning. Thus mysql error logs are automatically rotated as part of syslog rotation, and `Note` messages omitted.

**Describe your changes**
See the commit message.

**Testing performed**
Tested on an internal test cluster. New defaults are properly applied on existing clusters and can be rolled back by setting `log_error_verbosity` to 3 and `log_error_location` to `skip-syslog`. 

**Additional context**
None.

### DIFF
--- a/chef/cookbooks/bcpc/attributes/mysql.rb
+++ b/chef/cookbooks/bcpc/attributes/mysql.rb
@@ -28,6 +28,10 @@ default['bcpc']['mysql']['thread_cache_size'] = 1024
 default['bcpc']['mysql']['tmp_table_size'] = '128M'
 default['bcpc']['mysql']['wsrep_slave_threads'] = 16
 
+# error log settings
+default['bcpc']['mysql']['log_error_verbosity'] = 2
+default['bcpc']['mysql']['log_error_location'] = 'syslog'
+
 # slow query log settings
 default['bcpc']['mysql']['slow_query_log'] = true
 default['bcpc']['mysql']['slow_query_log_file'] = '/var/log/mysql/slow.log'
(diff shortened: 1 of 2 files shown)


zhanz1 (approved):
LGTM. Just a quick question that's not related to this PR, for the slow query logs, do we (and can we) also want to include them in `syslog` or we keep it as it is?

tj90241 (approved):
Holey moley, nice find!

mboruta1:
> LGTM. Just a quick question that's not related to this PR, for the slow query logs, do we (and can we) also want to include them in `syslog` or we keep it as it is?

Does not look like it according to [this|https://docs.oracle.com/cd/E17952_01/mysql-5.7-en/slow-query-log.html], and even if it was possible I would prefer to keep things as they are: at Percona's intended defaults.

zhanz1:
> > LGTM. Just a quick question that's not related to this PR, for the slow query logs, do we (and can we) also want to include them in `syslog` or we keep it as it is?
> 
> Does not look like it according to [this|https://docs.oracle.com/cd/E17952_01/mysql-5.7-en/slow-query-log.html], and even if it was possible I would prefer to keep things as they are: at Percona's intended defaults.

Makes sense :). Thanks for sharing the doc too!

PR #2278, opened 2024-02-26
RDTIBCC-5065: Add PKU feature flag when migrating towards Jammy

tj90241:
A recent upstream kernel change [1] addressed a bug wherein
XSAVE stored a PKU feature flag as part of the vCPU context,
but XRSTOR did not load it. A side-effect of this change is
that while VMs which ran on a hypervisor whose host CPU
supported PKU, but which did not expose the same to the
guest, can no longer live-migrate to those same
hypervisors after they said hypervisors are running a
kernel with the aforementioned patch applied.

To permit live-migration in the face of this issue, we use
the Nova live-migration window to add the PKU flag to the
guests as part of live-migrations to kernels which have
effectively have [1] applied to them.

[1] https://lore.kernel.org/lkml/20220205081658.562208-1-leobras@redhat.com/T/

### DIFF
--- a/chef/cookbooks/bcpc/files/default/nova/migration.py
+++ b/chef/cookbooks/bcpc/files/default/nova/migration.py
@@ -57,6 +57,7 @@ def get_updated_guest_xml(instance, guest, migrate_data, get_volume_config,
                           get_vif_config=None, new_resources=None,
                           is_target_focal=False):
     xml_doc = etree.fromstring(guest.get_xml_desc(dump_migratable=True))
+    xml_doc = _update_cpu_xml(xml_doc, is_target_focal)
     xml_doc = _update_graphics_xml(xml_doc, migrate_data)
     xml_doc = _update_serial_xml(xml_doc, migrate_data)
     xml_doc = _update_volume_xml(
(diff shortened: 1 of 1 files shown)


tj90241:
Unfortunately, the former approach for addressing this bug (namely, trying to take features away from the VM in order to address the bug referenced int he commit) does not work well with non-Linux-based guest operating systems. Hence, this is a new approach that seems to cover all bases I've seen so far which _adds_ the feature that was never meant to be as part of live-migration.

Associated changes must be made to `nova.conf` in cluster overrides when the CPU model in `nova.conf` is "Broadwell" -- specifically, it must have `pku` added to its list of additional flags. This essentially means that the VMs can no longer run on true Broadwell silicon - as Broadwell never shipped with PKU. But, so far, I have seen that this is the only way to effectively solve this issue in all cases. 

Note that when this change is applied and a live-migration is issued from a Focal hypervisor to Jammy hypervisor, the below error will be realized and the VM will not migrate:
```
Live Migration failure: unsupported configuration: Target CPU feature count 10 does not match source 9: libvirt.libvirtError: unsupported configuration: Target CPU feature count 10 does not match source 9
```

This is expected -- there are associated changes which must be made to `libvirt` in order for the live-migration to proceed.

comay (approved):

PR #2280, opened 2024-02-27
Remove files, directories referering to previous 'ufw' invocation, superseded by Calico rules

desrod:
This closes out RDITBCC-2045 with a vastly simplified implementation of removing ufw and its configuration from bootstrap nodes, where the firewall rules were superseded by Calico. 

The original playbook was much broader in scope and attempted to identify whether ufw was a.) installed, b.) configured, and c.) enabled before taking action. This is a larger hammer, and just removes it entirely.


tj90241 (changes requested):
Need to also remove `bcpc::ufw` from the the run list of any role which references it. I think you if you try to delete this recipe and run `chef-client`, it should fail as-is?

desrod:
> Need to also remove `bcpc::ufw` from the the run list of any role which references it. 

That was removed. Ah shoot, I forgot to push the actual file in the commit. One moment, amending.

comay (commented):
I believe `ansible/playbooks/roles/common/tasks/configure-system.yml` needs to be changed to add `ufw` to the list of services/packages under `Stop and disable unused services` and `Purge unused service packages`, respectively (please retain the sorted order).

This is necessary because even though the recipe was only applicable to the `bootstrap` role, the service/package are part of the base OS images.

desrod:
Removed my dedicated ufw play and replaced it with the addition of that service to the existing stop/purge plays that I didn't realize existed. 🙃 Now even more vastly simplified.

tj90241:
Only other things:
* commit signoff (DCO bot unhappy)
* cleanup/deployment items:
   * run list is not updated dynamically, so the release manager has to know to go into the bootstrap and meddle post release I think...
   * note to us: we can remove the `ufw` package removal after 8.4.13 deploys because `ufw` is not included in a base `curtin` install.

tj90241 (approved):
lgtm

comay (approved):
Please remember to squash the commits after rebasing your changes.

PR #2283, opened 2024-03-01
Increase token_cache_time for keystone_authtoken on Openstack Services

zhanz1:
## WHY
To prevent excessive effort for validating tokens, the `keystone_authtoken` (or `keystonemiddleware`) will cache previously seen tokens and set the cache entry's timeout to be `token_cache_time`. The default 300s is good, but we can take a step futher to increase it to 600s, so the token can stay in cache for a longer period of time. This will be beneficial when there are a lot of requests and thus a lot interactions between Openstack services.

## WHAT
```
token_cache_time = 300s => 600s
```

## HOW
`make all`!

## TEST
After building the testing cluster, the following configs are included in the file:
```
% ansible -i ansible/inventory.yml test-cloud-nodes -b -m shell -b -a "grep -Rnw /etc/*/*.conf -e token_cache_time"
test-cloud-node-1 | CHANGED | rc=0 >>
/etc/cinder/cinder.conf:61:token_cache_time = 600
/etc/glance/glance-api.conf:52:token_cache_time = 600
/etc/heat/heat.conf:86:token_cache_time = 600
/etc/neutron/neutron.conf:41:token_cache_time = 600
/etc/nova/nova.conf:64:token_cache_time = 600
/etc/placement/placement.conf:22:token_cache_time = 600
```

To make sure the config is loaded, we can check the log message of an Openstack service:
```
# Nova
root@test-cloud-node-1:~# grep -Rnw /var/log/nova/nova-api.log -e token_cache_time
35510:2024-03-01 16:01:24.562 704724 DEBUG oslo_service.service [-] keystone_authtoken.token_cache_time = 600 log_opt_values /usr/lib/python3/dist-packages/oslo_config/cfg.py:2609

# Cinder
root@test-cloud-node-1:~# grep -Rnw /var/log/cinder/ -e token_cache_time
/var/log/cinder/cinder-scheduler.log:21285:2024-03-01 16:06:20.237 705779 DEBUG cinder.service [req-d3744fef-bece-4121-bc35-12af7a0d6120 - - - - -] keystone_authtoken.token_cache_time = 600 log_opt_values /usr/lib/python3/dist-packages/oslo_config/cfg.py:2609
/var/log/cinder/cinder-scheduler.log:21719:2024-03-01 16:06:20.337 705779 DEBUG oslo_service.service [req-d3744fef-bece-4121-bc35-12af7a0d6120 - - - - -] keystone_authtoken.token_cache_time = 600 log_opt_values /usr/lib/python3/dist-packages/oslo_config/cfg.py:2609
/var/log/cinder/cinder-volume.log:273:2024-03-01 16:06:20.295 705781 DEBUG oslo_service.service [req-e99b8702-d0b2-4423-bdaf-265fc8a7a71a - - - - -] keystone_authtoken.token_cache_time = 600 log_opt_values /usr/lib/python3/dist-packages/oslo_config/cfg.py:2609
```

### DIFF
--- a/chef/cookbooks/bcpc/attributes/cinder.rb
+++ b/chef/cookbooks/bcpc/attributes/cinder.rb
@@ -32,3 +32,6 @@ default['bcpc']['cinder']['alternate_backends']['enabled'] = false
 default['bcpc']['cinder']['alternate_backends']['backends'] = []
 default['bcpc']['cinder']['alternate_backends']['cinder_internal_tenant_project_id'] = nil
 default['bcpc']['cinder']['alternate_backends']['cinder_internal_tenant_user_id'] = nil
+
+# specify keystone_authtoken configs
+default['bcpc']['cinder']['keystone_authtoken']['token_cache_time'] = 600
--- a/chef/cookbooks/bcpc/attributes/glance.rb
+++ b/chef/cookbooks/bcpc/attributes/glance.rb
@@ -22,3 +22,6 @@ default['bcpc']['glance']['ceph']['pool']['size'] = 3
 # image format
 default['bcpc']['glance']['image_format']['container_formats'] = ['bare']
 default['bcpc']['glance']['image_format']['disk_formats'] = ['raw']
+
+# specify keystone_authtoken configs
+default['bcpc']['glance']['keystone_authtoken']['token_cache_time'] = 600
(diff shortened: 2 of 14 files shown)


mboruta1:
LGTM, any reason why `watcher`'s config was not updated in a similar manner as well?

zhanz1:
> LGTM, any reason why `watcher`'s config was not updated in a similar manner as well?

I didn't know that was also Openstack :panic: 😢

mboruta1 (approved):
Nice! Perhaps in the future we will want to have a global `token_cache_time` option that may or may not be over-ridden by individual service configs, but until then this looks good to me.

zhanz1:
> Nice! Perhaps in the future we will want to have a global `token_cache_time` option that may or may not be over-ridden by individual service configs, but until then this looks good to me.

Believe so too. Was thinking about using the same variable for all but decided to stay consistent with how other configs are placed (like database's connection max size). We may want to clean things up in the future.

tj90241 (approved):

mboruta1 (approved):

PR #2284, opened 2024-03-07
RDTIBCCOPT-382: Replace deprecated configs + image cache dir

mboruta1:
The image cache is effectively disabled due to `flavor = keystone` under `paste_deploy`, remove reference to an image cache dir.

We use an RBD backend to store images, ephemeral volumes, and volumes. All RBD volumes, including ephemeral ones, created from images are created within the RBD backend using copy-on-write, bypassing the caching layer even if it were enabled. Non-RBD volumes created from images will utilize cinder's volume image cache.

**Describe your changes**
Replaced deprecated configs `stores` and `default_store` with `enabled_backends` and `default_backend`. Also removed `image_cache_dir` since it is not being utilized. Specifically, as per Ceph's [instructions](https://docs.ceph.com/en/latest/rbd/rbd-openstack/#disable-cache-management-any-openstack-version) for integrating with OpenStack, we have set [flavor = keystone](https://github.com/bloomberg/chef-bcpc/blob/development/chef/cookbooks/bcpc/templates/default/glance/glance-api.conf.erb#L66) under `paste_deploy`, thus disabling Glance's caching layer. See [this](https://docs.openstack.org/glance/latest/configuration/configuring.html#enabling-the-image-cache-middleware) for additional information.

**Testing performed**
Deployed on a test cluster, launched and deleted a VM + downloaded and uploaded an image,

**Additional context**
None.

### DIFF
--- a/chef/cookbooks/bcpc/templates/default/glance/glance-api.conf.erb
+++ b/chef/cookbooks/bcpc/templates/default/glance/glance-api.conf.erb
@@ -6,6 +6,7 @@
 # go to: https://docs.openstack.org/glance/yoga/configuration/glance_api.html
 #
 [DEFAULT]
+enabled_backends = ceph:rbd
 debug = <%= node['bcpc']['glance']['debug'] %>
 show_image_direct_url = true
 enable_v1_api = false
@@ -18,7 +19,6 @@ workers = <%= node['bcpc']['glance']['workers'] %>
 workers = <%= node['bcpc']['openstack']['services']['workers'] %>
 <% end %>
 admin_role = <%= node['bcpc']['keystone']['roles']['admin'] %>
-image_cache_dir = /var/lib/glance/image-cache/
 log_file = /var/log/glance/api.log
 transport_url = rabbit://<%= @rmqnodes.map{|x| "#{@config['rabbit']['username']}:#{@config['rabbit']['password']}@#{x['service_ip']}:5672" }.join(',') %>
 image_member_quota = -1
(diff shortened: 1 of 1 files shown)


zhanz1 (approved):
LGTM, with `enabled_backends` we're now enabling Glance multistore (even though we only have one backend lol)!  The only concern I had is actually `rbd_thin_provisioning`: if we move the config out of `[glance_store]` section to `[ceph]` section, will the behavior change? A quick look up in the [source code](https://github.com/openstack/glance/blob/519ce0b94dcc6dad229fbe08d8d21ccdefb6c99e/glance/api/v2/discovery.py#L128) and I believe it should fall under the custom store's config, so we're good :).

mboruta1:
> LGTM, with `enabled_backends` we're now enabling Glance multistore (even though we only have one backend lol)! The only concern I had is actually `rbd_thin_provisioning`: if we move the config out of `[glance_store]` section to `[ceph]` section, will the behavior change? A quick look up in the [source code](https://github.com/openstack/glance/blob/519ce0b94dcc6dad229fbe08d8d21ccdefb6c99e/glance/api/v2/discovery.py#L128), and believe it should fall under the custom store section, so we're good :).

Behavior will not change as currently all rbd_* params are passed into the rbd driver, which is our one and only backend. If PR is approved, said params will continue to be passed to the rbd driver used by our named backend, `ceph`. Also, once `stores` and `default_stores` are removed we will have to utilize `enabled_backends` anyway (see https://docs.openstack.org/glance/latest/configuration/configuring.html#configuring-glance-storage-backends)

comay (approved):
Nice work, MIke!

PR #2287, opened 2024-03-12
Explicitly install `rsyslog` package - RDTIBCC-5134

comay:
After `ufw` was removed in #2280, we began to see build failures where certain components failed to install due to a missing `rsyslog` service. Since there are a number of consumers of `rsyslog`, we should just add it to the list of common packages.

This PR also includes the removal of the installation of `/etc/updatedb.conf` which is no longer used as the client program is no longer installed. 

Tested via a full build and verified that the `rsyslog` service is present and healthy.


ansible/playbooks/roles/common/tasks/configure-system.yml
```diff
@@ -82,14 +81,12 @@
...
 
-# Configure /etc/updatedb.conf
-- name: Configure /etc/updatedb.conf
-  copy:
-    src: files/updatedb/conf
-    dest: /etc/updatedb.conf
-    owner: root
-    group: root
-    mode: '0644'
+# TODO: Remove stanza after a couple of releases have passed
```
tj90241, line 84: Do you have a quick tl;dr of why we are removing this instead of carrying it over as-is?

comay: In the ye olde days, we use to have `locate(1)` or `mlocate(1)` installed but when we switched to Focal (I think), that ended up disappearing because their corresponding packages were no longer included in the default installation image. So this is a bit of cleanup I noticed could be done while looking at this file.

tj90241 (approved):
🚀

PR #2290, opened 2024-03-20
Chrony optimizaitons to speed up time acquisition, dependent service startup speed

desrod:
This PR encompasses a few changes all focused on speeding up chrony and dependent service startup. These include: 

* Adding 'iburst' to the chrony config for each of the 3 defined NTP servers. Another option is to 'peer' 2 of the 3 servers with each other (latter not included in this PR)
* Disabling the `systemd-networkd-wait-online` service which adds a significant delay to the startup order. Since each machine has a static transit, 'waiting' to hear back from the network is adding 1-2 minutes of time to the boot time. 
* Added `optional: true` to each interface's netplan configuration, also adds bootup performance (https://netplan.io/faq#prevent-waiting-for-interface) 
* Asserted the `clocksource` of `tsc` for baremetal. This should be `kvm` for VMs and `tsc` for baremetal. When not defined, it is probed (https://www.kernel.org/doc/Documentation/timers/timekeeping.txt) 

The combination of each of these has made a measurable difference in the bootup speed of the host as well as speeding up the acquision of time from those NTP servers from 2 minutes to 3 seconds. 

It's a relatively simple PR, but the testing is what took the longest time, to get this down to the smallest manageable subset of non-distruptive changes. 

Comments and criticism welcome!


comay (approved):
Nice work, @desrod - before merging, please respond to Chris' question.

tj90241:
We have an internal service which places a `Wants=` dependency on `systemd-networkd-wait-online.service`. Although `Wants=` is a weak dependency, we should probably make such services no longer depend on `systemd-networkd-wait-online.service`.

ansible/playbooks/roles/common/templates/grub/default.j2
```diff
@@ -18,9 +18,10 @@ GRUB_TIMEOUT=0
 GRUB_DISTRIBUTOR=`lsb_release -i -s 2> /dev/null || echo Debian`
 GRUB_CMDLINE_LINUX_DEFAULT="vga=normal nomodeset 3 console=tty0 console=tty1"
 {% set ignore_ce = 'mce=ignore_ce' if not hardware_edac_polling else '' -%}
+{% set clocksource_cmdline = 'clocksource=tsc' if system_set_clocksource else '' -%}
```
tj90241, line 21: hmm... is there a benefit to this? For things like virtual/CI builds, we likely want `cvm-clock` here instead of `tsc`, and AFAIK `tsc` is already the default as long as it's determined to be reliable. Is there a case where the TSC was not reliable/default but we determine that it should still be the primary clocksource?

desrod:
> Can you explain what testing you did to verify this please?

Will do. I'll anonymize hostnames and details and provide `systemd-analyze` outputs and timings from other sources to back up this change.

ansible/playbooks/roles/common/templates/grub/default.j2
desrod: You're correct. `tsc` is the preferred clocksource on 64-bit arch, and has a _very negative impact_ when used inside VMs, which should be using `kvm`. 

This particular bit probably needs refactoring to detect baremetal vs. virtual, so it doesn't fall back to `acpi_pm` or non-`tsc` clocksources there. 

I'll take this out for now until I can sort out a better way to handle that, so the kernel doesn't have to 'decide' which one is best at boot time.

desrod:
While debugging this, I found that `systemd-bootchart` fails to run correctly on your hosts, segfaults for reasons I don't know yet. I was hoping to get granular details on the boot time per-service. Instead, I was able to quantify this with two other methods. Hosts and other identifying information has been redacted from the data below. 

**Host with chrony optimizations:** 

```
$ cat /sys/firmware/acpi/fpdt/boot/*
17848081060039
17779508206301
18025466245662
18023585955756
0
```

**Boot start time and relevant journal entries:** 

```
0 28ae7cd183924198bbdff42be30021f6 Sat 2024-03-23 18:50:13 UTC
```

```
Mar 23 18:50:14 chronyd[1441]: chronyd version 3.5 starting (+CMDMON +NTP +REFCLOCK +RTC +PRIVDROP +SCFILTER +SIGND +ASYNCDNS +SECHASH +IPV6 -DEBUG)
Mar 23 18:50:14 chronyd[1441]: Frequency -10.359 +/- 1.541 ppm read from /var/lib/chrony/chrony.drift
Mar 23 18:50:14 chronyd[1441]: Loaded seccomp filter
Mar 23 18:50:15 chronyd[1441]: Source 1.2.3.10 offline
Mar 23 18:50:15 chronyd[1441]: Source 1.2.3.12 offline
Mar 23 18:50:15 chronyd[1441]: Source 1.2.3.11 offline
Mar 23 18:50:16 chronyd[1441]: Source 1.2.3.10 online
Mar 23 18:50:16 chronyd[1441]: Source 1.2.3.12 online
Mar 23 18:50:16 chronyd[1441]: Source 1.2.3.11 online
Mar 23 18:50:21 chronyd[1441]: Selected source 1.2.3.11
Mar 23 18:50:22 chronyd[1441]: Selected source 1.2.3.10
Mar 23 18:51:27 chronyd[1441]: Selected source 1.2.3.12
Mar 23 19:00:04 chronyd[1441]: Selected source 1.2.3.11
```


**Host without any chrony optimization:** 

```
$ cat /sys/firmware/acpi/fpdt/boot/*
4337065967603
4331545703865
4395878351072
4393729402266
0
```

**Boot start time and relevant journal entries:** 

``` 
0 32a3e9a0c3644bd6a2695b0e8d4fbd27 Sat 2024-03-23 18:53:22 UTC—
```

```
Mar 23 18:53:23 chronyd[1471]: chronyd version 3.5 starting (+CMDMON +NTP +REFCLOCK +RTC +PRIVDROP +SCFILTER +SIGND +ASYNCDNS +SECHASH +IPV6 -DEBUG)
Mar 23 18:53:23 chronyd[1471]: Frequency -8.732 +/- 0.764 ppm read from /var/lib/chrony/chrony.drift
Mar 23 18:53:23 chronyd-starter.sh[1464]: WARNING: libcap needs an update (cap=40 should have a name).
Mar 23 18:53:23 chronyd[1471]: Loaded seccomp filter
Mar 23 18:53:24 chronyd[1471]: Source 1.2.3.12 offline
Mar 23 18:53:24 chronyd[1471]: Source 1.2.3.11 offline
Mar 23 18:53:24 chronyd[1471]: Source 1.2.3.10 offline
Mar 23 18:53:25 chronyd[1471]: Source 1.2.3.12 online
Mar 23 18:53:25 chronyd[1471]: Source 1.2.3.11 online
Mar 23 18:53:25 chronyd[1471]: Source 1.2.3.10 online
Mar 23 18:56:37 chronyd[1471]: Selected source 1.2.3.12
```

The important part here is that both hosts were booted at the same exact time, but the host _without_ the chrony optmizations took 3 minutes longer to boot and acquire time than the other.

npustchi (approved):

PR #2292, opened 2024-03-21
Rewrite Ceph reweight logic with a cleaner, faster alternative

desrod:
Just a bit of optimization I found while rebuilding Thermite multiple times. This reduces several lines of code and repeated subshells to call `awk` and `bc` for string formatting into 2 lines of code. 

Further optimizations could be made to slice the vars and do some string formatting, but this should be sufficient here.

### DIFF
--- a/ansible/playbooks/roles/common/tasks/reweight-ceph-osds.yml
+++ b/ansible/playbooks/roles/common/tasks/reweight-ceph-osds.yml
@@ -4,19 +4,10 @@
     set -o pipefail
     set -xe
 
-    ceph_df=$(ceph osd df tree --format json)
-    ceph_osd_id_size_map=$(echo $ceph_df | \
-      jq '.nodes[] | select(.type == "osd") | .id,.kb' | \
-      paste - -)
-
-    IFS=$'\n'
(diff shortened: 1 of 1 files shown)


desrod:
The only difference between the two, is the output being treated as a proper float: 

```
ceph osd crush reweight osd.0 0.01819
ceph osd crush reweight osd.4 0.01819
ceph osd crush reweight osd.6 0.01819
ceph osd crush reweight osd.2 0.01819
ceph osd crush reweight osd.3 0.01819
ceph osd crush reweight osd.7 0.01819
ceph osd crush reweight osd.1 0.01819
ceph osd crush reweight osd.5 0.01819
ceph osd crush reweight osd.8 0.01819
```

This can be altered if needed, but to Ceph, it's the same (.0181 vs. 0.0181) 

An updated variant that preserves the format _**identically**_, is: 

```
ceph osd df tree --format json | jq -rc '.nodes[] | select(.type == "osd") | [.id, (.kb / pow(1024;3))]|@tsv' \
    | while IFS=$'\t' read -r id size_tb; do
          ceph osd crush reweight osd.${id} ${size_tb:1:6};
      done;
```

PR #2297, opened 2024-04-01
Set `heartbeat_in_pthread` to False for Non-WSGI Services

zhanz1:
## WHY
On our clusters we are consistently observing `greenlet.error: cannot switch to a different thread` error for Nova and Neutron services. This is due to [a bug](https://github.com/eventlet/eventlet/issues/432) that is reported in `eventlet` library, and is affecting `oslo.messaging`. To fix this issue, as pointed out by [the commit message](https://github.com/openstack/oslo.messaging/commit/44b3427bc9efea9f341edfb8ea7aea38f25d1a5a) from `oslo.messaging` and the [bug report](https://bugs.launchpad.net/neutron/+bug/1965140/comments/8), we need to disable this config for non-wsgi services (nova and neutron). 

The said feature is still in progress of being fixed (which is why they took it off from deprecation list), and we'll need to follow up when this issue is fixed in the future. 

## WHAT
Change `heartbeat_in_pthread` config option to be false for non-wsgi services (nova, neutron). 

## HOW
`make all`!

#### Before changes
`greenlet.error: cannot switch to a different thread` error being observed when RabbitMQ servers are taken down. 

#### After changes
Manually bringing down one RabbitMQ server on testing cluster and this error was not observed. 

#### Check Config Changes
```
# On a headnode
$ sudo grep -Rnw /etc/*/*.conf -e heartbeat_in_pthread
/etc/cinder/cinder.conf:40:heartbeat_in_pthread = false
/etc/glance/glance-api.conf:60:heartbeat_in_pthread = false
/etc/heat/heat.conf:105:heartbeat_in_pthread = false
/etc/neutron/neutron.conf:57:heartbeat_in_pthread = false
/etc/nova/nova.conf:114:heartbeat_in_pthread = false
/etc/watcher/watcher.conf:30:heartbeat_in_pthread = false

# On a worknode
$ sudo grep -Rnw /etc/*/*.conf -e heartbeat_in_pthread
/etc/nova/nova.conf:114:heartbeat_in_pthread = false
```


psipika (dismissed):
✅

chef/cookbooks/bcpc/templates/default/nova/nova-compute.conf.erb
```diff
@@ -25,3 +25,6 @@ cpu_model = <%= @cpu_model %>
 <% unless @cpu_model_extra_flag.empty? %>
 cpu_model_extra_flags = <%= @cpu_model_extra_flag.join(',') %>
 <% end -%>
+
```
comay, line 28: As you're already setting this in `/etc/nova/nova.conf`, you shouldn't need to do this here as well as the script is called with the arguments `--config-file /etc/nova/nova.conf --config-file /etc/nova/nova-compute.conf`.

chef/cookbooks/bcpc/attributes/nova.rb
```diff
@@ -25,6 +25,13 @@
 default['bcpc']['nova']['reserved_host_memory_mb'] = 1024
 default['bcpc']['nova']['cpu_allocation_ratio'] = 2.0
 
+# oslo messaging
```
comay, line 28: Nit - in `neutron.rb` you have this comment as `oslo.messaging` so please do the same ere.

chef/cookbooks/bcpc/templates/default/nova/nova-compute.conf.erb
zhanz1: Didn't actually know this and thank you! :D

npustchi (dismissed):

zhanz1:
> How was this tested ? Regarding it fixes the issue, any test cluster ?

On our gamma cluster, I rolled out the changes on `neutron-server` and all nova services. Then I restarted a RMQ server (mimicking a maintenance/down situation) and observed if any `greenlet error` pops up from logs. No such log was found.

zhanz1:
Updated to include all non-wsgi services as pointed out by the [commit message](https://github.com/openstack/oslo.messaging/commit/44b3427bc9efea9f341edfb8ea7aea38f25d1a5a).

mzhang741 (dismissed):

comay:
@zhanz1 Thinking about this, does it make sense to have separate attribute values? I'm thinking it might make sense to just define `default['bcpc']['openstack']['oslo_messaging_rabbit']['heartbeat_in_pthread']` or even `default['bcpc']['oslo_messaging_rabbit']['heartbeat_in_pthread']`and use that value across all of the configuration files.

zhanz1:
> @zhanz1 Thinking about this, does it make sense to have separate attribute values? I'm thinking it might make sense to just define `default['bcpc']['openstack']['oslo_messaging_rabbit']['heartbeat_in_pthread']` or even `default['bcpc']['oslo_messaging_rabbit']['heartbeat_in_pthread']`and use that value across all of the configuration files.

Understood, but I still want to keep it this way, because even though by the explanation, all non-wsgi services are affected (and therefore they shall be treated the same - config set to false), we've only seen nova & neutron failed (and other people only complained about these two services too). This means that these services are interacting with `oslo.messaging` differently so I think an individual on/off switch for each service would make more sense, just in case the future fix "works for cinder, but not nova". Ideally the changes of this PR will be reverted if the issue is fully fixed.

comay:
> Updated to include all non-wsgi services as pointed out by the [commit message](https://github.com/openstack/oslo.messaging/commit/44b3427bc9efea9f341edfb8ea7aea38f25d1a5a).

Also, Keystone also. honors `[oslo_messaging_rabbit]/heartbeat_in_pthread`.

zhanz1:
> > Updated to include all non-wsgi services as pointed out by the [commit message](https://github.com/openstack/oslo.messaging/commit/44b3427bc9efea9f341edfb8ea7aea38f25d1a5a).
> 
> Also, Keystone also. honors `[oslo_messaging_rabbit]/heartbeat_in_pthread`.

Double checked and I think keystone is a wsgi service:
```
$ ps aux | grep keystone
keystone ..... (wsgi:keystone-pu -k start
keystone ..... (wsgi:keystone-pu -k start
```

mboruta1:
> > @zhanz1 Thinking about this, does it make sense to have separate attribute values? I'm thinking it might make sense to just define `default['bcpc']['openstack']['oslo_messaging_rabbit']['heartbeat_in_pthread']` or even `default['bcpc']['oslo_messaging_rabbit']['heartbeat_in_pthread']`and use that value across all of the configuration files.
> 
> Understood, but I still want to keep it this way, because even though by the explanation, all non-wsgi services are affected (and therefore they shall be treated the same - config set to false), we've only seen nova & neutron failed (and other people only complained about these two services too). This means that these services are interacting with `oslo.messaging` differently so I think an individual on/off switch for each service would make more sense, just in case the future fix "works for cinder, but not nova". Ideally the changes of this PR will be reverted if the issue is fully fixed.

I think what David proposes makes sense and reduces the need for attributes to be specified in each service, but since this will likely be reverted as part of a future Caracal upgrade and it gives us a bit of flexibility in the meantime I am fine with it as it is.

mboruta1 (dismissed):

comay:
My remaining concern for this PR is we're bifurcating the services in some will use `eventlet` and some will not. Since some of the services in a service group (for example, `cinder-api` and `cinder-volume`) share the same configuration value for `heartbeat_in_pthread`, we need to choose one value which is `False`.

What I'm suggesting is that for now, we set `heartbeat_in_pthread` to `False` in all consumers even if they're strictly WSGI-based, such as the two Keystone services. This would be akin to pulling in a patched `oslo.messaging` which is what will happen anyway when we eventually move the Yoga baseline forward.

Put another way, what benefit do we have in running any of the WSGI services with the new threading model for the heartbeat?

mboruta1:
> My remaining concern for this PR is we're bifurcating the services in some will use `eventlet` and some will not. Since some of the services in a service group (for example, `cinder-api` and `cinder-volume`) share the same configuration value for `heartbeat_in_pthread`, we need to choose one value which is `False`.
> 
> What I'm suggesting is that for now, we set `heartbeat_in_pthread` to `False` in all consumers even if they're strictly WSGI-based, such as the two Keystone services. This would be akin to pulling in a patched `oslo.messaging` which is what will happen anyway if we eventually move the Yoga baseline forward.
> 
> Put another way, what benefit do we have in running any of the WSGI services with the new threading model for the heartbeat?

Good point, no need to bifurcate :), setting everything to FALSE sounds even better. Equivalent to when Yoga baseline is moved forward.

mboruta1 (dismissed):

tj90241 (dismissed):
Fantastic work and great analysis!

comay:
Looks good, @zhanz1. Just a note that it looks like you have extra spacing on the line you added to the `.erb` files before the closing `%>`. I only noticed that because at least on my browser, https://github.com/bloomberg/chef-bcpc/pull/2297/files rendered the lines slightly differently from one another.

zhanz1:
> Looks good, @zhanz1. Just a note that it looks like you have extra spacing on the line you added to the `.erb` files before the closing `%>`. I only noticed that because at least on my browser, https://github.com/bloomberg/chef-bcpc/pull/2297/files rendered the lines slightly differently from one another.

Apologies, must've been the extra white spaces added when copy & pasting. Fixed now :)

comay (approved):

PR #2298, opened 2024-04-02
Implementing the preference of iptables over any existing nftables configuration

desrod:
Quick and easy PR to configure the iptables alternatives to prefer iptables over nftables, due to performance issues described in chef-bcpc/pull/2273. 

There are two variants to this, depending on whether nftables is installed, or intended to be installed, so you can toggle back and forth. This PR implements the most-basic of one variant, to assert iptables, when nftables is _not_ already installed.


ansible/playbooks/roles/common/tasks/configure-system.yml
```diff
@@ -269,6 +271,14 @@
     - "vim"
     - "xinetd"
 
+# Ensure iptables is installed and made the default option over nftables
+- name: Switch default to iptables over nftables
+  ansible.builtin.command:
```
comay, line 276: Besides addressing the above `ansible-lint` issue, could you replace `ansible.builtin.command` with just `command` to keep things consistent with the other invocations in this and other playbooks?

ansible/playbooks/roles/common/tasks/configure-system.yml
```diff
@@ -269,6 +271,14 @@
...
+    cmd: update-alternatives --set {{ item }} /usr/sbin/{{ item }}-legacy
+  loop:
+    - iptables
+    - ebtables
```
comay, line 280: So isn't this going to always switching this to using `iptables` even if `nftables` is being used on certain node types? I think this section of the playbook should always set the alternatives accordingly to which node type it is, so instead of being a `name: Switch default to iptables over nftables` it's more of a `name: Update alternatives to iptables or nftables". Then it should set the alternatives given to which backend is being used.

tj90241:
I might suggest eliding any `eatables` configuration since It's not present on Focal anyways, though not stuck on the idea either.

ansible/playbooks/roles/common/tasks/configure-system.yml
tj90241: fwiw, `ansible.builtin.command` (FQDN) is preferred generally, though I can see you wanting to be consistent within this repo.

comay: Yup - I figured consistency is preferable here. I considered making the switch as part of #2261 but I was already touching so many lines of code....

desrod:
> I might suggest eliding any `eatables` configuration since It's not present on Focal anyways, though not stuck on the idea either.

`ebtables` gets pulled in as a dependency of nova-compute-libvirt later on anyway, but I can't correct set the update-alternatives links for them at the point when configure-system runs, because the package isn't installed yet, that's why its here, else I would have to run update-alternatives _after_ `nova-compute-libvirt` and its dependencies are installed.

tj90241:
ooooh, nice catch -- was not looking at worknodes!

ansible/playbooks/roles/common/tasks/configure-system.yml
desrod: @comay this is a bit more complex than just flipping the update-alternatives links over for those running, or not running `nftables`. 

The `nftables` service has to be installed/enabled for those running nft, and disabled (and alternately uninstalled) for those not running it. 

My original pair of playbooks had this construct, but we simplified it based on wanting iptables everywhere, but now we're back to the original idea, where it has to switch intent based on the role, and not just the felix config preferring 'Legacy'. I'm reworking this now to try to shoehorn that back in.

tj90241:
@desrod the logic looks sound now!

Please address the linter problems above and I think we should be good to merge.

FYI, the `when` logic can also be condensed with `block:`:
```
- name: A collection of steps
  block:
  when: <my_condition>
    - name: Step A
      command: ...
    - name: Step B
      command: ...
```

psipika (approved):
Looking for a second 👍 from @tj90241

PR #2300, opened 2024-04-09
RDTIBCC-5206: cinder: Increase the default file descriptor limit

tj90241:
When deleting a lot of volumes with a high rate of
concurrency and when there are a lot of backend eventlets,
cinder-volume/librados seems to exhaust the default quota
of 64k fds. Bump the default to 1M to counteract this.


tj90241:
<h2>Testing performed:</h2>
On a cluster of ~300 OSDs, perform the following:
  * Create 100 volumes
  * Attach to 100 VMs
  * Write 1GB to each volume
  * Delete VMs

Finally, delete the volumes ~40x at a time. When deleting the volumes, sample the # of fds opened by `cinder-volume` at that instant every 0.25 seconds:
```
173
181
903
1267
2310 <-- first wave of 40 volumes
2126
916
199
199
207
194
562
1197
1725 <-- second wave of 40 volumes
2179
568
202
202
202
202
810
1897 <-- remaining wave of 20 volumes
926
213
198
```

... and observe that all volumes delete successfully. Next, artifically limit fds to 1k by leveraging this PR and repeat. When sampling fds similarly, we see:
```
170
170
170
170
170
170
170
278
780
820
1150
1150
1150
1150
1150
1150
1150
1150
1150
1150
1150
1150
1150
1150
1150
1150
1150
1150
1150
...
```

... at which point several volumes are seen stuck in a deleting state indefinitely.

It seems that the peak fd usage roughly corresponds to:
```
max_peak_fds = volumes_deleting * num_osds_in_cluster / # headnodes in the cluster
```

Thus, a cluster with ~5k OSDs can, in a theoretical case, only be expected to delete 65536*3/5000 = ~39 volumes simultaneously across 3 headnodes with the default fd limit of 64k (assuming that the volumes deletion requests are perfectly distributed across all headnodes). This also means that any value of `backend_native_threads_pool_size` >= 40/3 is likely problematic for any such cluster.

We hypothesize that what is happening is that if an fd needed cannot be allocated during volume deletion, then the volume deletion request ties up until fds become available. This, in turn, results in `cinder-volume` getting itself into a "log jam" situation wherein every deleting volume is waiting on every other deleting volume being processed to release some fds so that its own deletion can proceed.

Raising the fds to ~512k ensures even that if all requests are routed to a single headnode that it can process upto a minimum of ~100 concurrent volume deletions in roughly the worst possible case. This should mean that Cinder's `backend_native_threads_pool_size` can comfortably be raised to ~50-100, even for extremely large Ceph clusters.

chaowang987 (approved):
Looks good! nice document! 👍 👍

comay (approved):

